PR #47, opened 2017-05-11
(SDK-218) Update acceptance testing to use ENV['SHA']

james-stocks:
The actual environment variable that CI uses to store the puppet-sdk package reference is SHA


james-stocks:
Closing because PR needs rebased and also because we opened the repo and need a new PR anyway

PR #48, opened 2017-05-12
(SDK-197) 'new class' CLI

rodjek:
Mostly this is laying out a generic framework for generating new objects (class, defined type, fact, function, etc) inside a module. The only bits specific to generating new Puppet classes is a class containing 3 methods, the rest should be reusable in other parts of the CLI ('new defined_type' for example).

There is a [companion PR](https://github.com/puppetlabs/pdk-module-template/pull/1) to the pdk-module-template repo to add the templates for this command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The `pdk new module` command accepts the following arguments and options. Argume
 
 #### `--template-url=git_url`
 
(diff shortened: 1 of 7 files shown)


rodjek:
Due to changing this repo from private to public, my fork has been disconnected from this repo and I can no longer add new code to the PR. Merging this and opening a new PR to relax the data type validation.

PR #58, opened 2017-06-02
(SDK-232) operatingsystem_support, and serverspec acceptance testing

DavidS:


lib/pdk/cli/exec.rb
```diff
@@ -43,7 +45,9 @@ def self.git_bindir
       end
 
       def self.git(*args)
-        git_path = ENV['PDK_USE_SYSTEM_BINARIES'].nil? ? File.join(git_bindir, 'git') : 'git'
+        vendored_bin_path = File.join(git_bindir, 'git')
+        git_path = File.exists?(vendored_bin_path) ? vendored_bin_path : 'git'
+        PDK.logger.debug(_("Using git from the system PATH, instead of '%{vendored_bin_path}'") % { vendored_bin_path: vendored_bin_path})
```
rodjek: :+1:

lib/pdk/module/metadata.rb
```diff
@@ -17,6 +17,24 @@ class Metadata
...
+          },
+          {
+            "operatingsystem" => "Ubuntu",
+            "operatingsystemrelease" => [ "16.04" ]
+          },
+          {
+            "operatingsystem" => "Windows",
+            "operatingsystemrelease" => [ "2016" ]
+          },
+        ]
```
rodjek: Do we want to be having all these operating systems set by default or interview for them?

lib/pdk/cli/exec.rb
scotje: @DavidS I know you mentioned the possibility of adding some sort of "don't fall back to system PATH" option for testing but maybe it would be sufficient to assert that the above DEBUG message isn't logged? No big deal either way.

scotje:
I didn't try actually running them yet but sounded like @james-stocks was going to try that. Code changes look good.

DavidS:
moved resolved pending example (via #63) to the normal test

rodjek:
:sparkles:

PR #62, opened 2017-06-05
(SDK-261) Manage basic bundler operations for module dev

scotje:
This seems to be working pretty well at this point.

I added a conditional to disable the spinner stuff on Windows for now because there are some special considerations for making control sequences work under cmd/ps. Definitely solvable but probably more appropriate for a subsequent release.

There are a couple PRs backed up in the packaging repo to make this all work with packaged builds, but the paths etc. are all up to date with where things will eventually be.

### DIFF
--- a/lib/pdk/cli/exec.rb
+++ b/lib/pdk/cli/exec.rb
@@ -14,6 +14,9 @@ module PDK
         process.io.stderr = Tempfile.new('stderr')
 
         begin
+          # Make this process leader of a new group
+          process.leader = true
+
           # start the process
           process.start
 
@@ -23,6 +26,7 @@ module PDK
           stdout = process.io.stdout.open.read
           stderr = process.io.stderr.open.read
         ensure
+          process.stop unless process.exited?
           process.io.stdout.close
           process.io.stderr.close
         end
@@ -46,9 +50,24 @@ module PDK
 
       def self.git(*args)
         vendored_bin_path = File.join(git_bindir, 'git')
-        git_path = File.exists?(vendored_bin_path) ? vendored_bin_path : 'git'
-        PDK.logger.debug(_("Using git from the system PATH, instead of '%{vendored_bin_path}'") % { vendored_bin_path: vendored_bin_path})
-        execute(git_path, *args)
+
+        execute(try_vendored_bin(vendored_bin_path, 'git'), *args)
+      end
+
+      def self.bundle(*args)
+        bundle_bin = Gem.win_platform? ? 'bundle.bat' : 'bundle'
+        vendored_bin_path = File.join(pdk_basedir, 'private', 'ruby', '2.1.9', 'bin', bundle_bin)
+
+        execute(try_vendored_bin(vendored_bin_path, bundle_bin), *args)
+      end
+
+      def self.try_vendored_bin(vendored_bin_path, fallback)
+        if File.exists?(vendored_bin_path)
+          return vendored_bin_path
+        else
+          PDK.logger.debug(_("Trying '%{fallback}' from the system PATH, instead of '%{vendored_bin_path}'") % {fallback: fallback, vendored_bin_path: vendored_bin_path})
+          return fallback
+        end
       end
     end
   end
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -12,5 +12,19 @@ module PDK
       Dir::Tmpname.make_tmpname(File.join(Dir.tmpdir, base), nil)
     end
     module_function :make_tmpdir_name
+
+    # Returns the fully qualified path to a per-user PDK cachedir.
+    #
+    # @return [String] Fully qualified path to per-user PDK cachedir.
+    def cachedir
+      if Gem.win_platform?
+        basedir = ENV['APPDATA']
+      else
+        basedir = Dir.home
+      end
+
+      return File.join(basedir, '.pdk', 'cache')
+    end
+    module_function :cachedir
   end
 end
--- a/lib/pdk/util/bundler.rb
+++ b/lib/pdk/util/bundler.rb
@@ -1,0 +1,139 @@
+require 'bundler'
+require 'tty-spinner'
+require 'pdk/util'
+require 'pdk/cli/exec'
+
+module PDK
+  module Util
+    module Bundler
+      class BundleHelper; end
+
+      def self.ensure_bundle!
+        bundle = BundleHelper.new
+
+        if bundle.has_gemfile?
+          if !bundle.locked?
+            unless bundle.lock!
+              raise PDK::CLI::FatalError, _("Unable to resolve Gemfile dependencies.")
+            end
+          end
+
+          if !bundle.installed?
+            unless bundle.install!
+              raise PDK::CLI::FatalError, _("Unable to install missing Gemfile dependencies.")
+            end
+          end
+        end
+      end
+
+      class BundleHelper
+        def has_gemfile?
+          # return a pure boolean
+          !!gemfile
+        end
+
+        def locked?
+          # return a pure boolean
+          !!gemfile_lock
+        end
+
+        def installed?
+          output_start(_("Checking for missing Gemfile dependencies"))
+
+          result = invoke('check', "--gemfile=#{gemfile}", "--path=#{bundle_cachedir}")
+
+          output_end(:success)
+
+          return result[:exit_code] == 0
+        end
+
+        def lock!
+          output_start(_("Resolving Gemfile dependencies"))
+
+          result = invoke('lock')
+
+          if result[:exit_code] == 0
+            output_end(:success)
+          else
+            output_end(:failure)
+            $stderr.puts result[:stdout]
+            $stderr.puts result[:stderr]
+          end
+
+          return result[:exit_code] == 0
+        end
+
+        def install!
+          output_start(_("Installing missing Gemfile dependencies"))
+
+          result = invoke('install', "--gemfile=#{gemfile}", "--path=#{bundle_cachedir}")
+
+          if result[:exit_code] == 0
+            output_end(:success)
+          else
+            output_end(:failure)
+            $stderr.puts result[:stdout]
+            $stderr.puts result[:stderr]
+          end
+
+          return result[:exit_code] == 0
+        end
+
+        private
+
+        def invoke(*args)
+          ::Bundler.with_clean_env do
+            PDK::CLI::Exec.bundle(*args)
+          end
+        end
+
+        def find_upwards(target)
+          previous = nil
+          current  = File.expand_path(Dir.pwd)
+
+          until !File.directory?(current) || current == previous
+            filename = File.join(current, target)
+            return filename if File.file?(filename)
+            current, previous = File.expand_path("..", current), current
+          end
+        end
+
+        def gemfile
+          @gemfile ||= find_upwards('Gemfile')
+        end
+
+        def gemfile_lock
+          @gemfile_lock ||= find_upwards('Gemfile.lock')
+        end
+
+        def bundle_cachedir
+          @bundle_cachedir ||= File.join(PDK::Util.cachedir, 'bundler')
+        end
+
+        # These two output_* methods are just a way to not try to do the spinner stuff on Windows for now.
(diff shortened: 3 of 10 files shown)


lib/pdk/util/bundler.rb
```diff
@@ -0,0 +1,129 @@
...
+
+        def locked?
+          # return a pure boolean
+          !!gemfile_lock
+        end
+
+        def installed?
+          output_start(_("Checking for missing Gemfile dependencies"))
+
+          result = invoke('check', "--gemfile=#{gemfile}", "--path=#{pdk_cachedir}")
```
rodjek: I'd recommend changing the name of `pdk_cachedir` to something like `pdk_gemdir` to avoid confusion. In Bundler `cache` generally refers to the contents of `vendor/cache`, as opposed to the value of `--path` which is where the bundled gems are installed to.

lib/pdk/cli/exec.rb
```diff
@@ -46,6 +46,13 @@ def self.git(*args)
         git_path = ENV['PDK_USE_SYSTEM_BINARIES'].nil? ? File.join(git_bindir, 'git') : 'git'
         execute(git_path, *args)
       end
+
+      def self.bundle(*args)
+        bundle_path = ENV['PDK_USE_SYSTEM_BINARIES'].nil? ? File.join(pdk_basedir, 'private', 'ruby', '2.1.9', 'bin', 'bundle') : 'bundle'
```
DavidS: This conflicts with https://github.com/puppetlabs/pdk/pull/58/commits/2d0ca925a57f0db56da62674191f6f348f6dfb75 from #58 - maybe I should pull this change out to get it merged independently, so we can leave the `PDK_USE_SYSTEM_BINARIES` behind us.

lib/pdk/util/bundler.rb
```diff
@@ -0,0 +1,129 @@
...
+
+          output_end(:success)
+
+          result[:exit_code] == 0
+        end
+
+        def lock
+          output_start(_("Resolving Gemfile dependencies"))
+
+          result = invoke('lock')
```
DavidS: You're passing --gemfile to other bundler invocations. shouldn't this one too?

spec/util/bundler_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    allow(PDK::CLI::Exec).to receive(:execute).and_return({exit_code: 0})
+  end
+
+  describe '.ensure_bundle!' do
+    context 'when there is no Gemfile' do
+      before(:each) do
+        allow(File).to receive(:file?).with(/Gemfile$/).and_return(false)
+      end
+
+      it 'does nothing' do
```
DavidS: In which situations do you expect this to be the case?

spec/util/bundler_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    context 'when there is no Gemfile.lock' do
+      before(:each) do
+        allow(File).to receive(:file?).with(/Gemfile$/).and_return(true)
+        allow(File).to receive(:file?).with(/Gemfile\.lock$/).and_return(false)
+      end
+
+      it 'generates Gemfile.lock' do
+        expect(PDK::CLI::Exec).to receive(:execute).with(/bundle(\.bat)?$/, 'lock', any_args).and_return({exit_code: 0})
```
DavidS: You can save yourself some regexing here by `expect`ing on `PDK::CLI::Exec.bundle`, instead of `.exec`, and test the path resolver separately. This will also make changing the grizzards of `PDK::CLI::Exec` much easier on the tests.

lib/pdk/cli/exec.rb
DavidS: This got merged, so you can just use the fallback code from there.

DavidS:
With #58 merged, it would also be good to start with a bit of acceptance testing, to get into the swing of it. @james-stocks and I talked about negative test cases yesterday, and - since we're integrating a third party tool - we don't need to go overboard here. Checking whether our integration reacts correctly to a removed `Gemfile.lock`, and a broken `Gemfile` is probably enough.

lib/pdk/util/bundler.rb
scotje: For whatever reason, `bundle lock` doesn't have a `--gemfile` option.

lib/pdk/cli/exec.rb
scotje: Yes, I was planning to adopt that approach once it got merged.

spec/util/bundler_spec.rb
scotje: Well, technically a puppet module doesn't have to have a Gemfile right? Theoretically somebody could still use `pdk` to run lint or something on such a module. Did we decide to ignore any use cases that did not involve a module that was generated with `pdk new`?

spec/util/bundler_spec.rb
scotje: Yeah that would be simpler at this point, I think the original tests were written before the refactor that added the `PDK::CLI::Exec.bundle` method.

lib/pdk/util/bundler.rb
scotje: Yeah, right now it points to a generic directory we can use for any sort of caching we need to do with PDK and the default gem pathing of `ruby/2.1.0/...` gets added to the passed in value by bundler, so the value that is getting interpolated there isn't really specific to gems.

scotje:
@DavidS Yes we should make the bundler stuff exit on failure so that it doesn't proceed onto the unit test stuff. I also agree that we could move the spinner stuff out into the Exec class but I would prefer to do that in a new PR.

spec/util/bundler_spec.rb
DavidS: Yes, technically no Gemfile is required. Practically, all well-maintained do.

The goal is to provide a template so that `pdk new` generated modules look very similar to what is currently "on the market", to the point where the pdk can handle unmodified supported and voxpupuli modules.

For less well maintained modules, having a `pdk upgrade` command applying a template on top of an existing module provides more value, as it'll reduce long-term friction of supporting/working with multiple models.

spec/util/bundler_spec.rb
DavidS: 👍

bmjen:
@DavidS @rodjek Discussed this with @scotje yesterday, we also need a mechanism for running `bundle update`... not sure whether we should do this as a manual operation or programmatic, and if programmatic... at what frequency?

DavidS:
The more we talk about these things, the more I convince myself that iristyle had it right, in that we need to provide a completely version-locked gemset. In that case, folks would never have to bundle update until they want to use the next version of the gemset, where we can hide that in the template update code.

spec/util/bundler_spec.rb
scotje: I guess we just need to figure out how the most graceful way to handle the lack of a Gemfile is. We can change the method name to something like `require_bundle!` to make it clear that the tool should exit 1 if there is no Gemfile present in the module.

scotje:
Added some acceptance tests. I had to wrangle things quite a bit to be successful invoking `pdk` inside of a generated module without picking up unwanted context from being in the pdk working directory. I approached it by using a tmpdir and generating a binstub for `pdk` but the same thing could perhaps be accomplished with `BUNDLE_GEMFILE`.

Acceptance tests are currently not working in AppVeyor but I've run out of time tonight to debug, perhaps @DavidS or @james-stocks can take a look during their workday.

lib/pdk/util/bundler.rb
rodjek: :+1: to having a general cache dir for the PDK. To prevent possible conflicts, I think we should standardise on `#{pdk_cachedir}/<toolname>` (in case some other tool that we want to integrate also wants to create a `ruby` directory in its cache for example).

spec/acceptance/bundle_management_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    before(:all) do
+      shell_ex('rm -f test_module/Gemfile.lock')
+    end
+
+    after(:all) do
+      shell_ex('rm -f test_module/Gemfile.lock')
+    end
+
+    # TODO: come up with a way to invoke only the bundler stuff without trying to run unit tests
+    describe command("cd test_module && #{path_to_pdk} test unit") do
```
james-stocks: This fails on powershell as && is not an operator there.
I think the only command that works on both Linux and PS is "cd test_module; #{path_to_pdk} test unit"

That comes with the weakness of always running command B even if command A fails.
There are [workarounds](https://stackoverflow.com/questions/11104197/chain-commands-in-powershell-command-line) to have Powershell fail and not run command B; but it would mean a command that doesn't work in Linux.

_Maybe_ we could have a helper to hide the platform specifics, like

    describe command("cd test_module").followed_by_command.(#{path_to_pdk} test unit") do

?

james-stocks:
@rodjek @DavidS I added one commit to get appveyor passing; but I think the current failures might be because binstub scripts don't work in Windows?
i.e. I cannot do this on my Windows VM:

    C:/projects/pdk/bin/pdk test unit

...Windows just prompts me to pick a program for this unknown filetype.

I'm unsure if this is exactly why appveyor is failing; I'm not seeing the _Exec format error_ errors it's encountering

spec/acceptance/bundle_management_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper_acceptance'
+require 'fileutils'
+
+describe 'Managing Gemfile dependencies' do
+  before(:all) do
+    # Do everything in a clean tmpdir.
+    @old_pwd = Dir.pwd
+    @tmpdir = Dir.mktmpdir('pdk-acceptance')
+
+    Dir.chdir(@tmpdir)
```
DavidS: This will only work locally, but not when using beaker. We'll need to extend beaker-testmode_switcher's API to provide better support for such operations.

spec/acceptance/bundle_management_spec.rb
DavidS: I've checked serverspec's source, and there doesn't seem to be any way to specify any runtime environment (like CWD) to the `command()`.

spec/acceptance/bundle_management_spec.rb
scotje: Oh yes, I originally had the tmpdir code in spec_helper_acceptance and wrapped in a conditional for testmode == local. It would definitely be better to have the functionality built into the tooling.

lib/pdk/util.rb
```diff
@@ -12,5 +12,13 @@ def make_tmpdir_name(base)
       Dir::Tmpname.make_tmpname(File.join(Dir.tmpdir, base), nil)
     end
     module_function :make_tmpdir_name
+
+    # Returns the fully qualified path to a per-user PDK cachedir.
+    #
+    # @return [String] Fully qualified path to per-user PDK cachedir.
+    def cachedir
+      File.join(Dir.home, '.pdk', 'cache')
```
scotje: 👍

DavidS:
@james-stocks these errors are shell_ex from beaker-tms, not serverspec's command(), I'm pretty sure the two use different ways to execute commands, I would not be surprised if a single path_to_pdk does not work for them at all

scotje:
Going to split the acceptance tests back out of this PR, there is a distinct ticket for them anyway and it seems like there are some Windows things that need to be worked out.

scotje:
Made the cachedir changes suggested by @rodjek and squashed.

lib/pdk/cli/exec.rb
```diff
@@ -46,9 +50,24 @@ def self.git_bindir
 
       def self.git(*args)
         vendored_bin_path = File.join(git_bindir, 'git')
-        git_path = File.exists?(vendored_bin_path) ? vendored_bin_path : 'git'
-        PDK.logger.debug(_("Using git from the system PATH, instead of '%{vendored_bin_path}'") % { vendored_bin_path: vendored_bin_path})
```
scotje: note that this was always logging this message even when the vendored bin was being used :)

DavidS: d'oh

PR #65, opened 2017-06-06
(SDK-197) add acceptance tests for new class command

DavidS:
Surely these will be broken on non-linux, but I need to get it off this
workstation.

### DIFF
--- a/spec/acceptance/new_class_spec.rb
+++ b/spec/acceptance/new_class_spec.rb
@@ -10,7 +10,16 @@ describe 'Creating a new class' do
   end
 
   context 'when creating the main class' do
-    describe command("cd foo && BUNDLE_GEMFILE=../Gemfile #{path_to_pdk} new class foo") do
+    describe command("#{path_to_pdk} new class foo") do
+      before do
+        @old_pwd = Dir.pwd
+        Dir.chdir('foo')
+      end
+
+      after do
+        Dir.chdir(@old_pwd)
+      end
+
       its(:exit_status) { is_expected.to eq 0 }
       its(:stdout) { is_expected.to match(/Creating .* from template/) }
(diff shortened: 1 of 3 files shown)


spec/acceptance/new_class_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  before(:all) do
+    shell_ex("#{path_to_pdk} new module foo --skip-interview")
+  end
+
+  after(:all) do
+    shell_ex('rm -rf foo')
+  end
+
+  context 'when creating the main class' do
+    describe command("cd foo && BUNDLE_GEMFILE=../Gemfile #{path_to_pdk} new class foo") do
```
james-stocks: @jessereynolds work here might help with this:
https://github.com/puppetlabs/pdk/pull/62/files#diff-174f7dfb508155027cb4c3e9a1840591R28

DavidS: I assume you meant @scotje ?

Yeah binstubs look like a good way out here - I'll try that. It might also help with the bundle/rvm fighting we ran into yesterday...

jessereynolds: > I assume you meant @scotje ?

I assume so too :-) ... looks exciting thought.

rodjek: rvm: not even once

scotje: I think the binstubs do work on windows as well, you just have to explicitly run them with the ruby interpreter, so `ruby /path/to/binstub/pdk <args>`

rodjek:
I've changed the pdk path for local acceptance tests so that we generate and use the binstub instead of `bundle exec`. This way we can avoid having to set the `BUNDLE_GEMFILE` env var.

PR #67, opened 2017-06-07
(maint) refactor CLI initialisation to recommended CRI pattern

DavidS:
The convolute class structure was a overhang from previous Thor
experiments. Using CRI this can be simplified.
See https://github.com/ddfreyne/cri/tree/master/samples for examples.

Core points:
* Minimal includes
* flat structure
* no top-scope pollution
* each command needs only one require per subcommand
* subcommands are responsible for connecting up
* enable easier sharing of option declarations
* only require core parts when running a command

New CRI features used:
* use default values in declaration
* use default subcommands for top-level commands
  * always exit 0 on producing help


scotje:
In general this seems fine, I'm not a huge fan of requires at the bottom of a file but it's not the end of the world.

I do kind of like being able to see the attached subcommands from the parent command but that's also a minor thing.

DavidS:
rebased.

We could inline all sub-sub-commands, and only have singular new.rb, validate.rb, and test.rb. The commands aren't that big.

scotje:
Meh, I think I'll get over the requires at the bottom, this is one case where it actually makes some amount of sense. Looks like it needs a rebase again, I'll see if I can work that out though.

DavidS:
Thanks folks!

PR #68, opened 2017-06-07
(SDK-269) Add acceptance tests for bundle management.

scotje:
work in progress on adding bundle management acceptance tests, extracted from #62

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,5 @@
 # version: 1.0.{build}-{branch}
 
-cache:
-  - C:\gems
-
 environment:
   matrix:
     - RUBY_VERSION: 23
@@ -11,7 +8,7 @@ environment:
 install:
   - set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
   - SET LOG_SPEC_ORDER=true
-  - bundle install --jobs 4 --retry 2 --without development --path=C:\gems
+  - bundle install --jobs 4 --retry 2 --without development
 
 build: off
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,8 +2,8 @@
 
 environment:
   matrix:
-    - RUBY_VERSION: 23
-    - RUBY_VERSION: 21
+    - RUBY_VERSION: 23-x64
+    - RUBY_VERSION: 21-x64
 
 install:
   - set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,6 @@ test_script:
   - SET PATH=C:\%PLATFORM%\bin;%PATH%
   - SET LOG_SPEC_ORDER=true
   - bundle install --jobs 4 --retry 2 --without development
-  - bundle binstubs pdk
   - type Gemfile.lock
   - ruby -v
   - bundle exec rake spec acceptance:local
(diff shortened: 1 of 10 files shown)


DavidS:
Rebased on top of master with @rodjek's changes in #65. Let's see if this fares better.

DavidS:
Yey, this now fails on nokogiri 1.8.0 on ruby 2.1, and a mismatch in output on ruby 2.3. I'll call this a success.

spec/acceptance/bundle_management_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    describe file("#{target_dir}/Gemfile.lock") do
+      it { is_expected.to be_file }
+    end
+  end
+
+  context 'when there is an invalid Gemfile' do
+    before(:all) do
+      shell_ex("mv #{target_dir}/Gemfile #{target_dir}/Gemfile.old")
```
scotje: Not sure why it's not reflected in the error output, but I think `mv` is failing on Windows which prevents this test from generating the expected output.

scotje:
`nokogiri` thing seems to be an issue with the pinning in the puppet-module-gems for Ruby 2.1.9. cc @bmjen

spec/acceptance/bundle_management_spec.rb
scotje: Are there platform independent implementations for `rm` and `mv` available yet or do we need to add those?

DavidS: I had the feeling that all environments we're running in have cygwin or similar installed. Seems like I was wrong. Can we conclusively prove it one way or the other?

DavidS: re "silent failures", shell_ex returns a `BeakerResult` (or so) that has `exit_code`, `stdout`, and `stderr` members to inspect.

scotje:
I have a semi-permanent Windows dev machine set up in SLICE now so I'll try debugging the 2.3 failures with that.

spec/acceptance/bundle_management_spec.rb
rodjek: I think an important consideration here is also "Do/Can we assume that all our Windows users have a cygwin environment?"

DavidS:
pushed again with no beaker involved. These worked locally on Linux, as well as running against the 0a7b256aba8c7ae56c3867451a581ce409c439c7 puppet-sdk-x64.msi on a Win2012r2 VM.

DavidS:
So caching on appveyor is still broken, even with those changes, but at least we got a green build on appveyor now! 🎉

james-stocks:
The _acceptance:package_ command is now useless (I suspect it will appear to pass, but will just be executing the tests locally again).
I created SDK-275 for the work we want to do in having tests run against a VM

appveyor.yml
```diff
@@ -1,17 +1,24 @@
-build: off
+# version: 1.0.{build}-{branch}
 
-# ruby versions under test
-platform:
-  - Ruby23-x64
-  - Ruby21-x64
+environment:
+  matrix:
+    - RUBY_VERSION: 23
```
james-stocks: It looks like this changes from running 64bit Ruby to running 32 bit Ruby ( https://www.appveyor.com/docs/build-environment/#ruby )
Does that matter?

PR #70, opened 2017-06-08
(maint) Adds method to find base module dir.

bmjen:
Adds a method to find the base dir of the project module.


bmjen:
Closing b/c these commits are included in #71

PR #71, opened 2017-06-09
(SDK-240) Adds puppet-lint validation subcommand & other plumbing

bmjen:
In addition to getting the base puppet-lint validation subcommand working,
there are a lot of other plumbing changes included here. A layer of abstraction
was added here to be able to genericize the validation commands to
"Metadata", "Puppet", and "Ruby" with their respective generic validators.
These generic validators will be able to manage subvalidations that can be
created to handle the actual tools.

e.g. PuppetValidator is a wrapper that includes validation invocation for
PuppetLint and PuppetParser.

RubyValidator was created for consistency and future tools integrations
despite only having rubocop currently as the lone validator.

Requires #70


rodjek:
Pushed a commit to fix up the existing tests so that I can rebase the rubocop integration onto this and take advantage of your refactoring :)

PR #75, opened 2017-06-13
(SDK-244) Add rubocop validator

rodjek:
Refactors the validators further so that parsing targets from the CLI and invoking the command are part of the base class. This way, the only thing each validator needs to do is implement a method to build an array of arguments to pass to the command

### DIFF
--- a/lib/pdk/validators/base_validator.rb
+++ b/lib/pdk/validators/base_validator.rb
@@ -4,10 +4,27 @@ require 'pdk/cli/exec'
 module PDK
   module Validate
     class BaseValidator
+      def self.parse_targets(options)
+        # If no targets are specified, then we will run validations from the
+        # base module directory.
+        if options[:targets].nil? || options[:targets].empty?
+          [PDK::Util.module_root]
+        else
+          options[:targets]
+        end
+      end
+
+      def self.parse_options(options, targets)
+        targets
+      end
+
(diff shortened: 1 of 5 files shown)


lib/pdk/validators/ruby/rubocop.rb
```diff
@@ -11,13 +11,18 @@ def self.name
       end
 
       def self.cmd
-        'pwd'
+        'rubocop'
```
bmjen: I think for rubocop, we'll need to default to the `#{PDK::Util.module_root}/lib` directory instead, because rubocop isn't liking parsing of `.rubocop.yml` file in the module_root.

lib/pdk/validators/base_validator.rb
```diff
@@ -4,10 +4,27 @@
 module PDK
   module Validate
     class BaseValidator
+      def self.parse_targets(options)
```
bmjen: 👍  Nice DRY refactor.

lib/pdk/validators/ruby/rubocop.rb
bmjen: Tried this again on a newly generated module with updated templates, and rubocop is ok with the rubocop.yml, but still finding issues with Gemfile and spec_helper.rb. I'll investigate.

PR #76, opened 2017-06-14
(SDK-190) Acceptance tests for using commands outside module folder

james-stocks:
Adds an acceptance spec for using pdk new class when not inside a module folder.
Adds a pending acceptance spec for using pdk test unit when not inside a module folder.
Does not address pdk validate when not inside a module folder - SDK-277 is an existing bug for this.


james-stocks:
Closing and re-opening to trigger CI

DavidS:
Rebased on top of current master.

PR #46, opened 2017-05-09
(maint) rubocop autocorrect

DavidS:
This addresses more than half the complaints from rubocop. The only
necessary manual change was a `unless` condition in the Rakefile.

PR #78, opened 2017-06-14
(SDK-276) Rubocop rules and cleanup

DavidS:


DavidS:
Refresh for #46

DavidS:
@scotje agreed, those are two of the cops I also find weak. I'll merge this for now to avoid more rework, but it's definitely something we can revisit/change down the line, if they prove to be too obnoxious.

PR #82, opened 2017-06-15
(maint) Expand Windows 8.3 paths in templatedir

james-stocks:
PDK::Util.make_tmpdir_name might return a short (8.3) path on Windows.
This commits adds a canonical path method to allow assurance of using
the full path.

This fixes module generation on Windows; previously PDK was failing to
match C:\Users\Administrator and C:\Users\ADMIN~1

### DIFF
--- a/lib/pdk/cli/exec.rb
+++ b/lib/pdk/cli/exec.rb
@@ -135,7 +135,12 @@ module PDK
           begin
             @process.start
           rescue ChildProcess::LaunchError => e
-            raise PDK::CLI::FatalError, _("Failed to execute '%{command}': %{message}") % { command: argv.join(' '), message: e.message }
+            msg = if @process.respond_to?(:argv)
+                    _("Failed to execute '%{command}': %{message}") % { command: @process.argv.join(' '), message: e.message }
+                  else
+                    _('Failed to execute process: %{message}') % { message: e.message }
+                  end
+            raise PDK::CLI::FatalError, msg
           end
 
           if timeout
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -1,5 +1,6 @@
 require 'tmpdir'
 require 'tempfile'
+require 'puppet/util/windows'
 
 module PDK
   module Util
@@ -31,6 +32,23 @@ module PDK
       Dir::Tmpname.make_tmpname(File.join(Dir.tmpdir, base), nil)
     end
     module_function :make_tmpdir_name
+
+    # Return an expanded, absolute path
+    #
+    # @param path [String] Existing path that may not be canonical
+    #
+    # @return [String] Canonical path
+    def canonical_path(path)
+      if Gem.win_platform?
+        unless File.exist?(path)
+          raise PDK::CLI::FatalError, _("Cannot resolve a full path to '%{path}' as it does not currently exist") % { path: path }
+        end
+        Puppet::Util::Windows::File.get_long_pathname(path)
+      else
+        File.expand_path(path)
+      end
+    end
+    module_function :canonical_path
 
     # Returns the fully qualified path to a per-user PDK cachedir.
     #
--- a/lib/puppet/util/windows/api_types.rb
+++ b/lib/puppet/util/windows/api_types.rb
@@ -1,0 +1,278 @@
+require 'ffi'
+require 'puppet/util/windows/string'
+
+module Puppet::Util::Windows::APITypes
+  module ::FFI
+    WIN32_FALSE = 0
+
+    # standard Win32 error codes
+    ERROR_SUCCESS = 0
+  end
+
+  module ::FFI::Library
+    # Wrapper method for attach_function + private
+    def attach_function_private(*args)
+      attach_function(*args)
+      private args[0]
+    end
+  end
+
+  class ::FFI::Pointer
+    NULL_HANDLE = 0
+
+    def self.from_string_to_wide_string(str, &block)
+      str = Puppet::Util::Windows::String.wide_string(str)
+      FFI::MemoryPointer.new(:byte, str.bytesize) do |ptr|
+        # uchar here is synonymous with byte
+        ptr.put_array_of_uchar(0, str.bytes.to_a)
+
+        yield ptr
+      end
+
+      # ptr has already had free called, so nothing to return
+      nil
+    end
+
+    def read_win32_bool
+      # BOOL is always a 32-bit integer in Win32
+      # some Win32 APIs return 1 for true, while others are non-0
+      read_int32 != FFI::WIN32_FALSE
+    end
+
+    alias_method :read_dword, :read_uint32
+    alias_method :read_win32_ulong, :read_uint32
+    alias_method :read_qword, :read_uint64
+
+    alias_method :read_hresult, :read_int32
+
+    def read_handle
+      type_size == 4 ? read_uint32 : read_uint64
+    end
+
+    alias_method :read_wchar, :read_uint16
+    alias_method :read_word,  :read_uint16
+    alias_method :read_array_of_wchar, :read_array_of_uint16
+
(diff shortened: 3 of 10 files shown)


DavidS:
Poking around a bit in the win32 gems, I found that https://github.com/djberg96/win32-file has `long_path`/`short_path` methods that could make this nicer.

scotje:
I have a branch where I was working on this with tragiccode in slack yesterday, but the win32 gem solution may be more robust, I was using relative paths: https://github.com/puppetlabs/pdk/compare/debug_windows_path

DavidS:
according to the docs `relative_path_from` does not go to the file system, we assumed therefore that it won't work well with 8.3 confusion going on here.

scotje:
It actually seems to work fine, it just does `../` all the way back to the place before the 8.3 style path and then rebuilds from there. Like I said, the win32 stuff will probably be more elegant. :)

DavidS:
The errors in this PR seem to be connected to `pdk new module foo --skip-interview` silently not correctly rendering a module.

glennsarti:
As an FYI, we actively try to remove any of the win32-* gems from Puppet due to a variety of reasons.  Are you absolutely sure you need to bring in this gem to something as simple as 8.3 name expansion?

james-stocks:
@glennsarti @jpogran This PR has been updated to use _fiddle_ to access GetLongPathNameA within kernel32.dll, instead of using win32-gem

The PR still needs 1-2 test fixes plus a tidy-up; but a review specifically on this change would be appreciated, cheers

glennsarti:
@james-stocks You need to use the W version of the API plus be able to handle double byte characters.  Otherwise it'll break for non-English letters e.g. Jöhn

lib/pdk/util.rb
```diff
@@ -1,8 +1,38 @@
...
+
+        # Call into kernel32.dll to resolve short path e.g. C:\Users\~ADMIN1\foo
+        # should be changed to long path i.e. C:\Users\Administrator\foo
+        def long_path(short_path)
+          # Fetch long path name into a buffer
+          bufferlen = 255
+          buffer    = nil
+          loop do
+            buffer = ' ' * bufferlen
+            len    = GetLongPathNameA(short_path.to_s, buffer, buffer.size)
```
glennsarti: There should be a check here if GetLongPathName fails (RC = 0)

https://msdn.microsoft.com/en-us/library/windows/desktop/aa364980(v=vs.85).aspx

glennsarti: You need to use the W (`GetLongPathNameW`) version of the API plus be able to handle double byte characters.  Otherwise it'll break for non-English letters e.g. Jöhn

https://github.com/djberg96/win32-file/blob/ffi/lib/win32/file.rb#L189-L200

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/file.rb#L267-L279

lib/pdk/util.rb
```diff
@@ -1,8 +1,38 @@
...
+            # In the case of the path name being exceptionally long; resize the buffer
+            # to the actual size, then re-get it
+            if bufferlen < len
+              bufferlen = len
+              continue
+            end
+            break
+          end
+          buffer.rstrip.chomp("\0")
```
glennsarti: Also, you need to free the buffer after use otherwise you'll leak memory

lib/pdk/util.rb
```diff
@@ -1,8 +1,38 @@
...
+        def long_path(short_path)
+          # Fetch long path name into a buffer
+          bufferlen = 255
+          buffer    = nil
+          loop do
+            buffer = ' ' * bufferlen
+            len    = GetLongPathNameA(short_path.to_s, buffer, buffer.size)
+            # In the case of the path name being exceptionally long; resize the buffer
+            # to the actual size, then re-get it
+            if bufferlen < len
```
glennsarti: The correct way to get the buffer length is to call `GetLongPathName` with a buffer size of zero, which will then return the required buffer size.  Then you can allocate the buffer and call with the correct size.

lib/pdk/util.rb
```diff
@@ -1,8 +1,38 @@
...
 require 'tempfile'
+require 'fiddle/import' if Gem.win_platform?
 
 module PDK
   module Util
+    module WindowsFunctions
+      if Gem.win_platform?
+        extend Fiddle::Importer
+        dlload 'kernel32.dll'
+        extern('int GetLongPathNameA(char*, char*, int)', :stdcall)
```
glennsarti: The actual return type is uint32 but the ruby equivalent is int

pdk.gemspec
```diff
@@ -27,4 +27,6 @@ Gem::Specification.new do |spec|
 
   # Used in the pdk-module-template
   spec.add_runtime_dependency 'deep_merge', '~> 1.1'
+
+  spec.add_runtime_dependency 'win32-file', '~> 0.8'
```
DavidS: can we remove this now?

james-stocks:
@glennsarti This has been updated to re-use Puppet util code so we should be supporting wide strings now

lib/pdk/cli/exec.rb
```diff
@@ -135,7 +135,8 @@ def run_process!
           begin
             @process.start
           rescue ChildProcess::LaunchError => e
-            raise PDK::CLI::FatalError, _("Failed to execute '%{command}': %{message}") % { command: argv.join(' '), message: e.message }
+            raise PDK::CLI::FatalError, _('Failed to execute process: %{message}') % { message: e.message } unless @process.respond_to?(:argv)
+            raise PDK::CLI::FatalError, _("Failed to execute '%{command}': %{message}") % { command: @process.argv.join(' '), message: e.message }
```
DavidS: double raise

james-stocks: First raise is conditional. Rubocop preferred this _raise x unless y_ form. I'll see if I can re-arrange it so that it is easily readable and passes rubocop.

DavidS: ugh, yeah, the GuardClause cop is a little bit hit and miss. Maybe the better alternative would be to
```
message = if @process.respond_to?(:argv)
  # ...
raise PDK::CLI::FatalError, message
```

lib/puppet/util/windows/api_types.rb
```diff
@@ -0,0 +1,278 @@
+require 'ffi'
+require 'puppet/util/windows/string'
+
+module Puppet::Util::Windows::APITypes
```
glennsarti: Is there a plan to import the Puppet Gem into the PDK at some point?  If so, this looks like a namespacing issue.

DavidS:
Looks good to me. Since this incorporates the suggestions from @glennsarti and @Iristyle, I assume we'll be fine from that side too.

lib/puppet/util/windows/api_types.rb
DavidS: Since the PDK needs to be able to deal with multiple versions of puppet, the main binary will never reference the puppet gem.

For the record: I would prefer that utility code to live in a independent gem that others can use without having to depend on a full puppet package. But this is not relevant to this PR. In effect, we're treating this like a vendoring.

glennsarti: Ahh ok.

DavidS:
https://github.com/puppetlabs/pdk/pull/82/files#diff-f1961fa3f901b19ee585c1962e910d5eR56 and friends is the thing we need to test. We've started down this rabbit hole, because in some situations the regex at https://github.com/puppetlabs/pdk/blob/master/lib/pdk/module/templatedir.rb#L198 did not match up due to different encodings of the same path. It should possible to finagle in a test exercising this part. I'll look into it tomorrow.

DavidS:
Created SDK-286 to track the testing effort.

PR #86, opened 2017-06-15
(maint) avoid interfering with local ruby configs

DavidS:


scotje:
Is this confusing Travis/Appveyor?

DavidS:
IIRC this is used by rvm to determine which ruby to use. I'm not sure if other tools might not evaluate this too. Since we're still having troubles sorting out our dev and test environments, I'd prefer not to have such landmines around.

scotje:
Yeah, it's specifically how rbenv tracks what version it should switch to for the project, not sure if RVM adopted it as well. Let's maybe add it to .gitignore so we can still use it locally in our dev environments if needed?

scotje:
(I will push a commit to add to .gitignore)

PR #88, opened 2017-06-15
(MAINT) Create a class-based subprocess executor.

scotje:
This provides a better API for setting up sub processes than the existing class/module method based approach.

`PDK::CLI::Exec.execute` is still provided for backwards compatibility and nothing has yet been updated to directly use the new `Command` class.

### DIFF
--- a/lib/pdk/cli/exec.rb
+++ b/lib/pdk/cli/exec.rb
@@ -62,7 +62,7 @@ module PDK
           @stderr.sync = true
         end
 
-        def add_spinner(message, opts={})
+        def add_spinner(message, opts = {})
           @success_message = opts.delete(:success)
           @failure_message = opts.delete(:failure)
 
@@ -92,7 +92,7 @@ module PDK
 
           # Stop spinning when done (if configured).
           if @spinner
-            if @process.exit_code == 0 && @success_message
+            if @process.exit_code.zero? && @success_message
               @spinner.success(@success_message)
             elsif @failure_message
               @spinner.error(@failure_message)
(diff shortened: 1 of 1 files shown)


lib/pdk/cli/exec.rb
```diff
@@ -71,6 +41,108 @@ def self.try_vendored_bin(vendored_bin_path, fallback)
...
+          end
+
+          @context = new_context
+        end
+
+        def add_spinner(message, opts = {})
+          @success_message = opts.delete(:success)
+          @failure_message = opts.delete(:failure)
+
+          @spinner = TTY::Spinner.new(message, opts)
```
bmjen: Spinner message needs to be prepended with the "[:spinner]" indicator from your old implementation.

lib/pdk/cli/exec.rb
```diff
@@ -71,6 +41,108 @@ def self.try_vendored_bin(vendored_bin_path, fallback)
...
+                run_process!
+              end
+            end
+          else
+            run_process!
+          end
+
+          # Stop spinning when done (if configured).
+          if @spinner
+            if @process.exit_code.zero? && @success_message
```
bmjen: The spinner when configured will always leave at a "mid-spin" state b/c the `@success_message` or `@failure_message` variables aren't set. Do we want to process the stdout and stderr from the process for this? Or just set the spinner success/failure state with `:success` or `:failure` like you did in the implementation in `PDK::Util::Bundler`?

scotje: Yeah I guess we could still call `success` or `error` based on the exit_code.

lib/pdk/cli/exec.rb
scotje: Yeah it's probably better to enforce consistency here.

PR #89, opened 2017-06-16
(maint) release prep prep

DavidS:


lib/pdk/version.rb
```diff
@@ -1,3 +1,3 @@
 module PDK
-  VERSION = '0.1.0'.freeze
+  VERSION = '0.1.20170616'.freeze
```
bmjen: This is a nitpick, and it likely doesn't matter in pre-release but I'm not a fan of this versioning. I'd rather not see the datestamp in the Y field. Would prefer something like `0.1.0-20160616`

DavidS: Updated to `0.2.0`; not much point in having `0.` pre-release versions.

DavidS:
@scotje we can get this merged, and do the actual release whenever we feel comfortable with a minor update to the changelog.

PR #90, opened 2017-06-19
(SDK-274) Adds --version option

bmjen:
To allow easy checking and reporting of the current pdk version, {{--version}} should report the installed version of PDK, as well as the vanagon build SHA from a built package or the git ref from a local git repo.

### DIFF
--- a/lib/pdk/util/version.rb
+++ b/lib/pdk/util/version.rb
@@ -1,0 +1,34 @@
+require 'pdk/version'
+require 'pdk/cli/exec'
+
+module PDK
+  module Util
+    module Version
+      def self.version_string
+        "#{PDK::VERSION} #{pdk_ref}".strip.freeze
+      end
+
+      def self.pdk_ref
+        ref = pkg_sha || git_ref
+        ref.nil? ? nil : "(#{ref})"
+      end
+
+      def self.pkg_sha
+        version_file = File.join(File.expand_path('../../..', File.dirname(__FILE__)), 'VERSION')
+
+        if File.exist? version_file
+          ver = File.read(version_file)
+          sha = ver.strip.split('.')[-1] unless ver.nil?
+        end
+
+        sha
+      end
+
+      def self.git_ref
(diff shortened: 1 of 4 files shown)


lib/pdk/util/version.rb
```diff
@@ -0,0 +1,36 @@
...
+module PDK
+  module Util
+    module Version
+
+      def self.version_string
+        "#{PDK::VERSION} #{pdk_ref}".strip
+      end
+    
+      def self.pdk_ref
+        ref = pkg_sha || git_ref
```
bmjen: Debating flipping this prioritization. I think if someone installs the official pdk package, but modifies the underlying code (with an initialized git repo), we should be aware of that.

lib/pdk/util/version.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module PDK
+  module Util
+    module Version
+      def self.version_string
+        "#{PDK::VERSION} #{pdk_ref}".strip.freeze
+      end
+
+      def self.pdk_ref
+        ref = pkg_sha || git_ref
```
bmjen: Debating flipping this prioritization. I think if someone installs the official pdk package, but modifies the underlying code (with an initialized git repo), we should be aware of that.

scotje: In that case it would be good to know both, so maybe they should just each be conditionally added if available?

lib/pdk/util/version.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      def self.pkg_sha
+        version_file = File.join(File.expand_path('../../..', File.dirname(__FILE__)), 'VERSION')
+
+        if File.exist? version_file
+          ver = File.read(version_file)
+          sha = ver.strip.split('.')[-1] unless ver.nil?
+        end
+
+        sha
```
scotje: Seems like this is going to be an undefined var if the version file doesn't exist?

bmjen: If the version file doesn't exist, I'm expecting it to return `nil` so that we can fall back to `git_ref` up above.

scotje: Oh right, ruby hoists vars out of `if/else` blocks.

PR #91, opened 2017-06-20
(maint) Guard PDK::Util::Bundle.ensure_bundle! to only run once

rodjek:
Adds a guard into `PDK::Util::Bundle.ensure_bundle!` so that it more or less idempotent. We'll need to be calling this in multiple places when we shell out to commands inside the bundle and although this method won't actually change anything on subsequent executions as it is, it will cause the spinner to be displayed multiple times.

### DIFF
--- a/lib/pdk/util/bundler.rb
+++ b/lib/pdk/util/bundler.rb
@@ -11,6 +11,11 @@ module PDK
       def self.ensure_bundle!
         bundle = BundleHelper.new
 
+        if already_bundled?(bundle.gemfile)
+          PDK.logger.debug(_('Bundle has already been installed, skipping run'))
+          return
+        end
+
         unless bundle.gemfile?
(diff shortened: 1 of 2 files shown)


lib/pdk/util/bundler.rb
```diff
@@ -100,6 +115,10 @@ def binstubs!(gems)
           result[:exit_code].zero?
         end
 
+        def gemfile
```
DavidS: Why make this public?

DavidS:
A good trade-off between convenience and robustness - when people and processes start finagling around with details of the module while PDK is running, all bets are off.

lib/pdk/util/bundler.rb
rodjek: So that it can be called in the `ensure_bundle!` class method (outside the public scope of the BundleHelper instance)

DavidS: ah, that one always gets me.

PR #81, opened 2017-06-15
Report format implementation

rodjek:
Injests the output from the validators as JSON into `PDK::Report` and then implements `PDK::Report#to_text` and `PDK::Report#to_junit` to handle munging the outputs into the requested formats.

In an ideal world we'd use upstream junit formatting, but neither puppet-lint nor metadata-json-lint have junit formatters currently and the rubocop junit plugin seems to be abandoned. Eventually we could get this implemented upstream, but at the moment it'll be quicker to handle it ourselves.

Example output
```
cyndane :0: pdk/foo (git:junit!)$ ../bin/pdk validate
pdk (INFO): Running all available validators...
manifests/test.pp:1:8: warning: double quoted string containing no variables
Gemfile:1:1: convention: The name of this source file (`Gemfile`) should use snake_case.
Gemfile:1:29: convention: Prefer single-quoted strings when you don't need string interpolation or special symbols.
Gemfile:4:26: convention: Use `%r` around regular expression.
Gemfile:5:22: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:5:30: convention: Avoid the use of Perl-style backrefs.
Gemfile:5:34: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:5:45: convention: Avoid the use of Perl-style backrefs.
Gemfile:5:49: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:6:29: convention: Use `%r` around regular expression.
Gemfile:7:16: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:7:42: convention: Avoid the use of Perl-style backrefs.
Gemfile:7:47: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:9:26: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:14:26: convention: Use `%r` around regular expression.
Gemfile:16:9: convention: Use `String#start_with?` instead of a regex match anchored to the beginning of the string.
Gemfile:16:29: convention: Use `%r` around regular expression.
Gemfile:27:56: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:27:75: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:28:56: convention: Use the new Ruby 1.9 hash syntax.
Gemfile:29:3: convention: Gems should be sorted in an alphabetical order within their section of the Gemfile. Gem `(str "puppet-module-posix-dev-r")` should appear before `(str "puppet-module-win-default-r")`.
[...]

PR #92, opened 2017-06-20
(maint) unbreak tests after fixing pdk-module-template

DavidS:


rodjek:
Going to close this one out as the tests were fixed as part of #81

PR #94, opened 2017-06-20
(SDK-137) Adds Puppet Parser syntax validation

bmjen:
Adds validation for puppet syntax using `puppet parser validate`.

This PR also includes various cleanups and refactoring of the validation handlers.

### DIFF
--- a/lib/pdk/validators/base_validator.rb
+++ b/lib/pdk/validators/base_validator.rb
@@ -38,7 +38,7 @@ module PDK
         _('Invoking %{cmd}') % { cmd: cmd }
       end
 
-      def self.invoke(options = {})
+      def self.invoke(report, options = {})
         PDK::Util::Bundler.ensure_binstubs!(cmd)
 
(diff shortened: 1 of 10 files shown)


lib/pdk/validators/base_validator.rb
```diff
@@ -16,7 +20,8 @@ def self.parse_targets(options)
         targets.map { |target|
           if respond_to?(:pattern)
             if File.directory?(target)
-              Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              files_glob = Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              files_glob.flatten.empty? ? target : files_glob
```
rodjek: Why return target if there are no files that match the pattern?

bmjen: `puppet-lint` and `puppet parser validate` require a path input to work. The alternative is to not call them at all if no files match.

PR #96, opened 2017-06-21
(SDK-245) Add acceptance tests for the output of the ruby validator

rodjek:
Adds tests for both `text` and `junit` formats.

It's hard to find an official spec for the JUnit XML format as everyone who implements it releases their own different version (except for the Apache Foundation, of course, who didn't release one), but I believe the one I've found the original XSD. In order to comply with the XSD, I've had to make a few minor changes to our JUnit formatter to implement attributes and elements that the docs I was originally following had marked as optional.

I ended up implementing my own RSpec matcher for XML as the best existing gem I could find (rspec-xml) hasn't been updated for RSpec 3 and throws deprecation notices around when using the matchers it provides.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ end
 group :test do
   gem 'rake', '~> 10.0'
   gem 'rspec', '~> 3.0'
+  gem 'rspec-xsd'
   gem 'rubocop', '= 0.49.1'
   gem 'rubocop-rspec', '= 1.15.1'
 end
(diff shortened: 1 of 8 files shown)


DavidS:
rebased on top of current master.

DavidS:
@rodjek while this is literally an "enhancement", I do not think that users care too much about being told about added tests in the changelog, therefore I untagged these two PRs

PR #98, opened 2017-06-21
(SDK-147) Add 'test unit' runner and basic output formatting

scotje:


lib/pdk/report/event.rb
```diff
@@ -270,6 +275,28 @@ def sanitise_column(value)
...
+      #
+      # @param value [Array] Array of stack trace lines
+      #
+      # @return [Array] Array of stack trace lines with less relevant lines excluded
+      def sanitise_trace(value)
+        return nil if value.nil?
+
+        valid_types = [Array]
+
+        unless valid_types.include?(value.class)
```
DavidS: That seems a bit excessive for a single class test :->

lib/pdk/report.rb
```diff
@@ -78,7 +78,7 @@ def to_junit(target = self.class.default_target)
     def to_text(target = self.class.default_target)
       events.each do |_tool, tool_events|
         tool_events.each do |event|
-          target.puts("\n" << event.to_text) unless event.pass?
+          target.puts(event.to_text) unless event.pass?
```
DavidS: This just undoes a change from the previous commit. If you `rebase --interactive`, and remove the change from the first commit, you can get rid of it here too.

lib/pdk/cli/exec.rb
```diff
@@ -55,6 +55,7 @@ class Command
         attr_reader :argv
         attr_reader :context
         attr_accessor :timeout
+        attr_accessor :environment
```
DavidS: 👍 on the change itself, it does feel like it could be a separate commit though, as it adds a feature unrelated to the "puppetlabs_spec_helper invocation for unit tests" mentioned in the commit message.

scotje:
Yes I'll try to rebase this into a saner set of commits before merge.

scotje:
@DavidS @rodjek @james-stocks This is ready to go now. Reviewing commit-by-commit should be pretty clear.

Once merged, James should be able to rebase his `--list` PR on top and hopefully most of the test issues will be resolved.

lib/pdk/tests/unit.rb
```diff
@@ -6,20 +6,89 @@ module PDK
...
+            )
+          end
+        else
+          raise PDK::CLI::FatalError, _('Unable to parse results of unit test tool invocation: %{results}') % { results: json_data.to_s }
+        end
+
+        return unless json_data['summary']
+
+        # TODO: standardize summary output
+        $stderr.puts '  ' << _('Evaluated %{total} tests in %{duration} seconds: %{errors} errors, %{failures} failures, %{pending} pending') % {
```
bmjen: Shouldn't this summary line be added to the report to be printed out at the end?

scotje: The report may go to a file or be in XML or all kinds of things. I consider the summary to be diagnostic information which is why I write it to `$stderr` once the run is finished.

scotje: (Note also that a report generator could create it's own summary when appropriate based on the events we added when parsing the output.)

lib/pdk/tests/unit.rb
```diff
@@ -6,20 +6,89 @@ module PDK
...
+
+        result[:exit_code]
+      end
+
+      def self.parse_output(report, json_data)
+        if json_data['messages'] && !json_data['messages'].empty?
+          json_data['messages'].each do |msg|
+            report.add_event(
+              source: 'rspec',
+              state: :error,
```
james-stocks: I was looking at the rspec json output yesterday, and I was unsure if a message from rspec is necessarily an error; a quick browse of the rspec-core code suggests they're pretty neutral on what a _message_ is.
I think it's a fair assumption that a message from rspec is something we can consider an error.

lib/pdk/tests/unit.rb
```diff
@@ -6,20 +6,89 @@ module PDK
...
+              message: msg,
+            )
+          end
+        elsif json_data['summary'] && json_data['summary']['failure_count'].zero?
+          report.add_event(state: :passed, severity: :ok)
+        elsif json_data['examples']
+          failures = json_data['examples'].reject { |e| e['status'] == 'passed' }
```
james-stocks: Should pending also be rejected here?

lib/pdk/tests/unit.rb
```diff
@@ -6,20 +6,89 @@ module PDK
...
+        if json_data['messages'] && !json_data['messages'].empty?
+          json_data['messages'].each do |msg|
+            report.add_event(
+              source: 'rspec',
+              state: :error,
+              message: msg,
+            )
+          end
+        elsif json_data['summary'] && json_data['summary']['failure_count'].zero?
+          report.add_event(state: :passed, severity: :ok)
```
james-stocks: Is this all the report includes in the case of all examples passing? I guess the details of each example are not needed if everything is passing; but wouldn't it be desirable to record at least the number of examples that passed?

scotje:
The JSON output from RSpec in general leaves a lot to be desired, we might want to consider writing a our own custom reporter eventually and shipping it as part of the template.

I'll see what I can do to improve things with what we have though.

scotje:
Updated and simplified the report parsing.

Currently some of the information (such as context) is being passed to the reporter but the `to_text` method of the `Report::Event` class could use improvement. That seems like a separate PR unless there is something this code isn't passing to the reporter that it should be.

lib/pdk/tests/unit.rb
```diff
@@ -6,20 +6,95 @@ module PDK
...
+        }
+
+        json_data['examples'] && json_data['examples'].each do |ex|
+          example_results[ex['status']] << ex if example_results.key?(ex['status'])
+        end
+
+        example_results.each do |result, examples|
+          state = case result
+                  when 'passed' then :passed
+                  when 'failed' then :failure
```
james-stocks: It looks like at this point we immediately translate rspec result values to the result values our CI uses

I think our report code ( https://github.com/puppetlabs/pdk/blob/master/lib/pdk/report/event.rb#L83 ) should be clearly documented to explain that rspec pending gets mapped to a skipped result in XML reports; I'll update that outside of this PR

lib/pdk/tests/unit.rb
```diff
@@ -6,20 +6,96 @@ module PDK
...
+          'pending' => [],
+        }
+
+        json_data['examples'] && json_data['examples'].each do |ex|
+          example_results[ex['status']] << ex if example_results.key?(ex['status'])
+        end
+
+        example_results.each do |result, examples|
+          # Translate rspec example results to JUnit XML testcase results
+          state = case result
```
james-stocks: FYI I created https://github.com/puppetlabs/pdk/pull/115 (now merged) to support pending/skipped results better

PR #102, opened 2017-06-22
(TEST) Does appveyor fail as required?

james-stocks:


james-stocks:
Closing. If anyone's interested, there's an issue here where pdk invokes 'pwd' as a system command; which is not valid on Windows, but is valid on appveyor (maybe it has cygwin binaries on the path?). It's not a real issue because the pdk code in question is placeholder (https://github.com/puppetlabs/pdk/blob/d1d7fc48e4bc5983d878e86507d750f156374509/lib/pdk/tests/unit.rb#L12)

PR #103, opened 2017-06-23
(SDK-284) Add guidance for users during new module interview

rodjek:
Refactors the module metadata interview to use [TTY::Prompt](https://github.com/piotrmurach/tty-prompt). This provides a much nicer user experience when an invalid response is given in that it will blank and reprint the question on the same line rather than reprinting the question on a new line.

Additionally, the questions are now numbered (automatically) and can be specified with some help text to provide guidance to the users. The interview definition has also been separated from the metadata update process in order to make it easier to update the questions in the future.


scotje:
Love this is general, it does have some slightly sad times on Windows when it repositions the cursor, which is a similar issue to tty-spinner which hopefully will be resolved by `ansicon`:

```
PS C:\Users\Administrator> ruby C:\Users\Administrator\src\pdk\bin\pdk new module test_prompt
pdk (INFO): Creating new module: test_prompt

We need to create a metadata.json file for this module, so we're going to ask you 8 quick questions.
If the question is not applicable to this module, just leave the answer blank.

[Q 1/8]
What is your Puppet Forge username?
This will be used when uploading your module to the Forge. You can opt out of this at any time.
←[2K←[1G←[1A←[2K←[1G--> administrator

[Q 2/8]
What version is this module?
Puppet uses Semantic Versioning (semver.org) to version modules.
←[1A←[2K←[1G--> 0.1.0

[Q 3/8]
Who wrote this module?
The person who gets credit for creating the module.
--> (Administrator)
```

In the meantime perhaps we can add a Windows fallback UI? (Or we can just prioritize getting `ansicon` into the packaging/system requirements.)

lib/pdk/generators/module.rb
```diff
@@ -77,63 +78,98 @@ def self.prepare_module_directory(target_dir)
...
-          retry
-        end
+        questions = [
+          {
+            name:             'name',
+            question:         _('What is your Puppet Forge username?'),
+            help:             _('This will be used when uploading your module to the Forge. You can opt out of this at any time.'),
+            required:         true,
+            validate_pattern: %r{\A[a-z0-9]+\Z},
+            validate_message: _('Forge usernames can only contain letters and numbers'),
```
scotje: **lowercase** letters

scotje: (Technically the Forge will accept uppercase but I think we want to strongly discourage that, we may even coerce new usernames to all lowercase.)

rodjek: Yeah, I just stole that pattern from `PDK::Module::Metadata`'s name validation :)

DavidS:
I've spent a bit of time today to explore our options. Turns out that full VT100 emulation is supported in the windows console. Starting from Windows 10 TH2 :-/ See https://stackoverflow.com/a/44482740 for details.

As a fallback, @scotje (as well as the SO answer linked above) suggest ansicon, which works fine in my testing. I'll ticket ansicon integration separately. I expect that packaging to be straightforward, so we can merge this in expectation that we'll have ansicon available RSN?

DavidS:
not my day. the additional commit should have gont to #108

scotje:
Yes, the very latest Windows 10 has control sequence support but you do have to opt into it still. :(

DavidS:
https://github.com/DavidS/pdk/commit/6ec007e733f59447df8eee346835366308eb8886 has a prototype for supporting Windows' VT100 support. Since this support is so new, we can't depend on it, and need to investigate/implement alternative routes, like [ansicon](https://github.com/adoxa/ansicon/), see SDK-293.

@scotje as this is primarily an aesthetic problem, can we merge this, and revisit a local fallback here vs. using ansicon, in the next few days when we understand better the costs of ansicon?

scotje:
@DavidS (shrug) We've gone out of our way to provide a fallback for the spinner stuff but I understand the prompt is more complicated. I basically didn't want to assume we would get `ansicon` (or equivalent) into place before release and it kind of gives a "didn't test on windows" vibe to users to have the control sequences bleeding into the output...

But if we want to just commit to control sequence support on Windows sooner rather than later, I'm fine skipping any fallback for now in favor of `ansicon` work or whatever.

DavidS:
I need to do the sprint demo tonight, and would love to ship a new code drop before my PTO next week, so I'll hack on this, and the ansicon stuff today.

PR #104, opened 2017-06-23
(SDK-285) Add --auto-correct flag to validators that support it

rodjek:
This will just silently no-op for validators that don't support auto correction. There's no acceptance tests for puppet-lint integration at the moment, so I've only added an acceptance test for auto-correcting rubocop violations.

### DIFF
--- a/spec/acceptance/support/in_a_new_module.rb
+++ b/spec/acceptance/support/in_a_new_module.rb
@@ -1,7 +1,11 @@
+require 'open3'
 
 shared_context 'in a new module' do |name|
   before(:all) do
-    system("pdk new module #{name} --skip-interview") || raise
+    output, status = Open3.capture2e('pdk', 'new', 'module', name, '--skip-interview')
+
+    raise "Failed to create test module:\n#{output}" unless status.success?
+
     Dir.chdir(name)
   end
 
--- a/README.md
+++ b/README.md
@@ -58,6 +58,22 @@ Running validations on `new_module`:
 [...]
 ```
 
+Some validators support automatic correction of detected problems (for example,
+both rubocop and puppet-lint can automatically correct many common code style
+problems). To enable this functionality, run `pdk validate` with the
+`--auto-correct` option.
+
+```
+$ pdk validate --auto-correct
+pdk (INFO): Running all available validators...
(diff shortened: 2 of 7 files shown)


spec/acceptance/support/in_a_new_module.rb
```diff
@@ -1,7 +1,8 @@
 
 shared_context 'in a new module' do |name|
   before(:all) do
-    system("pdk new module #{name} --skip-interview") || raise
+    output = `pdk new module #{name} --skip-interview`
```
james-stocks: I'm not sure this captures stderr, appending 2>&1 should work on all platforms

spec/acceptance/support/in_a_new_module.rb
```diff
@@ -1,7 +1,8 @@
 
 shared_context 'in a new module' do |name|
   before(:all) do
-    system("pdk new module #{name} --skip-interview") || raise
+    output = `pdk new module #{name} --skip-interview 2>&1`
+    raise "Failed to create test module:\n#{output}" unless File.directory?(name)
```
DavidS: Then you can also `raise` here on the exit status returned from capture2e, and not rely on environmental factors.

DavidS:
Please also add a note to the README about how and where to use this flag.

rodjek:
> Even if there are no puppet-lint acceptance tests yet, please do start them when you're adding to it. The lack of testing over there, won't get better by ignoring it!

Sorry, didn't mean to imply that I was ignoring the lack of tests. I knew that puppet-lint integration was still a work in progress with open tickets and didn't want to duplicate work.

rodjek:
Updated the `in a new module` context to use `Open3.capture2e`. Will open a new PR to add the basic puppet-lint acceptance tests and then loop back to this PR to add acceptance tests for puppet-lint auto-correct here so that we keep this PR on topic.

rodjek:
Acceptance tests for puppet-lint integration added in #109. Once they are merged in, I'll come back to this and add acceptance tests for puppet-lint with --auto-correct.

rodjek:
@DavidS acceptance tests added for auto-correcting puppet-lint problems

PR #99, opened 2017-06-22
(SDK-256) Acceptance tests for metadata validator behavior and output

rodjek:

PR #105, opened 2017-06-23
(SDK-137) Add puppet syntax validation

bmjen:
This feature requires this fix https://github.com/voxpupuli/puppet-syntax/pull/83 to be published.


lib/pdk/report.rb
```diff
@@ -85,6 +85,9 @@ def to_junit(target = self.class.default_target)
     # @param target [#write] an IO object that the report will be written to.
     #   Defaults to PDK::Report.default_target.
     def to_text(target = self.class.default_target)
+      # Extra defaulting here, b/c the Class.send method will pass in nil
+      target ||= self.class.default_target
```
bmjen: Good idea. And the formatting happens in the `to_text` method of the `Event` class.

lib/pdk/validators/puppet/puppet_syntax.rb
```diff
@@ -0,0 +1,61 @@
...
+          }
+          inputs[:severity] = severity.strip unless severity.nil?
+          inputs[:line] = line.strip unless line.nil?
+          inputs[:column] = column.strip unless column.nil?
+          report.add_event(inputs)
+        end
+      end
+
+      def self.sanitize_console_output(line)
+        line.gsub!(%r{\e\[([;\d]+)?m}, '')
```
bmjen: I wouldn't want to change the default behavior of `puppet-syntax` itself and strip the colors, since users of this tool outside of PDK may still want the distinct console colors. I could add a Rakefile option to disable the colors in puppet-syntax and toggle it on or off based on that.

DavidS: good enough for now.

DavidS:
Still needs a rebase :-/

bmjen:
I'll rebase after I get these acceptance tests done. Conflicts came in after #99 and #109 were merged.

DavidS:
👍

rodjek:
Made a couple of changes to fix up the acceptance tests on Windows:
 * Changed how the location information is parsed in the puppet-syntax validator to handle file path that have colons in them
 * Changed how we spawn ruby processes on windows to disable warnings generated by `Kernel.warn`. I don't like this change but unfortunately `readline` on Ruby 2.1 on Windows includes a deprecated library and Puppet requires `readline`. We patch our vendored build of Ruby 2.1 to remove the deprecation warning (see PUP-3060) but as we're dealing with system ruby here... Yeah.

rodjek:
An alternative to disabling warnings would be to remove matching known problematic deprecation warnings and removing them from the output before parsing, but that seems like a brittle solution. Group consensus on the least awful solution? :)

DavidS:
Filtering the known warnings for test purposes would be preferable, as it would allow us to detect new warnings on incoming code before it hits master.

No reason to block this PR for it though.

PR #107, opened 2017-06-26
(SDK-148) Implement pdk test unit --list

james-stocks:
Implements the _pdk test unit --list_ option that outputs a list of the module's spec tests

### DIFF
--- a/lib/pdk/cli/test/unit.rb
+++ b/lib/pdk/cli/test/unit.rb
@@ -24,27 +24,34 @@ module PDK::CLI
       report = nil
 
       if opts[:list]
-        puts _('List of all available unit tests: (TODO)')
-        exit 0
+        examples = PDK::Test::Unit.list
+        if examples.empty?
+          puts _('No examples found.')
+        else
+          puts _('Examples:')
+          examples.each do |example|
+            puts _("%{id}\t%{description}" % { id: example[:id], description: example[:full_description] })
+          end
+        end
+      else
+        report = PDK::Report.new
+        report_formats = if opts[:format]
+                           PDK::CLI::Util::OptionNormalizer.report_formats(opts[:format])
+                         else
+                           [{
+                             method: PDK::Report.default_format,
+                             target: PDK::Report.default_target,
+                           }]
+                         end
+
+        exit_code = PDK::Test::Unit.invoke(report, opts)
+
+        report_formats.each do |format|
+          report.send(format[:method], format[:target])
+        end
+
+        exit exit_code
       end
-
-      report = PDK::Report.new
-      report_formats = if opts[:format]
-                         PDK::CLI::Util::OptionNormalizer.report_formats(opts[:format])
-                       else
-                         [{
-                           method: PDK::Report.default_format,
-                           target: PDK::Report.default_target,
-                         }]
-                       end
-
-      exit_code = PDK::Test::Unit.invoke(report, opts)
-
-      report_formats.each do |format|
-        report.send(format[:method], format[:target])
-      end
-
-      exit exit_code
     end
   end
 end
--- a/lib/pdk/tests/unit.rb
+++ b/lib/pdk/tests/unit.rb
@@ -1,6 +1,7 @@
 require 'pdk'
 require 'pdk/cli/exec'
 require 'pdk/util/bundler'
+require 'json'
 
 module PDK
   module Test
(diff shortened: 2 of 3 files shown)


lib/pdk/tests/unit.rb
```diff
@@ -21,6 +23,26 @@ def self.invoke(tests, report = nil)
...
+
+      def self.list
+        command_argv = ['rspec', '--dry-run', '-fjson']
+        list_command = PDK::CLI::Exec::Command.new(*command_argv)
+        list_command.context = :module
+        output = list_command.execute!
+        rspec_json_output = JSON.parse(output[:stdout])
+        if rspec_json_output['examples'].empty?
+          rspec_message = rspec_json_output['messages'][0]
+          raise PDK::CLI::FatalError, _('Unable to enumerate examples. rspec reported: %{message}' % { message: rspec_message }) unless rspec_message == 'No examples found.'
```
james-stocks: I'm not a fan of the 'guard clause' that rubocop pushes me towards here - the 'if' clause of this raise is off the edge of the screen :/

james-stocks:
Appveyor is failing where the module spec tests require rspec-puppet-facts 

Repeating locally/manually; I get a bundle install failure for a generated module - I'm now trying to confirm if this is the same issue or something different

scotje:
@james-stocks the issue is that the windows puppet-module meta gems currently don't depend on `rspec-puppet-facts` which is the same reason my unit test PR is currently failing on windows.

lib/pdk/cli/test/unit.rb
```diff
@@ -20,20 +20,20 @@ module PDK::CLI
       report = nil
 
       if opts[:list]
-        puts _('List of all available unit tests: (TODO)')
+        PDK::Test::Unit.list
```
rodjek: I would recommend moving the output logic from `PDK::Test::Unit.list` and into here. If it just returns an array of test names and leaves it to the CLI to print them out for the user, it can be potentially be utilised in other places in the future (e.g. validating the contents of `opts[:tests]` before handing off to rspec).

lib/pdk/tests/unit.rb
```diff
@@ -1,6 +1,8 @@
 require 'pdk'
 require 'pdk/cli/exec'
 require 'pdk/util/bundler'
+require 'rspec/core'
```
rodjek: Does this need to be required?

lib/pdk/tests/unit.rb
```diff
@@ -21,6 +23,32 @@ def self.invoke(tests, report = nil)
         output = PDK::CLI::Exec.execute(cmd(tests))
         report.write(output) if report
       end
+
+      def self.list
+        PDK::Util::Bundler.ensure_bundle!
+        PDK::Util::Bundler.ensure_binstubs!('rspec-core')
+        command_argv = [File.join(PDK::Util.module_root, 'bin', 'rspec'), '--dry-run', '-fjson']
```
rodjek: My preference is to use the more verbose forms of options in scripts (`--format json` instead of `-fjson`) as it makes the intentions clearer.

lib/pdk/tests/unit.rb
```diff
@@ -21,6 +23,32 @@ def self.invoke(tests, report = nil)
         output = PDK::CLI::Exec.execute(cmd(tests))
         report.write(output) if report
       end
+
+      def self.list
```
rodjek: A general stylistic suggestion for this method: Split the code up into logical paragraphs with a newline between them (e.g. paragraph 1 - bundler stuff, paragraph 2 - building the command array, paragraph 3 - executing the command, paragraph 4 - processing the output), makes it easier to grok a chunk at a time :)

spec/acceptance/test_spec.rb
```diff
@@ -17,5 +17,64 @@
...
+    end
+  end
+end
+EOF
+        end
+      end
+
+      describe command('pdk test unit --list') do
+        its(:exit_status) { is_expected.to eq 0 }
+        its(:stdout) { is_expected.to match(%r{Examples:.*demo_spec.rb\[1:1:1\].*demo_spec.rb\[1:2:1\].*demo_spec.rb\[1:3:1\]}m) }
```
rodjek: Rather than using a big multiline regex, I'd suggest doing multiple matches on stdout

```ruby
its(:stdout) { is_expected.to match(%r{^demo_spec\.rb\[1:1:1\]}) }
its(:stdout) { is_expected.to match(%r{^demo_spec\.rb\[1:2:1\]}) }
its(:stdout) { is_expected.to match(%r{^demo_spec\.rb\[1:3:1\]}) }
```

DavidS:
Rekicking appveyor after the new puppet-module-gems helped.

spec/acceptance/test_spec.rb
james-stocks: It's a single expectation that we match all three at the same time; though I guess multiple matches like you suggest is more readable

lib/pdk/tests/unit.rb
james-stocks: Have split with newlines

james-stocks:
@rodjek Have addressed your comments, thanks

lib/pdk/tests/unit.rb
```diff
@@ -21,6 +22,36 @@ def self.invoke(tests, report = nil)
...
+        command_argv = [File.join(PDK::Util.module_root, 'bin', 'rspec'), '--dry-run', '--format', 'json']
+        command_argv.unshift('ruby') if Gem.win_platform?
+        list_command = PDK::CLI::Exec::Command.new(*command_argv)
+        list_command.context = :module
+        output = list_command.execute!
+
+        rspec_json_output = JSON.parse(output[:stdout])
+        if rspec_json_output['examples'].empty?
+          rspec_message = rspec_json_output['messages'][0]
+          if rspec_message == 'No examples found.' # rubocop:disable Style/GuardClause
```
rodjek: You can get rid of this rubocop disable by rewriting this conditional as a suffix to the return statement and letting it fall through to the raise if it doesn't match

```
return [] if rspec_message == 'No examples found.'

raise PDK::CLI::FatalError, _(...)
```

DavidS:
Rebased, and merged the test suite into the now existing `pdk test unit` tests.

DavidS:
@james-stocks feel free to merge, if you're satisfied with my changes.

PR #108, opened 2017-06-27
(SDK-290) Make sure that all usernames are processed when creating a new module

austb:


rodjek:
Long term, I think we'll want to save the answers to the interview questions (where they're not specific to the module) into a file and read those in as the default values, so that users only have to provide forge username, preferred license etc once. For now though, I think aborting and asking the user to run without `--skip-interview` is a good way to go.

DavidS:
I've added some pending unit tests that show the behaviour I'd like to see.

The failing acceptance tests are from a hidden merge conflict to master. Please rebase this, and address those issues.

lib/pdk/generators/module.rb
```diff
@@ -51,11 +51,10 @@ def self.invoke(opts = {})
       end
 
       def self.prepare_metadata(opts)
-        username = Etc.getlogin
-        if username != username.gsub(%r{[^0-9a-z]}i, '')
-          raise PDK::CLI::FatalError, _('Unable to select a valid Forge username, run without --skip-interview to enter one') if opts[:'skip-interview']
-
-          username = ''
+        username = Etc.getlogin.gsub(%r{[^0-9a-z]}i, '')
+        username = 'username' if username == ''
```
austb: This protects against an entirely non-alphanumeric username

PR #113, opened 2017-06-28
(maint) Make binstub generation quiet unless it fails

rodjek:
Currently, the output of the validators is overly noisy. Each validator needs to ensure the presence of its appropriate binstub, which prints a "Checking for required Bundler binstubs" spinner for each validator.

```
$ ../bin/pdk validate
pdk (INFO): Running all available validators...
[✔] Checking for missing Gemfile dependencies
[✔] Checking for required Bundler binstubs
[✔] Checking metadata.json
[✔] Checking for required Bundler binstubs
[✔] Checking Puppet manifest style
[✔] Checking for required Bundler binstubs
[✔] Checking Puppet manifest syntax
[✔] Checking for required Bundler binstubs
[✔] Checking Ruby code style
```

This change instead makes the binstub generation silent unless it fails, as binstub generation is quick and is an implementation detail that the user shouldn't have to care about.

```
$ ../bin/pdk validate
pdk (INFO): Running all available validators...
[✔] Checking for missing Gemfile dependencies
[✔] Checking metadata.json
[✔] Checking Puppet manifest style
[✔] Checking Puppet manifest syntax
[✔] Checking Ruby code style
```

### DIFF
--- a/lib/pdk/cli/exec.rb
+++ b/lib/pdk/cli/exec.rb
@@ -149,15 +149,12 @@ module PDK
         protected
 
         def run_process!
+          command_string = argv.join(' ')
+          PDK.logger.debug(_("Executing '%{command}'") % { command: command_string })
           begin
             @process.start
           rescue ChildProcess::LaunchError => e
-            msg = if @process.respond_to?(:argv)
-                    _("Failed to execute '%{command}': %{message}") % { command: @process.argv.join(' '), message: e.message }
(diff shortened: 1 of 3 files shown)


DavidS:
Amended this with adding logging in the core executor, so users still have the ability to understand what's going on under the hood.

rodjek:
:+1:

PR #115, opened 2017-06-28
(maint) Support for skipped/pending tests

james-stocks:
Documents that skipped tests will include rspec pending examples.
Adds skipped tests to our XML reports
Updates the test fixture XML schema to allow for skipped tests

### DIFF
--- a/lib/pdk/report.rb
+++ b/lib/pdk/report.rb
@@ -59,6 +59,7 @@ module PDK
         testsuite.attributes['tests'] = testcases.length
         testsuite.attributes['errors'] = testcases.select(&:error?).length
         testsuite.attributes['failures'] = testcases.select(&:failure?).length
+        testsuite.attributes['skipped'] = testcases.select(&:skipped?).length
         testsuite.attributes['time'] = 0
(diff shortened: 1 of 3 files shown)


rodjek:
This seems like a good change to me. This schema (without these changes) is the same one used by the most popular rspec plugin as a reference. That plugin also uses the `skipped` element, so I guess it doesn't pass the XSD validation either 😆. Despite that, I don't see any issues on that repo where the skipped element is causing any problems, so i'm :+1: to adding it to the XSD.

DavidS:
Some more spelunking. The current XMLJUnitFormatter code from ANT: https://git-wip-us.apache.org/repos/asf?p=ant.git;a=blob;f=src/main/org/apache/tools/ant/taskdefs/optional/junit/XMLJUnitResultFormatter.java;h=d62a847d490b3b3acbe8fa0ace23a104081c5679;hb=HEAD ; a community contributed XSD for junit4's output: https://gist.github.com/kuzuha/232902acab1344d6b578 ; 

I wasn't able to locate the junit-plugins actual parsing code in https://github.com/jenkinsci/junit-plugin/, but there is ample material referencing the "skipped" properties

PR #121, opened 2017-06-29
(SDK-294) Avoid module name conflict in acceptance tests.

james-stocks:
SDK-294 is a bug where one acceptance test fails to create 'foo' as a module folder, in spite of the previous acceptance test apparently deleting it.

This commit does not directly fix that issue; but instead has acceptance tests use unique folder names instead of all using 'foo'


DavidS:
PS: labelling this as a bug makes it appear in the CHANGELOG, is this a change that users need to be aware of?

PR #124, opened 2017-06-30
(SDK-296) Allow target selection for the metadata validator

rodjek:
In order to accommodate this functionality, I needed to refactor the base validator a bit so that we can optionally invoke the validator once per target (as metadata-json-lint only supports testing a single file).

### DIFF
--- a/lib/pdk/validators/base_validator.rb
+++ b/lib/pdk/validators/base_validator.rb
@@ -42,7 +42,7 @@ module PDK
         targets
       end
 
-      def self.spinner_text
+      def self.spinner_text(_targets = nil)
         _('Invoking %{cmd}') % { cmd: cmd }
       end
 
@@ -67,7 +67,7 @@ module PDK
 
           command = PDK::CLI::Exec::Command.new(*cmd_argv).tap do |c|
             c.context = :module
-            c.add_spinner(spinner_text)
+            c.add_spinner(spinner_text(invokation_targets))
           end
 
           result = command.execute!
--- a/lib/pdk/validators/base_validator.rb
+++ b/lib/pdk/validators/base_validator.rb
@@ -4,6 +4,14 @@ require 'pdk/cli/exec'
 module PDK
   module Validate
     class BaseValidator
+      # Controls how many times the validator is invoked.
+      #
+      #   :once -       The validator will be invoked once and passed all the
+      #                 targets.
+      #   :per_target - The validator will be invoked for each target
+      #                 separately.
+      INVOKE_STYLE = :once
+
       def self.cmd_path
         File.join(PDK::Util.module_root, 'bin', cmd)
       end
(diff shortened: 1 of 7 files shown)


spec/acceptance/validate_metadata_spec.rb
```diff
@@ -70,4 +70,71 @@
       end
     end
   end
+
+  context 'when validating a specific file' do
+    include_context 'in a new module', 'foo'
```
james-stocks: Due to SDK-294 (an intermittent bug when re-using the foo folder on Windows) it's preferable to give each feature its own test module name instead of always using 'foo'

lib/pdk/validators/base_validator.rb
```diff
@@ -44,19 +46,26 @@ def self.invoke(report, options = {})
...
 
         PDK::Util::Bundler.ensure_binstubs!(cmd)
-        cmd_argv = parse_options(options, targets).unshift(cmd_path)
-        cmd_argv.unshift('ruby', '-W0') if Gem.win_platform?
 
-        command = PDK::CLI::Exec::Command.new(*cmd_argv).tap do |c|
-          c.context = :module
-          c.add_spinner(spinner_text)
-        end
+        targets = (self::INVOKE_STYLE == :per_target) ? targets.combination(1).to_a : Array[targets]
```
james-stocks: I don't follow how the invoke style actually changes here. For a validation of _*.json_ , I think the two possible outcomes for _targets_ here are:

    [['a.json'],['b.json'],['c.json']]

or

    ['a.json', 'b.json', 'c.json']

But in the code below, in both cases, _targets.each_ just iterates on each of the 3 in turn?

lib/pdk/validators/base_validator.rb
```diff
@@ -4,6 +4,8 @@
 module PDK
   module Validate
     class BaseValidator
+      INVOKE_STYLE = :once
```
james-stocks: I'm unsure if the intention of _:once_ is immediately understandable - might _:combined_ or _:all_targets_ be more immediately understandable?
Or leave it as _:once_ but add some commenting here on the possible values of INVOKE_STYLE

rodjek: :+1: That constant definitely needs some documentation.

lib/pdk/validators/base_validator.rb
rodjek: Yeah, this is to try and avoid duplicating the logic below when handling the different invoke styles. If `:per_target`, we take the array of targets and using `Array#combination`, split it into a 2 dimensional array, where each element inside `targets` is itself an array of 1 target.

Given
```ruby
targets = [
  'target1',
  'target2',
  'target3',
]
```
If the `INVOKE_STYLE` is `:per_target`, the `targets` array becomes `[['target1'], ['target2'], ['target3']]`. If the `INVOKE_STYLE` is `:once`, the `targets` array becomes `[['target1', 'target2', 'target3']]`.

By doing it this way, we ensure that `targets` will always be an array of arrays and just loop through it and avoid having to put a conditional around the execution logic to determine if how many times the validated should be invoked and with which targets.

james-stocks: :+1:

lib/pdk/validators/base_validator.rb
```diff
@@ -4,6 +4,14 @@
 module PDK
   module Validate
     class BaseValidator
+      # Controls how many times the validator is invoked.
```
james-stocks: 👍

PR #132, opened 2017-07-04
(SDK-305) Answer file to cache module interview answers, template-url etc

rodjek:
Adds the `PDK::AnswerFile` class to handle reading and saving user-supplied answers to questions.

The responses from the module interview for the following questions are saved and used as the default values for new modules:
 * Forge username
 * Author name
 * License

If the license is specified on the command line with `--license`, then that value is saved to the answer file as well.

Additionally, if the user specifies their own module template using `--template-url`, that value is also saved and used automatically for any subsequent modules.

I had to add a hidden option to the CLI (`--answer-file`) in order to run the acceptance tests with clean answer files. Not sure if we want to make this option visible or not (my gut reaction is to keep it hidden unless there is a need for it).

### DIFF
--- a/lib/pdk.rb
+++ b/lib/pdk.rb
@@ -1,3 +1,4 @@
+require 'pdk/answer_file'
 require 'pdk/generate'
 require 'pdk/i18n'
 require 'pdk/logger'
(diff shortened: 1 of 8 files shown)


scotje:
This still WIP @rodjek ?

rodjek:
@scotje not anymore :)

scotje:
Hahah, OK I'll take a look tomorrow then. :)

PR #139, opened 2017-07-10
(maint) Finish PDK::Validate::* unit tests

rodjek:


lib/pdk/validators/ruby/rubocop.rb
```diff
@@ -16,6 +16,10 @@ def self.cmd
         'rubocop'
       end
 
+      def self.pattern
+        ['**/*.rb']
```
bmjen: I recall we had a brief discussion about how rubocop should lint anything that has ruby in it, including Gemfile and Rakefile. Adding a pattern will ignore those files now, is that desired behavior?

rodjek: That's a good point (although our `.rubocop.yml` template in puppetlabs/pdk-module-template is explicitly excluding `Gemfile` and `Rakefile` at the moment, but we can fix that later). I'll remove the pattern and update the spec

PR #141, opened 2017-07-11
(maint) Fixes for acceptance tests against packages

james-stocks:
Ensure that acceptance tests against a package install execute the installed pdk and NOT pdk from source

Also updates the beaker pre-suite to use the new install path _/opt/puppetlabs/pdk_ instead of _sdk_


james-stocks:
Closing, suspect this wasn't opened correctly.

PR #144, opened 2017-07-11
(SDK-144) Add option to run validate in parallel

austb:
This addes the `--parallel` flag to the `pdk validate` command. This
flag will cause the validations to run in their own threads.

**UPDATE:** waiting for release of `tty-spinner` gem (also probably need to wait for a release of `tty-prompt` with compatible `tty-cursor` range.

Remaining

- [x] Remove short_spinner_text
- [x] Use PDK::Util.spinner_opts_for_platform
- [x] Use released versions of `tty-spinner` and `tty-prompt` gems
- [ ] Add Docs


lib/pdk/cli/validate.rb
```diff
@@ -15,6 +15,7 @@ module PDK::CLI
 
     flag nil, :list, _('list all available validators')
     flag :a, 'auto-correct', _('automatically correct problems (where possible)')
+    flag nil, :parallel, _('run validations in parallel')
```
bmjen: For the upcoming community release, we'll need to comment this line and hardcode line 74 with `false`.

rodjek: Not necessary to comment it out, can just make it a hidden flag (like --answer-file), so that it's still available for dev & testing but not shown in the help output.

lib/pdk/cli/validate.rb
```diff
@@ -70,15 +71,47 @@ module PDK::CLI
...
       # Ensure that the bundle is installed and tools are available before running any validations.
       PDK::Util::Bundler.ensure_bundle!
 
-      validators.each do |validator|
-        exit_code = validator.invoke(report, options)
-        break if exit_code != 0
+      mod_root = PDK::Util.module_root
+      unless mod_root
+        raise PDK::CLI::FatalError, _('Current working directory is not part of a module. (No metadata.json was found.)')
+      end
```
rodjek: This block can be replaced with `PDK::CLI::Util.ensure_in_module!` however i'm not sure that we can to restrict validators to run in a module (for example, a module consumer would probably want to run `pdk validate` from the root of their control repo to run validation on everything at once).

lib/pdk/validators/metadata/metadata_syntax.rb
```diff
@@ -14,13 +14,29 @@ def self.pattern
         'metadata.json'
       end
 
+      def self.spinner_text(targets = [])
+        _('Checking metadata syntax (%{targets})') % {
+          targets: targets.map { |t| Pathname.new(t).absolute? ? Pathname.new(t).relative_path_from(Pathname.pwd) : t }.join(' '),
```
rodjek: As this is now being used in multiple places, we should probably split it out into a helper method (`PDK::Util.targets_relative_to_pwd` or something) to avoid duplicating the logic.

lib/pdk/cli/validate.rb
austb: I did not view this as a behavior change. previously we had this code around the console command call in cli/exec.rb. This had to be brought out to this level because Dir.chdir changes the working dir of the process, not the thread.

austb:
@rodjek The de po file has been removed entirely from this build. It contained outdated strings which were causing problems when creating new threads. I did not see any way to update a po file while keeping the old strings.

lib/pdk/cli/spinner.rb
```diff
@@ -0,0 +1,146 @@
...
+          raise PDK::CLI::FatalError, _("The event :#{key} does not exist. Only :done, :success, and :error allowed.") unless [:done, :success, :error].include? key
+
+          @spinner.on(key, &block)
+        end
+      end
+    end
+
+    # This is a placeholder until we integrate ansicon into Windows packaging
+    # or come up with some other progress indicator for Windows.
+    class WindowsSpinner
```
DavidS: ansicon is already integrated. we can drop this.

lib/pdk/cli/spinner.rb
```diff
@@ -0,0 +1,146 @@
+require 'tty-spinner'
+
+module PDK
+  module CLI
+    module Spinner
+      PREFIX = 'spinner_key'.freeze
+      LOCK = Monitor.new
```
DavidS: locks should be private to avoid other callers violating expectations.

lib/pdk/cli/validate.rb
```diff
@@ -70,15 +71,43 @@ module PDK::CLI
...
+              require 'pdk/i18n'
+              exit_codes << validator.invoke(report, options)
+            end
+          end
+        end
+
+        threads.each(&:join)
+        exit_codes.sort!
+
+        if exit_codes.last != 0
```
DavidS:
```
exit_codes.all? { |c| c == 0 }
```

https://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-all-3F

lib/pdk/util.rb
```diff
@@ -100,5 +100,43 @@ def module_root
       end
     end
     module_function :module_root
+
+    # Returns the targets' paths relative to the working directory
+    #
+    # @ return Array<String> The absolute or path to the target
```
DavidS: should that be `@return` instead of `@ return` ?

Where/how can we check the YARD output?

locales/de/pdk.po
```diff
@@ -1,604 +0,0 @@
-# German translations for puppet development kit package.
```
DavidS: this file should not be deleted

lib/pdk/cli/spinner.rb
austb: We also need unicode before our ✔️  or ✖️  will show up in windows. We could default to the text `passed` or `failed` on windows, ie

```
[passed] Checking metadata (metadata.json)
```

lib/pdk/cli/validate.rb
austb: I sorted them to return the highest error code with last in order to mirror the functionality [`base_validator.rb`](https://github.com/puppetlabs/pdk/blob/master/lib/pdk/validators/base_validator.rb#L79)

lib/pdk/util.rb
austb: We need to add a rake task to generate YARD output https://github.com/lsegal/yard#generating-documentation

locales/de/pdk.po
austb: This file caused a lot of problems for threads because it was no longer up to date with the current pdk.pot file, and when talking to the i18n folks they were confused why we even had a PO file in our repo as they will be managed by transifex. We should discuss this in more detail if you have concerns, but if we would like both the German PO file and threading the German PO file must be kept up to date and in sync with the pdk.pot file. Keeping a PO file in sync with the POT file without deleting all the existing translations is not a feature provided by the gettext-setup-gem we use because the Transifex workflow creates brand new PO files and populates them with the previous translations that match current strings.

lib/pdk/util.rb
DavidS: 👍

locales/de/pdk.po
DavidS: I would be interested how the non-up-to-date de.po file causes issues with threading, especially outside the `de` locale.

`rake gettext:update_pot` and `rake gettext:po[de]` worked fine until now to update the po file as needed, and to the right thing for gettext. I'm sure transifex can one-time import the existing translations once we get around to integrating.

> Another potential solution would be for you to keep the German file locally and not checked in to the repo.

I'm not even using the thing, what value would it have on my local machine?

lib/pdk/cli/validate.rb
DavidS: So that also needs fixing then: http://ruby-doc.org/core-2.4.0/Enumerable.html#method-i-max

lib/pdk/cli/spinner.rb
DavidS: having the words instead of the unicode chars is a good idea. I wouldn't block the PR for the change, but we need to do it at one point....

locales/de/pdk.po
austb: I believe we ran into two error messages, one involves fuzzy messages and the other is for old messages it leaves in comments at the bottom.

Commented
```
Warning: obsolete msgid exists.
         #~ msgstr "Verwende ein eigenes Vorlagerepository für dieses Modul."
```

Fuzzy
```
Warning: fuzzy message was used.
  : msgid 'Unable to write '%{file}': %{msg}'
```

lib/pdk/cli/spinner.rb
austb: Ok, I'll leave it as is for now, but get a fix for that up once we get this through.

austb:
Re-kicking Travis build, one of the `rake spec` jobs never started.

lib/pdk/util.rb
austb: This is fixed, just not shown as outdated for some reason

locales/de/pdk.po
austb: I have added a Rake task that cleans the file for us, one extra step to make sure it works, but we can keep the German file now

austb:
@DavidS I found the [root of the problem](https://github.com/puppetlabs/pdk/pull/144/commits/07a91939a21f3e9ce5724c57a656eed6caaea093). It occurred in the English locale because the threads did not correctly negotiate their locale (this setting is not carried between threads). This fixes the problem for the default locale (it will never output those warning messages), but the warning messages are still a possibility unless the po file was cleaned with `rake gettext:clean_po[language]`.

austb:
Opened #196 and #197 to fix two points made during review.

locales/de/pdk.po
DavidS: Excellent work, @austb . Thank you very much for your attention here!

DavidS: (specifically to discovering the root cause of the uninitialized threads)

DavidS:
Added a few small cleanups while testing

DavidS:
@austb apologies I didn't see your fixups in time, and force-pushed over them

austb:
@DavidS no problem 👍 if it's green, I'm happy

lib/pdk/validators/metadata/metadata_syntax.rb
```diff
@@ -14,12 +14,39 @@ def self.pattern
...
+
+      def self.invoke(report, *args)
+        options = args.pop if args.last.is_a?(::Hash)
+        threaded_spinner = args.pop if args.last.is_a?(::TTY::Spinner::Multi)
         targets = parse_targets(options)
 
         return 0 if targets.empty?
 
-        return_val = 0
+        @return_val = 0
```
scotje: Was there a reason this had to become an instance var? Something to do with threads?

lib/pdk/cli/exec.rb
```diff
@@ -137,31 +142,15 @@ def execute!
...
-            unless mod_root
-              @spinner.error
-
-              raise PDK::CLI::FatalError, _('Current working directory is not part of a module. (No metadata.json was found.)')
-            end
-
-            Dir.chdir(mod_root) do
-              ::Bundler.with_clean_env do
-                run_process!
-              end
```
scotje: We can test `Dir.pwd` so that we only try to `chdir` if we aren't already in the right place. That should keep the child threads from complaining.

scotje:
@jbondpdx Austin is going to add a blurb to the README about this feature and then we plan to merge. Despite the size of the PR, the actual user impact/README stuff should be pretty simple.

PR #145, opened 2017-07-11
(SDK-306) Use vendored development gems in package install

scotje:


lib/pdk/cli/exec.rb
```diff
@@ -101,9 +104,16 @@ def execute!
...
+              # Subprocesses use their own set of gems which are pre-populated from the package source.
+              @process.environment['GEM_HOME'] = File.join(PDK::Util.cachedir, 'ruby', RbConfig::CONFIG['ruby_version'])
+
+              # This allows the subprocess to find the 'bundler' gem, which isn't in the cachedir above.
+              @process.environment['GEM_PATH'] = File.join(PDK::Util.pdk_package_basedir, 'private', 'ruby', RUBY_VERSION, 'lib', 'ruby', 'gems', RbConfig::CONFIG['ruby_version'])
+            end
+
             # TODO: we should probably more carefully manage PATH and maybe other things too
-            @process.environment['GEM_HOME'] = File.join(PDK::Util.cachedir, 'bundler', 'ruby', RbConfig::CONFIG['ruby_version'])
-            @process.environment['GEM_PATH'] = pdk_gem_path
```
scotje: Hmmm, I might actually need to put this back for gem installs now that I'm doing `--path` again...

scotje: Updated this and have a companion PR in the vanagon repo here: https://github.com/puppetlabs/puppet-sdk-vanagon/pull/24

PR #154, opened 2017-07-14
(SDK-312) Add option --parallel to `pdk test unit`

austb:
Waiting for a new version of puppetlabs_spec_helper for the updated rake task

### Update:

I now believe it's waiting for a new version of puppet-module-gems

### Update 2: 

v0.1.0 of puppet-module-gems has been released

### DIFF
--- a/lib/pdk/tests/unit.rb
+++ b/lib/pdk/tests/unit.rb
@@ -6,9 +6,10 @@ require 'json'
 module PDK
   module Test
     class Unit
-      def self.cmd(_tests)
+      def self.cmd(_tests, opts = {})
         # TODO: test selection
-        [File.join(PDK::Util.module_root, 'bin', 'rake'), 'spec']
+        rake_task = opts.key?(:parallel) ? 'parallel_spec' : 'spec'
+        [File.join(PDK::Util.module_root, 'bin', 'rake'), rake_task]
       end
 
       def self.invoke(report, options = {})
@@ -17,12 +18,13 @@ module PDK
 
         tests = options.fetch(:tests)
 
-        cmd_argv = cmd(tests)
+        cmd_argv = cmd(tests, options)
         cmd_argv.unshift('ruby') if Gem.win_platform?
 
         command = PDK::CLI::Exec::Command.new(*cmd_argv).tap do |c|
           c.context = :module
-          c.add_spinner('Running unit tests')
+          spinner_msg = options.key?(:parallel) ? _('Running unit tests in parallel') : _('Running unit tests')
+          c.add_spinner(spinner_msg)
           c.environment['CI_SPEC_OPTIONS'] = '--format j'
         end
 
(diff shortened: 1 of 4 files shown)


lib/pdk/tests/unit.rb
```diff
@@ -6,9 +6,13 @@
 module PDK
   module Test
     class Unit
-      def self.cmd(_tests)
+      def self.cmd(_tests, opts = {})
         # TODO: test selection
-        [File.join(PDK::Util.module_root, 'bin', 'rake'), 'spec']
+        if opts.key?(:parallel)
+          [File.join(PDK::Util.module_root, 'bin', 'rake'), 'parallel']
```
austb: This will need to change once puppetlabs_spec_helper is updated. Added the updated `parallel_spec` task to the module's Rakefile as `parallel` to avoid name conflicts.

rodjek: As this is just deciding between two rake tasks, you can remove a bit of duplication here by changing the method content to something like this

```ruby
rake_task = opts.key?(:parallel) ? 'parallel' : 'spec'
[File.join(PDK::Util.module_root, 'bin', 'rake'), rake_task]
```

austb:
puppet-module-gems 0.1.0 has been released, so this is ready to review.

spec/unit/pdk/test/unit_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    end
+
+    it 'assigns given duration to the result' do
+      json_result = described_class.merge_json_results(results, duration)
+
+      expect(json_result['summary']['duration']).to eq(duration)
+    end
+  end
+
+  describe '#cmd' do
```
rodjek: This is a class method, so `.cmd` would be better.

spec/unit/pdk/test/unit_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+    context 'when run without the parallel option' do
+      it 'uses the spec rake task' do
+        cmd = described_class.cmd(nil)
+
+        expect(cmd.join('/')).to match(%r{bin/rake/spec})
+      end
+    end
+  end
+
+  describe '#parse_output' do
```
rodjek: This is a class method, so `.parse_output` would be better.

spec/unit/pdk/test/unit_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+        expect($stderr).to receive(:puts).once.with(%r{Evaluated 30 tests in 32 seconds})
+
+        described_class.parse_output(report, json)
+      end
+    end
+  end
+
+  # Allow any_instance stubbing of Commands
+  # rubocop:disable RSpec/AnyInstance
+  describe '#list' do
```
rodjek: This is a class method, so `.list` would be better.

spec/unit/pdk/test/unit_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+        }'
+      end
+
+      it 'returns an empty array' do
+        expect(described_class.list).to eq([])
+      end
+    end
+  end
+
+  describe '#invoke' do
```
rodjek: This is a class method, so `.invoke` would be better.

PR #168, opened 2017-07-18
(SDK-319) force usage of our ruby

DavidS:
When running commands in module context, we always want to use our private
ruby. Especially bundler's binstubs are using ruby from the PATH, which
led to weird failures, as the multiple rubies would not agree on installed
gems, etc.


scotje:
Is there a reason we aren't using `@process.environment` to do this instead of all the `begin/ensure` business?

DavidS:
@scotje  nothing worse than my incompetence

scotje:
OK, I'll lump it a simplification with my bundle path fix then.

PR #169, opened 2017-07-18
(maint) fix `new module` description in README

### LINKED ISSUE #159: `new module` docs differ from reality
The README has an example of the modulename of the `new module` command https://github.com/puppetlabs/pdk#module_name

```
Required. Specifies the name of the module being created, including the namespace. e.g.: username-my_module
```

When running with this example the run fails:

```
pdk new module username-my_module
pdk (FATAL): 'username-my_module' is not a valid module name.
Module names must begin with a lowercase letter and can only include lowercase letters, digits, and underscores.
```

Relates to #157 in that it doesn't appear possible to specify the namespace.

DavidS:
This fixes #159.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ Suppress interactive queries for initial values. All questions will use the defa
 
 #### `module_name`
 
-**Required**. Specifies the name of the module being created, including the namespace. e.g.: `username-my_module`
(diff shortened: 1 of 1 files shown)


DavidS:
@scotje definitely - could be part of the work when we come around to persisting default answers

PR #172, opened 2017-07-18
(maint) mention the execution policy for windows

DavidS:


DavidS:
@jpogran can you confirm that this note is sufficient for Windows folks to figure out what they need to do? If no, what else should it say?

scotje:
A link to https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.core/about/about_execution_policies might be more comprehensive

scotje:
This is fine for now.

jpogran:
I'm sorry I didn't get to reply, I'm on a different timezone than you.

I would have not used the word 'required' as different versions of PowerShell have different default execution policies. We can avoid having to mention this at all by signing the module

DavidS:
@jpogran I'm aware that you are not at my beck and call, just because I have a release deadline. As Jesse said, this is fine for now.

Your clarification is still highly valuable. Nothing is as simple as it appears on the tin :-D

Do we have infrastructure in place to make this signing possible, and who do I need to talk to to make it happen?

jpogran:
We need a `Authenticode Digital Certificate` or sometimes called `Code Signing Certificate` then its a matter of passing the file and cert to a PowerShell cmdlet:

```
Set-AuthenticodeSignature $file $cert
```

We would do this to the module (psm1) and not the manifest (psd1)

scotje:
I don't think the Authenticode signing actually mitigates the issue everywhere because the *default* policy on Windows 10 is "Restricted" which just won't run scripts period (at least that's my understanding from the above-linked document). It would however let us work under the "AllSigned" policy though.

PR #173, opened 2017-07-18
(MAINT) Release prep for 0.4.4

scotje:


CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 All notable changes to this project will be documented in this file.
 
 
+## [v0.4.4](https://github.com/puppetlabs/pdk/tree/v0.4.4) (2017-07-18)
+[Full Changelog](https://github.com/puppetlabs/pdk/compare/v0.4.3...v0.4.4)
+
+**Fixed bugs:**
+
+- Validate fails on existing module [\#158](https://github.com/puppetlabs/pdk/issues/158)
+- Can't create module if new module fails to get login [\#157](https://github.com/puppetlabs/pdk/issues/157)
```
austb: This is just the issue for the commit below on line 29. Do we want this? Should it be removed entirely, or maybe just moved down to the v0.4.2 section? It seems confusing to have it in our changelog in two versions.

This probably highlights a bigger issue of duplicate changelog entries for issues and prs tagged with "bug".

scotje: Yeah I don't know that I care about somewhat duplicate entries, there may be value in users seeing their exact issue under the list of fixed things since the commit/PR message might be more general.

austb: And is it ok that it is in a different version in the changelog than the commit?

scotje: Different version?

PR #175, opened 2017-07-19
(SDK-317) Ensure parent of 'pdk new module' is writable before generation

rodjek:
Does an initial check before starting the interview, so that we don't waste the user's time. Also rescues `Errno::EACCES` from `FileUtils.mv` in the event that the permissions change between the pre-run check and the module generation.

### DIFF
--- a/spec/generators/module_spec.rb
+++ b/spec/generators/module_spec.rb
@@ -76,6 +76,7 @@ describe PDK::Generate::Module do
       include_context 'mock metadata.json'
 
       let(:temp_target_dir) { '/path/to/temp/dir' }
+      let(:target_parent_writeable) { true }
 
       before(:each) do
         allow(File).to receive(:exist?).with(target_dir).and_return(false)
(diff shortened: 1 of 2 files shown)


spec/generators/module_spec.rb
```diff
@@ -76,13 +76,25 @@
...
         allow(PDK::Util::Version).to receive(:version_string).and_return('0.0.0')
         allow(described_class).to receive(:prepare_module_directory).with(temp_target_dir)
+        allow(File).to receive(:writable?).with(File.dirname(target_dir)).and_return(target_parent_writeable)
+      end
+
+      context 'when the parent directory of the target is not writable' do
+        let(:target_parent_writeable) { false }
+
+        it 'raises a FatalError' do
+          expect {
```
DavidS: Do you also need to expect that it doesn't start the interview here first?

scotje: The ways I tried doing this as a unit test were all pretty clunky a likely to just silently continue passing if the underlying code was refactored. I'm inclined to just merge as-is.

PR #179, opened 2017-07-20
(SDK-329) implement running arbitrary commands in PDK's environment

DavidS:


README.md
```diff
@@ -270,13 +270,36 @@ Multiple `--format` options can be specified as long as they all have distinct o
 
 Specifies options to pass through to the actual test-runner. In the default template (and most commonly across modules), this is [rspec](https://relishapp.com/rspec/rspec-core/docs/command-line).
 
+#### `pdk bundle` command
```
james-stocks: I think this is indented one hash too many

lib/pdk/cli/bundle.rb
```diff
@@ -0,0 +1,24 @@
+
```
james-stocks: I think this blank line shouldn't be here; I'm surprised rubocop doesn't complain about it

README.md
```diff
@@ -270,13 +270,36 @@ Multiple `--format` options can be specified as long as they all have distinct o
...
+
+This experimental command allows you to execute arbitrary commands in a bundler context within the module you're currently working on. Arguments to this command are passed straight through to bundler.
```
james-stocks: The command text says this is for 'advanced users' - we might want at least that level of warning here; if not a stronger warning that the user should know what they're doing (e.g. they might install a gem to system with this command, causing problems on their machine)?

lib/pdk/cli/bundle.rb
DavidS: this is not the only file to have this.

README.md
```diff
@@ -270,13 +270,36 @@ Multiple `--format` options can be specified as long as they all have distinct o
 
 Specifies options to pass through to the actual test-runner. In the default template (and most commonly across modules), this is [rspec](https://relishapp.com/rspec/rspec-core/docs/command-line).
 
+### `pdk bundle` command
+
+This experimental command allows you to execute arbitrary commands in a bundler context within the module you're currently working on. Arguments to this command are passed straight through to bundler.
+
+Note that for most advanced uses it is required to use the `--` to separate bundler options from pdk options. Compare the following two commands:
```
james-stocks: The command text says this is for 'advanced users' - we might want at least that level of warning here; if not a stronger warning that the user should know what they're doing (e.g. they might install a gem to system with this command, causing problems on their machine)?

README.md
DavidS: ha, I copied it from the one heading that had it wrong. fixed both now.

README.md
DavidS: Good call. I'll make the language a bit stronger.

DavidS:
Updated, addressing all recommendations.

PR #181, opened 2017-07-21
(maint) Move contributor's notes to separate file

DavidS:
This both avoids confusion, and makes the content easier to discover.


MichalBryxi:
Ah right. That explains a lot :-D

PR #182, opened 2017-07-21
(maint) change message to match validate_pattern for forge username

DavidS:


DavidS:
@scotje please confirm with actual forge behaviour!


cc @MichalBryxi

MichalBryxi:
Forge frontend validation regexp:
`^[a-zA-Z0-9]+$`

Error message from forge when trying to bypass that validation (I guess it's from the server side):
screen shot 2017-07-21 at 09 36 11

So this change is LGTM.

austb:
The Forge regex allows users to pretend they have capital letters, but they are coerced to lowercase before they are sent to the API. Capital letters were originally allowed in Forge usernames. They are no longer allowed (FORGE-117). Older accounts may still have uppercase usernames, but [are compared case-insensitive](https://github.com/puppetlabs/puppet-forge-api/pull/304/files).

scotje:
I would vote for *not* merging this, we are definitely trying to move people away from mixed-case usernames and I don't think a mismatch of case causes a publishing issue, although if it does we can just fix that on the Forge side.

MichalBryxi:
Yep, tested. Forge does not allow me to create duplicated username - considering that it's ignoring casing.

screen shot 2017-07-24 at 10 40 54

My problem here is:
1) I try to use PDK
2) I'm prompted to put in Forge username
3) I don't remember Forge username, so I go to Forge and copy+paste it to PDK
5) PDK tells me that it's not correct
6) I'm sad & confused

IMO we should do at least one of these:
a) Make sure that forge at least displays the username in lowercase. For this one scenario. Not sure if there are other that might ruin users day though...
b) Provide better message in PDK that explains this disparity between PDK and forge. Might be even better, because if user actually remembers their username, they might be confused. I can actually use "MichalBryxi" now in the forge (website) to log in.
c) Do the downcasing for the user in PDK wizard. And maybe inform them why we did that if we did.

austb:
I would recommend that we coerce their usernames to lowercase for them, so that way we are doing the same thing during the module generation that we do during the Forge user create/login. Then we can log a warning such as "Your username was coerced to lowercase, this will work with the Forge regardless of your username case."

scotje:
Yeah, I would vote for option C I guess.

scotje:
So there is a `:modify` option in tty-prompt which would let us easily downcase the string before validation. The only problem is that it doesn't update the value shown on the prompt line after the question, so the prompt still says:

```
[Q 1/8] What is your Puppet Forge username?
This will be used when uploading your module to the Forge. You can opt out of this at any time.
--> JesseScott
```

when the resulting metadata would be:

```
"name": "jessescott-testusername",
```

We could PR tty-prompt to change how the prompt is value is displayed after entry.

scotje:
@austb do you want to look at updating tty-prompt to display the post-`modify` value?

bmjen:
Added a `do not merge` label as it looks like this PR needs re-work.

PR #183, opened 2017-07-21
(SDK-321) add `pdk validate help`

DavidS:


lib/pdk/cli/validate.rb
```diff
@@ -17,6 +17,11 @@ module PDK::CLI
     flag :a, 'auto-correct', _('automatically correct problems (where possible)')
 
     run do |opts, args, _cmd|
+      if args == ['help']
```
james-stocks: Is there a Cri way to do this? For other commands (test, new) it looks like we get the help subcommand for free. But it looks like 'validate' is different since on it's own (just 'pdk validate') it is a valid command

austb: Adding `@validate_cmd.add_command Cri::Command.new_basic_help` works

DavidS: `validate` is "special" because the choice of validator conflicts with CRI's parsing of `validate help`. Try it, acceptance tests will fail.

PR #184, opened 2017-07-21
(maint) Change package testing to beaker tests

james-stocks:
Acceptance on package installs needs to test package specific things, and also needs to test Powershell.
Therefore the previous approach of running the spec/acceptance/*_spec.rb tests is not appropriate.
This commit changes package testing to use beaker tests.


scotje:
vmpooler has been down today so I haven't been able to test this, but it looks good in principle! :)

scotje:
vmpooler is back and this looks good!

PR #190, opened 2017-07-24
(MAINT) Move all unit tests under spec/unit

scotje:
This simplifies running unit tests with your own rspec invocation
without having to copy the pattern from the rake task.


scotje:
Some of the renaming was also to comply with rubocop rules.

scotje:
I think it wants them to match is the thing, but it also only complains about files once you touch them in a commit. :)

scotje:
I'll leave this open so that BFS has a chance to weigh in.

rodjek:
:+1: this and `lib/pdk/validators` => `lib/pdk/validate`, `lib/pdk/generators` => `lib/pdk/generate` has been bugging me for a while

DavidS:
👍

remove the relevant exclusions from https://github.com/puppetlabs/pdk/blob/master/.rubocop_todo.yml

scotje:
I think I successfully updated the exclusions.

DavidS:
Looks good to me!

PR #191, opened 2017-07-24
(DISCUSS) (SDK-331) Use vendored Gemfile.lock for new modules if present

scotje:
This will be paired with a change in the packaging to vendor the generated Gemfile.lock.

Note that the Gemfile.lock will still be .gitignored.

### DIFF
--- a/lib/pdk/generators/module.rb
+++ b/lib/pdk/generators/module.rb
@@ -50,6 +50,15 @@ module PDK
 
         PDK.answers.update!('template-url' => template_url)
 
+        # In packaged installs, try to use vendored Gemfile.lock as a starting point.
+        if PDK::Util.package_install?
+          vendored_gemfile_lock = File.join(PDK::Util.package_cachedir, 'Gemfile.lock')
+
+          if File.exist?(vendored_gemfile_lock)
+            FileUtils.cp(vendored_gemfile_lock, File.join(temp_target_dir, 'Gemfile.lock'))
+          end
+        end
+
         FileUtils.mv(temp_target_dir, target_dir)
       end
 


scotje:
https://github.com/puppetlabs/puppet-sdk-vanagon/pull/37 is the puppet-sdk-vanagon PR

scotje:
Exact pinning seems very inflexible for any users that might have additional gem dependencies.

I'm also curious what about this seems like "tricking around with guts of the tooling" as it all seems very straightforward to me. This simply makes bundler prefer to resolve dependencies onto the gems it will already have available locally, but doesn't do anything to impede or even modify the standard dependency resolution behavior, we've just pre-run the first `bundle lock` for them.

DavidS:
> Exact pinning seems very inflexible for any users that might have additional gem dependencies.

Adding more gems on top of the meta-gems will still work, as long as the pinned versions do not conflict with the requirements of the additional gems. I'd also like to see a regular validated update of the meta-gems, so that we detect issues early (when validation fails), and keep the recommended gems fresh. This of course rides on the expectation that those additional gems are also well-maintained.

> I'm also curious what about this seems like "tricking around with guts of the tooling" as it all seems very straightforward to me. This simply makes bundler prefer to resolve dependencies onto the gems it will already have available locally, but doesn't do anything to impede or even modify the standard dependency resolution behavior, we've just pre-run the first bundle lock for them.

I'm not very familiar with bundler's resolution algorithms. In my view, the `Gemfile.lock` is an opaque implementation detail of bundler. A private part of it. Reaching into an external product like this is a measure of last resort.

Considering the case where an additional gem dependency conflicts with the recommended versions, the choice is between an invalid configuration (uninstallable gems due to version conflicts), and (potential) (subtle) errors due to unexpected interactions with untested newer/older versions of components.

I will trust your judgement on whatever solution you finally choose after this.

scotje:
We should probably hash this out in a meeting or via e-mail once everyone (*cough* Bryan *cough*) is back in the office.

PR #192, opened 2017-07-25
(maint) Remove unused vcs option from 'pdk new module'

rodjek:
This option doesn't currently do anything as it's waiting on getting libgit2 into the package (which is blocked on devkit in the windows packages). We should remove this option for now as it suggests functionality that doesn't exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Generates a new module.
 Usage:
 
 ```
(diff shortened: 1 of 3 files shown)


DavidS:
It's also mentioned in the README

scotje:
Rebased and updated README

PR #201, opened 2017-07-27
(maint) enable coveralls

DavidS:


.travis.yml
```diff
@@ -3,29 +3,24 @@ language: ruby
 before_install: gem install bundler -v 1.15.1
 bundler_args: "--without development"
 script:
-  - "cat Gemfile.lock"
-  - "bundle exec rake $CHECK"
+- cat Gemfile.lock
+- bundle exec rake $CHECK
```
scotje: I'm confused by the elimination of this indentation, still seems to be valid YAML either way but was there something that prompted this (and the similar change to the matrix/include section below)?

PR #204, opened 2017-07-28
(maint) Add unit tests for PDK::Util

rodjek:


spec/unit/pdk/util_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+      context 'and the target file exists' do
+        before(:each) do
+          allow(File).to receive(:file?).with('/path/to/something/metadata.json').and_return(true)
+        end
+
+        it { is_expected.to eq('/path/to/something/metadata.json') }
+      end
+
+      context 'and the target file does not exist' do
+        it { is_expected.to be_nil }
```
james-stocks: Should this have something like

    expect(File).to receive(:file?).with('/path/to/something/metadata.json').and_return(false)

?
Maybe the code would be returning nil due to being broken, rather than due to not finding the file

james-stocks: Discussed with David, it's maybe better to keep this as is and not add a dependency on how File behaves

james-stocks:
Pushed a commit that simplifies find_upwards and its unit tests

spec/unit/pdk/util_spec.rb
```diff
@@ -0,0 +1,242 @@
...
+
+  describe '.find_upwards' do
+    subject { described_class.find_upwards(target, start_dir) }
+
+    let(:target) { 'metadata.json' }
+    let(:start_dir) { nil }
+    let(:actual_start_dir) { '/path/to/something/deep/in/a/module' }
+
+    before(:each) do
+      allow(File).to receive(:directory?).with('/path/to/something/deep/in/a/module').and_return(true)
```
james-stocks: I added a commit that unrolled the loop here so we just have a dumb list of expected messages; to keep the unit test code as simple as possible.

DavidS: 👍

PR #207, opened 2017-07-28
(SDK-317) Replace File.writable? test with actually creating a test file

scotje:


rodjek:
:+1: Eventually, we should probably look at adding win32-file to the Windows package, which seems to implement many of the `File` methods in a way that actually supports NTFL ACLs.

DavidS:
The win32-* gems are spottily maintained, and nobody's stepping up to fixing their existing issues :-(

PR #214, opened 2017-08-01
(MAINT) Add ability to test locally built package with beaker

scotje:
This change lets you run the package-based acceptance tests but on a package you have built locally instead of requiring it to be on the builds server.

For example:

```
LOCAL_PKG=/path/to/package/pdk-0.5.0.0.11.g612fd91-1.el7.x86_64.rpm TEST_TARGET=redhat7-64workstation. bundle exec rake acceptance
```

I found this greatly reduced my cycle times for testing acceptance test changes that also required code changes.

Thoughts? Is there a more standard way to do this?

### DIFF
--- a/package-testing/pre/000_install_package.rb
+++ b/package-testing/pre/000_install_package.rb
@@ -6,6 +6,10 @@ test_name 'Install pdk package on workstation host' do
       # BKR-1109 requests a neater way to install an MSI
       msi_url = "http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/repos/windows/pdk-x64.msi"
       generic_install_msi_on(workstation, msi_url)
+    elsif ENV['LOCAL_PKG']
+      pkg_fullname = File.basename(ENV['LOCAL_PKG'])
+      scp_to(workstation, ENV['LOCAL_PKG'], "/root/#{pkg_fullname}")
+      workstation.install_package(pkg_fullname)
     else
       install_puppetlabs_dev_repo(workstation, 'pdk', ENV['SHA'], 'repo-config')
       workstation.install_package('pdk')
--- a/package-testing/pre/000_install_package.rb
+++ b/package-testing/pre/000_install_package.rb
@@ -2,17 +2,24 @@ test_name 'Install pdk package on workstation host' do
   workstation = find_at_most_one('workstation')
 
   step 'Install pdk package' do
+    if ENV['LOCAL_PKG']
+      pkg = File.basename(ENV['LOCAL_PKG'])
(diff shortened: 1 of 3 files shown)


package-testing/pre/000_install_package.rb
```diff
@@ -6,6 +6,10 @@
       # BKR-1109 requests a neater way to install an MSI
       msi_url = "http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/repos/windows/pdk-x64.msi"
       generic_install_msi_on(workstation, msi_url)
+    elsif ENV['LOCAL_PKG']
+      pkg_fullname = File.basename(ENV['LOCAL_PKG'])
+      scp_to(workstation, ENV['LOCAL_PKG'], "/root/#{pkg_fullname}")
+      workstation.install_package(pkg_fullname)
```
james-stocks: Either Windows should be supported; or the run should abort if LOCAL_PKG is attempted when workstation is Windows

james-stocks:
@jessereynolds I'll test and merge this tomorrow. I've one more minor comment that I'll add, but again I can address the comment myself tomorrow

package-testing/pre/000_install_package.rb
```diff
@@ -2,13 +2,24 @@
   workstation = find_at_most_one('workstation')
 
   step 'Install pdk package' do
+    if ENV['LOCAL_PKG']
+      pkg = File.basename(ENV['LOCAL_PKG'])
+      scp_to(workstation, ENV['LOCAL_PKG'], pkg)
```
james-stocks: It's worth testing the local file exists before calling scp_to because beaker doesn't have a lot of parameter validation and probably won't cleanly fail if the local file doesn't exist.
Though we might be already covered by the call to File.basename on L6

scotje: I don't think `basename` touches the filesystem. I was thinking of adding a `File.exist?` check but I wasn't sure how beaker expected a pre-suite to indicate failure, do you just `abort` or what?

james-stocks: It seems like we get a reasonable error message already:

    Install pdk package on workstation host
    
      * Install pdk package
        localhost $ scp john_cena.msi windows2012r2-64-1:john_cena.msi {:ignore => }
        IOError: No such file or directory - john_cena.msi
        /home/centos/code/pdk/package-testing/pre/000_install_package.rb:7
        /home/centos/code/pdk/package-testing/pre/000_install_package.rb:4

I will add a File.exist? check to the Rakefile though, as it's nicer to abort immediately before beaker spends time on setting up VMs

PR #216, opened 2017-08-02
(PDK-373) Make test unit --list consistent with test unit

james-stocks:
`pdk test unit` was using a pattern defined in puppetlabs_spec_helper while `pdk test unit --list` was using no pattern.
This commit updates the --list option to use a puppetlabs_spec_helper rake task so it uses the same pattern.

### DIFF
--- a/lib/pdk/tests/unit.rb
+++ b/lib/pdk/tests/unit.rb
@@ -32,14 +32,7 @@ module PDK
 
         result = command.execute!
 
-        json_result = nil
-        json_result = [] if options.key?(:parallel)
-
-        if options.key?(:parallel)
-          json_result.push(PDK::Util.find_valid_json_in(result[:stdout]))
-        else
-          json_result = PDK::Util.find_valid_json_in(result[:stdout])
-        end
+        json_result = PDK::Util.find_valid_json_in(result[:stdout], !options.key?(:parallel))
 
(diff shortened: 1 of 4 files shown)


james-stocks:
I can't find a way to make _PDK::Test::Unit.find_valid_json_in_ private that works for unit tests AND acceptance tests AND rubocop.
I think I can and should move the method to PDK::Util instead of having it private to PDK::Test::Unit

lib/pdk/tests/unit.rb
```diff
@@ -32,22 +32,13 @@ def self.invoke(report, options = {})
...
-              json_result.push(JSON.parse(str))
-            else
-              json_result = JSON.parse(str)
-              break
-            end
-          rescue JSON::ParserError
-            next
-          end
+        if options.key?(:parallel)
+          json_result.push(PDK::Util.find_valid_json_in(result[:stdout]))
```
austb: This will only get the first JSON result, but in parallel there will be many. I'll push a fix and an acceptance test.

james-stocks: Apologies @austb I didn't realise that this was the point at which parallel results were actually gathered together

austb: No problem @james-stocks, just pushed a new acceptance test. And it made me realize that it can only parallelize when there are multiple spec files. So I learned something about what I did 👍

PR #218, opened 2017-08-02
(MAINT) Fix fatal error in test unit --parallel

austb:
When there are no tests `pdk test unit` outputs `No examples found`, but
the parallel_tests gem aborts with an error and does not output JSON.
This catches that error situation and standardizes the output.

To reproduce simply make a new module with `pdk new module` and run `pdk test unit` and `pdk test unit --parallel`

### DIFF
--- a/lib/pdk/tests/unit.rb
+++ b/lib/pdk/tests/unit.rb
@@ -12,6 +12,12 @@ module PDK
         [File.join(PDK::Util.module_root, 'bin', 'rake'), rake_task]
       end
 
+      def self.parallel_with_no_tests?(ran_in_parallel, json_result, result)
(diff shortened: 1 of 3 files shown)


lib/pdk/tests/unit.rb
```diff
@@ -50,6 +50,14 @@ def self.invoke(report, options = {})
           end
         end
 
+        if options.key?(:parallel) && json_result.empty? &&
+           !result[:exit_code].zero? &&
+           result[:stderr].strip =~ %r{Pass files or folders to run$}
```
scotje: Maybe create a little helper method for this condition to improve readability?

```
if parallel_with_no_tests?
  json_result = foo
end
```

PR #221, opened 2017-08-03
Package testing: beaker needs to have keys configured

james-stocks:
When running in Jenkins CI, beaker needs to be told what key(s) to load.
The current tests work fine on our workstations but fail to connect to VMs in Jenkins CI.

### DIFF
--- a/package-testing/config/options.rb
+++ b/package-testing/config/options.rb
@@ -3,4 +3,7 @@
   pre_suite: [
     'pre/000_install_package.rb',
   ],
+  :ssh => {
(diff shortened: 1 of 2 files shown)


DavidS:
Please remember to talk to @scotje about the test changes.

scotje:
Probably wasn't showing up because the build with the Gemfile.lock stuff promoted also had the STDOUT -> STDERR change merged, so it should be `outcome.stderr` now. 

997ae414deaec80565507856b0f3855b8aae40a1 is the build that should include the Gemfile.lock stuff.

PR #223, opened 2017-08-04
Pdk preview docs

jbondpdx:
These have a reworked structure that should be much better for what users actually want to *do* with PDK. See pdk.md for the overview / getting started.


docs/pdk.md
```diff
@@ -0,0 +1,32 @@
...
+## Getting started
+
+To get started, install the PDK, and then create and test a module.
+
+<!--TK: overview workflow graphic-->
+
+1. [Generate a module](./pdk_generating_modules.md) using the `pdk new module` command.
+1. [Validate your module](./pdk_testing.md), to verify that it is well-formed.
+1. [Unit test](./pdk_testing.md) your module, to verify that all dependencies and directories are present.
+1. [Generate a class]](./pdk_generating_modules.md) for your module, using the `pdk new class` command.
```
bmjen: Extra `]` here, also should the link be to `./pdk_generating_modules.md#generate-a-new-class`

docs/pdk_generating_modules.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Generate a module with pdk
+
+To generate a module with PDK's default template, use the `pdk new module` command.
+
+Before you begin, ensure that you've installed the PDK package.
+
+1. From the command line, run the `pdk new module` command, specifying the name of the module: `pdk new module module_name`
+   
+   Optionally, to skip the interview questions and generate the module with default values, use the ``skip-interview` flag when you generate the module: `pdk new module module_name --skip-interview`
```
bmjen: Extra ` here.

docs/pdk_generating_modules.md
```diff
@@ -0,0 +1,84 @@
...
+   1. Forge username: Enter your Forge username, if you have a Forge account.
+   2. Version: Enter the semantic version of your module, such as "0.1.0".
+   3. Author: Enter the name of the module author.
+   4. License: If you want to specify a license other than "Apache-2.0," specify that here, such as "MIT", or "proprietary".
+   5. Description: Enter a one-sentence summary that helps other users understand what your module does.
+   6. Source code repository: Enter the URL to your module's source code repository.
+   7. Where others can learn more: If you have a website where users can learn more about your module, enter the URL.
+   8. Where others can report issues: If you have a public bug tracker for your module, enter the URL.
+
+1. If the metadata that PDK displays is correct, confirm to generate the module. If it is incorrect, enter `n` to cancel and start over.
```
bmjen: Should `confirm to generate the module` indicate hitting the `Enter` key or typing `y` and then `Enter`?

docs/pdk.md
jbondpdx: I added more specific links to all these.

docs/pdk_generating_modules.md
jbondpdx: I was thinking about this, but it also says what to do right in the interface. That said, I can add it though, I think it seems OK with it too.

jbondpdx:
Updated with respect to @bmjen 's comments.

docs/pdk_reference.md
```diff
@@ -0,0 +1,83 @@
...
+```
+pdk new module [--template-url=git_url] [--license=spdx_identifier] module_name [target_dir]
+```
+
+The `pdk new module` command accepts the following arguments and options. Arguments are optional unless otherwise specified.
+
+Argument   | Description   | Values      | Default
+----------------|:---------------:|:------------------:|-------------------------
+`--template-url=git_url` | Overrides the template to use for this module. | A valid Git URL.    | No default.
+`--license=spdx_identifier` | Specifies the license this module is written under. | See https://spdx.org/licenses/ for a list of open source licenses, or use `proprietary`.    | Apache 2.0
```
scotje: Tiny nit but the default license value is `Apache-2.0`

bmjen: Good point, that `-` is actually rather important when it comes to linting and parsing on the Forge right?

scotje: Yes, but the actual value in PDK seems correct.

jbondpdx: Yeah, legit, good catch @scotje

scotje: Might be worth clarifying that they should use the "identifier" from that SPDX link.

PR #225, opened 2017-08-07
(PDK-385) Support package testing on OSX

james-stocks:
Add package install for OSX (beaker does not have a helper for this, yet. See https://tickets.puppetlabs.com/browse/BKR-1109 for having a single helper for all platforms implemented).

I'm unhappy with the amount of code here when beaker and vanagon should provide a 2-liner method to get a .dmg from the build server and install it on the workstation host.

Tested against a number of builds (including per-commit builds and tag builds). Also tested that this works with the LOCAL_PKG option to point to a local .dmg file instead of one on the build server

### DIFF
--- a/package-testing/pre/000_install_package.rb
+++ b/package-testing/pre/000_install_package.rb
@@ -7,12 +7,23 @@ test_name 'Install pdk package on workstation host' do
       scp_to(workstation, ENV['LOCAL_PKG'], pkg)
     end
 
-    if workstation['platform'] =~ %r{windows}
-      pkg ||= "http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/repos/windows/pdk-x64.msi"
-
-      # BKR-1109 requests a neater way to install an MSI
+    case workstation['platform']
+    # TODO: BKR-1109 requests a supported way to install packages on Windows and OSX
+    when %r{windows}
+      pkg ||= "http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/artifacts/windows/pdk-#{ENV['SUITE_VERSION']}-x64.msi"
       generic_install_msi_on(workstation, pkg)
+    when %r{osx}
+      version, arch = workstation['platform'].split('-')[1, 2]
+      pkg ||= "http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/artifacts/apple/#{version}/PC1/#{arch}/pdk-#{ENV['SUITE_VERSION']}-1.osx#{version}.dmg"
+
+      # The beaker helper for dmg needs to know the /Volumes folder name the dmg will mount to, and the pkg filename contained within that folder
+      package_volume_name = "pdk-#{ENV['SUITE_VERSION']}"
+      package_filename = "#{package_volume_name}-1-installer.pkg"
+
+      logger.info("About to install '#{package_filename}' from '#{pkg}' on '#{workstation.hostname}'")
+      workstation.generic_install_dmg(pkg, package_volume_name, package_filename)
     else
+      # For most platforms, beaker will install the dev repo from the build server then 'install_package('pdk')' can simply be used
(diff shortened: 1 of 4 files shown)


package-testing/pre/000_install_package.rb
```diff
@@ -7,12 +9,28 @@
...
-      # BKR-1109 requests a neater way to install an MSI
       generic_install_msi_on(workstation, pkg)
+
+    when /osx/
+      unless pkg
+        # TODO: Replace this code that fetches build yaml and slices details from it. Is there a way to know the URL to the dmg file, given only SHA? 
+        build_yaml = open("http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/repos/#{ENV['SHA']}.yaml").read
+        build_yaml = YAML.load(build_yaml)
+        artifact = build_yaml[:platform_data][workstation['platform']][:artifact][1..-1]
+        pkg = "http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/repos#{artifact}"
```
james-stocks: I think Jenkins provides the required build string here as SUITE_VERSION.
If I use that then I can remove this yaml and string slicing code.
That would mean one more ENV for a local user to have to manage; but that's better than having to rely on this code...

scotje:
Vanagon now helpfully writes a `SHA.build_metadata.json` to the `artifacts` directory which contains the `SUITE_VERSION`, so we can use that to calculate the needed version values without requiring `SUITE_VERSION`.

package-testing/pre/000_install_package.rb
```diff
@@ -1,18 +1,42 @@
...
+
+      begin
+        build_info = JSON.parse(open(metadata_url).read)
+      rescue StandardError => e
+        abort("Could not get build metadata from build server. Tried to request '#{metadata_url}' and got: #{e.message}")
+      end
 
-    if workstation['platform'] =~ %r{windows}
-      pkg ||= "http://#{ENV['BUILD_SERVER']}/pdk/#{ENV['SHA']}/repos/windows/pdk-x64.msi"
+      full_version = build_info['version']
```
scotje: We could do all this in the `Rakefile` and actually set `SUITE_VERSION` only if unset. Would that be better?

scotje:
Oh and here is where the version-less MSI is getting created: https://github.com/puppetlabs/packaging/blob/master/tasks/ship.rake#L619-L642

package-testing/pre/000_install_package.rb
james-stocks: It might be most helpful to use SUITE_VERSION if it's already set; and fall back to build_metadata.json if it's not set.

I'm unsure if I prefer the code being here or in the Rakefile. It might be better in the Rakefile in order to ensure we have all the essential data before executing beaker and allocating VMs

package-testing/pre/000_install_package.rb
```diff
@@ -1,18 +1,42 @@
...
+
 test_name 'Install pdk package on workstation host' do
   workstation = find_at_most_one('workstation')
 
   step 'Install pdk package' do
     if ENV['LOCAL_PKG']
       pkg = File.basename(ENV['LOCAL_PKG'])
       scp_to(workstation, ENV['LOCAL_PKG'], pkg)
-    end
+    else
```
james-stocks: This `else` looks wrong - if you're using `LOCAL_PKG` then you still need the build version string for generic_install_dmg to get the pkg name inside the dmg

james-stocks:
@scotje I pushed a commit that moves the determination of SUITE_VERSION to Rakefile instead of doing it in the test code.
Also has a case where it can try to figure out SUITE_VERSION from the package filename if you're using LOCAL_PKG but didn't specify SUITE_VERSION. Note that SUITE_VERSION is needed for a LOCAL_PKG run because beaker needs to know the filename of the .pkg file inside the .dmg file; and that .pkg filename will have SUITE_VERSION in it

PR #226, opened 2017-08-07
(MAINT) Re-add package acceptance test for Gemfile.lock vendoring

scotje:


package-testing/tests/validate_a_new_module.rb
```diff
@@ -19,6 +19,10 @@
       on(workstation, "test -f #{module_name}/Gemfile.lock", accept_all_exit_codes: true) do |lock_check_outcome|
         assert_equal(0, lock_check_outcome.exit_code, 'pdk validate should have caused a Gemfile.lock file to be created')
       end
+
+      on(workstation, "cmp #{install_dir(workstation)}/share/cache/Gemfile.lock #{module_name}/Gemfile.lock", accept_all_exit_codes: true) do |gemfile_cmp|
+        assert_equal(0, gemfile_cmp.exit_code, 'newly created Gemfile.lock should match vendored Gemfile.lock')
```
james-stocks: If stdout or stderr has the diff; it would be valuable to add it to the assertion message here. 
I think beaker by default is not verbose on the output of host commands.

Then again, we might not care here what the differences are; just the fact that we've found a discrepancy between the cached Gemfile and the live one. In that case, the assertion string should describe why the mismatch is a test failure

james-stocks: @scotje added a commit for this, lemme know what you think and I can squash and merge

scotje: Looks good to me!

PR #228, opened 2017-08-08
(PDK-390) Implement spec:coverage rake task

DavidS:
This rake task will enable coverage tracking, execute the specs, and print a summary at the bottom of the test run.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -9,8 +9,10 @@ if ENV['COVERAGE'] == 'yes'
     Coveralls::SimpleCov::Formatter,
   ]
   SimpleCov.start do
+    track_files 'lib/**/*.rb'
     add_filter '/spec'
     # do not track vendored files
+    add_filter '/lib/puppet'
     add_filter '/vendor'
     add_filter '/.vendor'
   end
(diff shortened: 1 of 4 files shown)


DavidS:
Coverage reduction is expected, as it adds the lib/pdk/version.rb file under coverage that was ignored before.

It's a feature, not a bug!

spec/spec_helper.rb
```diff
@@ -1,10 +1,20 @@
...
+    Coveralls::SimpleCov::Formatter,
+  ]
   SimpleCov.start do
-    add_filter '/spec/'
+    track_files 'lib/**/*.rb'
+    add_filter '/spec'
+    # do not track vendored files
+    add_filter '/lib/puppet'
+    add_filter '/vendor'
+    add_filter '/.vendor'
```
james-stocks: I'm suspicious that there is no `add_filter '/package-testing'` but that folder doesn't show in the Coveralls report - are these add_filter statements redundant, given that track_files is specifying `lib/**/*.rb` ?

DavidS: track_files is the lower bar, add_filter excludes things that get used during execution, but should not be tracked. e.g. when running with bundler defaults, gems get installed into `vendor`, and would show up in reports, when they shouldn't.

james-stocks: Cool, so package-testing does not need excluded because it will never bring code into coverage scope during execution... 👍

PR #232, opened 2017-08-08
(PDK-388, PDK-392) New module generation improvements.

bmjen:
-Adds a puppet requirement to metadata.json
-Adds generating a skeleton README.md when generating new module
-Adds generating a skeleton CHANGELOG.md when generating new module

Requires https://github.com/puppetlabs/pdk-module-template/pull/24

### DIFF
--- a/lib/pdk/module/templatedir.rb
+++ b/lib/pdk/module/templatedir.rb
@@ -15,6 +15,8 @@ module PDK
       #
       # @param path_or_url [String] The path to a directory to use as the
       # template or a URL to a git repository.
+      # @param module_metadata [Hash] A Hash containing the module metadata.
+      # Defaults to an empty Hash.
       # @yieldparam self [PDK::Module::TemplateDir] The initialised object with
       # the template available on disk.
       #
@@ -34,7 +36,7 @@ module PDK
       # @raise [ArgumentError] (see #validate_module_template!)
       #
       # @api public
-      def initialize(path_or_url)
+      def initialize(path_or_url, module_metadata = {})
         if File.directory?(path_or_url)
           @path = path_or_url
         else
@@ -60,6 +62,8 @@ module PDK
         @moduleroot_dir = File.join(@path, 'moduleroot')
         @object_dir = File.join(@path, 'object_templates')
         validate_module_template!
+
+        @module_metadata = module_metadata
 
         yield self
       ensure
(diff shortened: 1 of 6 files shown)


bmjen:
Acceptance tests will fail until https://github.com/puppetlabs/pdk-module-template/pull/24 is merged.

lib/pdk/module/templatedir.rb
```diff
@@ -221,6 +223,7 @@ def config_for(dest_path)
           if File.file?(config_path) && File.readable?(config_path)
             begin
               @config = YAML.safe_load(File.read(config_path), [], [], true)
+              @config.merge!('global' => { 'userdata' => @userdata })
```
rodjek: Rather than merging this into the `@config` hash (which i'm planning to eventually deprecate), it would be cleaner (imho) to add this as the `module_metadata` key in the `file_config` hash, and then access it in the templates as `module_metadata['version']` etc.

lib/pdk/module/templatedir.rb
```diff
@@ -34,7 +34,7 @@ class TemplateDir
       # @raise [ArgumentError] (see #validate_module_template!)
       #
       # @api public
-      def initialize(path_or_url)
+      def initialize(path_or_url, userdata = {})
```
james-stocks: Also needs a yardoc entry

spec/acceptance/new_module_spec.rb
```diff
@@ -27,5 +27,15 @@
                                                                     'operatingsystemrelease' => ['8']))
       end
     end
+
+    describe file('foo/README.md') do
```
james-stocks: These new examples are dependent on the module template providing these files. I think it's more valuable to have these examples, with the risk of pdk-module-template breaking them; than to not have them 👍

bmjen:
New updates require: https://github.com/puppetlabs/pdk-module-template/pull/27/files to pass tests.

james-stocks:
Coveralls informs how many new lines of code are causing a coverage drop (in this case, 2 lines in lib/pdk/module/templatedir.rb ) : https://coveralls.io/builds/12769685

@bmjen we should at least have unit tests that satisfy coveralls.
If there's a need for higher levels of testing for generated content let me know. Possibly we need to address pdk cloning pdk-module-template master, to remove the tests being dependent on the state of that repo?

bmjen:
@james-stocks There is a ticket open to address defaulting to using the templates in the package. This *should* improve the pdk->pdk-module-template dependency a bit, however we do need to enact some structure on the pdk-module-template to aid in making sure compatibility isn't a problem. We'll need to start versioning the pdk-module-template repo and add version ranges in pdk to prevent breakage.

PR #235, opened 2017-08-09
[WIP] (maint) Add unit tests for PDK::CLI::Exec and children

rodjek:


DavidS:
These tests were fine when written, but now there seem to be other tests there. Can we either fix, or close this?

PR #236, opened 2017-08-09
(PDK-386) Remove parameter options from 'new class'

austb:


rodjek:
Rebased and modified the commit to remove a few more mentions of the class parameters from the docs.

PR #239, opened 2017-08-11
(PDK-401) Do not accept targets for metadata validator

james-stocks:
Only check metadata.json. If the user passes targets through the CLI, inform user they are being ignored.

### DIFF
--- a/lib/pdk/validators/metadata_validator.rb
+++ b/lib/pdk/validators/metadata_validator.rb
@@ -18,6 +18,11 @@ module PDK
       def self.invoke(report, options = {})
         exit_code = 0
 
+        if options[:targets] && options[:targets] != []
+          PDK.logger.info(_('metadata validator only checks metadata.json. The specified files will be ignored: %{targets}') % { targets: options[:targets].join(', ') })
+          options.delete(:targets)
+        end
+
         metadata_validators.each do |validator|
           exit_code = validator.invoke(report, options)
           break if exit_code != 0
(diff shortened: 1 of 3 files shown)


lib/pdk/validators/metadata_validator.rb
```diff
@@ -18,6 +18,11 @@ def self.metadata_validators
       def self.invoke(report, options = {})
         exit_code = 0
 
+        if options[:targets] && options[:targets] != []
+          PDK.logger.info(_('metadata validator only checks metadata.json. The specified files will be ignored: %{targets}') % { targets: options[:targets].join(', ') })
+          options.delete(:targets)
```
DavidS: does `options` get `dup`ed somewhere? otherwise this might destroy data for other callers.

bmjen: If we `dup` the options hash anywhere it should be in `cli/validate.rb` where `invoke` is called.

james-stocks: I was going to change this to remove :targets only in a copy of options passed to the two metadata validators.
It might be better to `dup` the options in `cli/validate.rb` to allow for further future special modifications to the options; cheers @bmjen

james-stocks:
https://github.com/puppetlabs/pdk/pull/248 may largely supersede this - it would report additional targets from the user as 'skipped'
It doesn't completely satisfy PDK-401 though, as it allows for _**/metadata.json_ to be validated, when this ticket specifies only the module root metadata.json should be checked

bmjen:
Integrated into #248

PR #240, opened 2017-08-11
(MAINT) Use non-forked tty-prompt gem

austb:


scotje:
Will need to bump this and `tty-spinner` in the vanagon packaging as well.

PR #241, opened 2017-08-11
(MAINT) Don't check coverage on gitignored files

scotje:
Not 100% if this will work in CI but we'll see in a minute!


scotje:
Hooray, it works!

PR #243, opened 2017-08-11
(PDK-397) Log output of bundler commands at appropriate levels

scotje:
Output of `bundle check` should go to DEBUG but non-zero exit status for the rest of the bundler commands represent fatal errors.


scotje:
Coveralls failure is because there was a net decrease of 5 lines in a covered file, coverage of that file is still 100% of relevant lines...

PR #245, opened 2017-08-12
(maint) Allow bundler to install gems in parallel

james-stocks:
Set jobs=2 when PDK calls bundle install to allow bundler to install gems in parallel.

It could be set higher than 2 but that might require testing to check it has no issues and even saves any time on workstations with less processing resource.

### DIFF
--- a/lib/pdk/util/bundler.rb
+++ b/lib/pdk/util/bundler.rb
@@ -103,7 +103,7 @@ module PDK
         end
 
         def install!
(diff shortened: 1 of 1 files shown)


james-stocks:
This doesn't make a noticeable difference on either travis or appveyor (though both systems are subject to variance, so this would be expected).
On my 2 core laptop; setting -j2 makes no difference to not setting it at all (about 37 seconds average in both cases).
Setting -j4 reduces the average to about 23 seconds.

PR #248, opened 2017-08-14
(PDK-401, PDK-402, PDK-403, PDK-404) Update validators to handle targets better

bmjen:
Updates validators to handle targets consistently and disallows input of
explicit targets that are incompatible with the validator's defined
filter pattern. Also updates the spinner text to be informative about what
kind of files the validator is validating against.

Relates to #239

### DIFF
--- a/lib/pdk/validators/base_validator.rb
+++ b/lib/pdk/validators/base_validator.rb
@@ -16,6 +16,17 @@ module PDK
         File.join(PDK::Util.module_root, 'bin', cmd)
       end
 
+      # Parses the target strings provided from the CLI
+      #
+      # @param options [Hash] A Hash containing the input options from the CLI.
+      #
+      # @return targets [Array] An Array of Strings containing target file paths
+      #                         for the validator to validate.
+      # @return skipped [Array] An Array of Strings containing targets
+      #                         that are skipped due to target not containing
+      #                         any files that can be validated by the validator.
+      # @return invalid [Array] An Array of Strings containing targets that do
+      #                         not exist, and will not be run by validator.
       def self.parse_targets(options)
         # If no targets are specified, then we will run validations from the
         # base module directory.
@@ -25,17 +36,32 @@ module PDK
                     options[:targets]
                   end
 
-        targets.map { |target|
+        skipped = []
+        invalid = []
+        matched = targets.map { |target|
           if respond_to?(:pattern)
             if File.directory?(target)
-              Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              target_list = Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              skipped << target if target_list.flatten.empty?
+              target_list
+            elsif File.file?(target)
+              if target.eql? pattern
+                target
+              elsif Array[pattern].flatten.map { |p| File.fnmatch(p, File.expand_path(target)) }.include? true
+                target
+              else
+                skipped << target
+                next
+              end
             else
-              target
+              invalid << target
+              next
             end
           else
             target
           end
-        }.flatten
+        }.compact.flatten
+        [matched, skipped, invalid]
       end
 
       def self.parse_options(_options, targets)
@@ -46,8 +72,37 @@ module PDK
         _('Invoking %{cmd}') % { cmd: cmd }
       end
 
+      def self.process_skipped(report, skipped = [])
+        skipped.each do |skipped_target|
+          PDK.logger.debug(_('%{validator}: Skipped \'%{target}\'. Target does not contain any files to validate (%{pattern}).') % { validator: name, target: skipped_target, pattern: pattern })
+          report.add_event(
+            file:     skipped_target,
+            source:   name,
+            message:  _('Target does not contain any files to validate (%{pattern}).') % { pattern: pattern },
+            severity: :info,
+            state:    :skipped,
+          )
+        end
+      end
+
+      def self.process_invalid(report, invalid = [])
+        invalid.each do |invalid_target|
+          PDK.logger.debug(_('%{validator}: Skipped \'%{target}\'. Target file not found.') % { validator: name, target: invalid_target })
+          report.add_event(
+            file:     invalid_target,
+            source:   name,
+            message:  'File does not exist.',
+            severity: :error,
+            state:    :error,
+          )
+        end
+      end
+
       def self.invoke(report, options = {})
-        targets = parse_targets(options)
+        targets, skipped, invalid = parse_targets(options)
+
+        process_skipped(report, skipped)
+        process_invalid(report, invalid)
 
         return 0 if targets.empty?
 
--- a/lib/pdk/validators/metadata/metadata_syntax.rb
+++ b/lib/pdk/validators/metadata/metadata_syntax.rb
@@ -42,7 +42,10 @@ module PDK
       end
 
       def self.invoke(report, options = {})
-        targets = parse_targets(options)
+        targets, skipped, invalid = parse_targets(options)
+
+        process_skipped(report, skipped)
+        process_invalid(report, invalid)
 
         return 0 if targets.empty?
 
@@ -57,18 +60,6 @@ module PDK
         JSON.parser = JSON::Pure::Parser
 
         targets.each do |target|
-          unless File.file?(target)
-            report.add_event(
-              file:     target,
-              source:   name,
-              state:    :failure,
-              severity: 'error',
-              message:  _('not a file'),
-            )
-            return_val = 1
-            next
-          end
-
           unless File.readable?(target)
             report.add_event(
               file: target,
--- a/lib/pdk/validators/ruby/rubocop.rb
+++ b/lib/pdk/validators/ruby/rubocop.rb
@@ -16,8 +16,12 @@ module PDK
         'rubocop'
       end
 
+      def self.pattern
+        '**/**.rb'
+      end
+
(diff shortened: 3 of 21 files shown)


bmjen:
I'll need to add some test coverage to the spec and acceptance tests. Will also need to integrate with @james-stocks 's metadata changes to complete the UX improvements on validation target handling.

lib/pdk/validators/base_validator.rb
```diff
@@ -25,17 +25,29 @@ def self.parse_targets(options)
                     options[:targets]
                   end
 
-        targets.map { |target|
+        skipped = []
```
james-stocks: It would be great to add yardoc to describe what the method now returns; and also to briefly explain what _skipped_ and _invalid_ mean

lib/pdk/validators/base_validator.rb
```diff
@@ -25,17 +25,29 @@ def self.parse_targets(options)
...
-              Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              target_list = Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              skipped << target if target_list.flatten.empty?
+              target_list
+            elsif File.file?(target)
+              if Array[pattern].flatten.map { |p| File.fnmatch(p, File.expand_path(target)) }.include? true
+                target
+              else
+                skipped << target
+                next
```
james-stocks: Is this `next` needed? It looks like there's only `end` lines below it

lib/pdk/validators/base_validator.rb
```diff
@@ -47,8 +59,29 @@ def self.spinner_text(_targets = nil)
...
       def self.invoke(report, options = {})
-        targets = parse_targets(options)
+        targets, skipped, invalid = parse_targets(options)
 
+        skipped.each do |skipped_target|
+          PDK.logger.debug(_("%{validator} skipped unmatched target: '%{target}'") % { validator: name, target: skipped_target })
+          report.add_event(
+            file:     skipped_target,
+            source:   name,
+            message:  'Target skipped. No files to validate.',
```
james-stocks: From the code above; I believe that _skipped_ means that the file/directory given by the user as a target does exist on the filesystem, but does not match the pattern built into the validator.
I'm not sure this message conveys that. And 'No files to validate' might be confusing if other targets in the command did have valid targets.

    message: 'Target was skipped as it does not match the validator\'s pattern'

?

lib/pdk/validators/base_validator.rb
```diff
@@ -25,17 +25,29 @@ def self.parse_targets(options)
...
                   end
 
-        targets.map { |target|
+        skipped = []
+        invalid = []
+        matched = targets.map { |target|
           if respond_to?(:pattern)
             if File.directory?(target)
-              Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              target_list = Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
```
james-stocks: At this point if the target directory contains a mixture of file types; we glob those that match the pattern and ignore those that don't.
We don't report those unmatching files as skipped... I think that's OK.

lib/pdk/validators/metadata/metadata_syntax.rb
```diff
@@ -42,7 +42,7 @@ def self.stop_spinner(exit_code)
       end
 
       def self.invoke(report, options = {})
-        targets = parse_targets(options)
+        targets, _skipped, _invalid = parse_targets(options)
```
james-stocks: This class does not report on `_skipped` or `_invalid` and doesn't seem to call `super.invoke` - should reporting of skipped and invalid targets be added?

lib/pdk/validators/metadata/metadata_json_lint.rb
```diff
@@ -26,7 +26,7 @@ def self.spinner_text(targets = [])
       end
 
       def self.pattern
-        'metadata.json'
```
james-stocks: [PDK-401|https://tickets.puppetlabs.com/browse/PDK-401] specifies _Only validate <module_root>/metadata.json_ so I think this pattern should be unchanged.

lib/pdk/validators/metadata/metadata_syntax.rb
```diff
@@ -11,7 +11,7 @@ def self.name
       end
 
       def self.pattern
-        'metadata.json'
```
james-stocks: [PDK-401](https://tickets.puppetlabs.com/browse/PDK-401) specifies _Only validate <module_root>/metadata.json_ so I think this pattern should be unchanged.

lib/pdk/validators/ruby/rubocop.rb
```diff
@@ -16,8 +16,12 @@ def self.cmd
         'rubocop'
       end
 
+      def self.pattern
+        '**/**.rb'
```
james-stocks: Rubocop checks a number of different files e.g. .irbrc, Rakefile, Gemfile - should this pattern definitely restrict to *.rb ?
I think for a Puppet module we want `{Rakefile,Gemfile,**/**.rb}` (or whatever the glob syntax for _or_ is )

lib/pdk/validators/metadata/metadata_syntax.rb
```diff
@@ -11,7 +11,7 @@ def self.name
...
+        '**/metadata.json'
```
james-stocks: [PDK-401](https://tickets.puppetlabs.com/browse/PDK-401) specifies _Only validate <module_root>/metadata.json_ so I think this pattern should be unchanged.

lib/pdk/validators/metadata/metadata_json_lint.rb
```diff
@@ -26,7 +26,7 @@ def self.spinner_text(targets = [])
...
+        '**/metadata.json'
```
james-stocks: [PDK-401](https://tickets.puppetlabs.com/browse/PDK-401) specifies _Only validate <module_root>/metadata.json_ so I think this pattern should be unchanged.

spec/acceptance/validate_metadata_spec.rb
```diff
@@ -100,7 +100,7 @@
       end
 
       its(:stdout) do
-        is_expected.to have_junit_testcase.in_testsuite('metadata-json-lint').with_attributes(
+        is_expected.not_to have_junit_testcase.in_testsuite('metadata-json-lint').with_attributes(
           'classname' => 'metadata-json-lint.dependencies',
           'name'      => 'broken.json',
         ).that_failed
```
james-stocks: .that_failed can be removed now that we're asserting the testcase does not exist

james-stocks: .that_failed can be removed now that we're asserting the testcase does not exist

spec/acceptance/validate_metadata_spec.rb
```diff
@@ -100,7 +100,7 @@
       end
 
       its(:stdout) do
-        is_expected.to have_junit_testcase.in_testsuite('metadata-json-lint').with_attributes(
+        is_expected.not_to have_junit_testcase.in_testsuite('metadata-json-lint').with_attributes(
```
james-stocks: May now want to assert that metadata.json was included in the test results

spec/support/it_accepts_metadata_json_targets.rb
```diff
@@ -17,28 +17,29 @@
...
       context 'and the module does not contain a metadata.json file' do
         it 'returns no targets' do
-          expect(parsed_targets).to eq([])
+          expect(parsed_targets.first).to eq([])
         end
       end
     end
 
-    context 'when given specific target files' do
+    context 'when given unmatched target files' do
```
james-stocks: Not sure I'd understand this immediately when reading.

Perhaps

    context 'when given unmatching target files' do

or

    context 'when given a target that will not match the validator\s pattern' do

?

spec/support/it_accepts_metadata_json_targets.rb
```diff
@@ -17,28 +17,29 @@
...
         targets.each do |target|
           allow(File).to receive(:directory?).with(target).and_return(false)
+          allow(File).to receive(:file?).with(target).and_return(true)
         end
       end
 
-      it 'returns the targets' do
-        expect(parsed_targets).to eq(targets)
+      it 'skips the targets' do
+        expect(parsed_targets.first).to eq([])
```
james-stocks: Might be valuable to also put an expectation on the _skipped_ and _invalid_ values

lib/pdk/validators/ruby/rubocop.rb
bmjen: Currently we have `Gemfile` and `Rakefile` excluded in `.rubocop.yml`

lib/pdk/validators/metadata/metadata_syntax.rb
```diff
@@ -42,7 +42,10 @@ def self.stop_spinner(exit_code)
       end
 
       def self.invoke(report, options = {})
-        targets, _skipped, _invalid = parse_targets(options)
+        targets, skipped, invalid = parse_targets(options)
+
+        process_skipped(skipped)
+        process_invalid(invalid)
```
austb: I think we need to pass the report object here. `process_skipped(report, skipped)` and `process_invalid(report, invalid)`

bmjen: 👍  good catch.

rodjek:
Squashed the fixup commits into their counterparts

PR #257, opened 2017-08-15
(maint) monkey patch TTY::Prompt::Reader::WinConsole to make it blocking

rodjek:
The non-blocking keypress read in the latest version of tty-prompt is causing issues with the module interview on windows - it's not waiting for any input and just using the default value.


rodjek:
Built a new windows package and confirmed that this will fix the issue until we can get it fixed upstream.

PR #262, opened 2017-08-17
(SDK-336) Add operating system question to the new module interview

rodjek:
WIP because I need to test this out on windows


DavidS:
Appveyor needed a rebuild to become green, after failing to git clone some dependencies.

PR #265, opened 2017-08-22
(PDK-430) Do not cache template-url answer if using the default template

rodjek:


lib/pdk/generators/module.rb
```diff
@@ -68,7 +68,7 @@ def self.invoke(opts = {})
           end
         end
 
-        PDK.answers.update!('template-url' => template_url)
+        PDK.answers.update!('template-url' => template_url) unless template_url == default_template_url
```
james-stocks: I don't get the intent of the original line - is this updating the answer for template_url with a local copy of the template instead of the initial remote URL?

DavidS: as implemented this change also means that there is no way to reset to using the default without deleting/editing the ansers.json file.

rodjek: @james-stocks yeah, the intent is that if a user specifies a custom template URL (for example, a company which has their own internal module template that they'd like to use), it will be remembered for subsequent `pdk new module` runs.

@DavidS good point!

PR #268, opened 2017-08-24
(maint) Debug output GEM_HOME and GEM_PATH before executing module commands

james-stocks:
To aid troubleshooting in cases of ruby environment doubts.

### DIFF
--- a/lib/pdk/cli/exec.rb
+++ b/lib/pdk/cli/exec.rb
@@ -217,6 +217,10 @@ module PDK
         def run_process!
           command_string = argv.join(' ')
           PDK.logger.debug(_("Executing '%{command}'") % { command: command_string })
+          if context == :module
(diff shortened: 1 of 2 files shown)


lib/pdk/cli/exec.rb
```diff
@@ -217,6 +217,9 @@ def run_process_in_clean_env!
         def run_process!
           command_string = argv.join(' ')
           PDK.logger.debug(_("Executing '%{command}'") % { command: command_string })
+          if context == :module
+            PDK.logger.debug(_("GEM_HOME is '%{gem_home}' and GEM_PATH is '%{gem_path}'") % { gem_home: @process.environment['GEM_HOME'], gem_path: @process.environment['GEM_PATH'] })
```
james-stocks: Hmm this should perhaps specify that these are the environment values for the command process

PR #269, opened 2017-08-24
(MAINT) Add strings to POT file

austb:


austb:
@jbondpdx I added a few more strings to POT file and updated it here.

PR #270, opened 2017-08-24
(PDK-365) Inform and prompt user following new module generate

bmjen:
Upon successful generation of a new module, inform the user:
- that the module has been generated successfully
- where it is located
- what to do next

### DIFF
--- a/lib/pdk/generators/module.rb
+++ b/lib/pdk/generators/module.rb
@@ -71,7 +71,10 @@ module PDK
         PDK.answers.update!('template-url' => template_url)
 
         begin
-          FileUtils.mv(temp_target_dir, target_dir)
+          if FileUtils.mv(temp_target_dir, target_dir)
+            PDK.logger.info(_('Module \'%{name}\' generated at path \'%{path}\'.') % { name: opts[:name], path: target_dir })
+            PDK.logger.info(_('In your new module directory, add classes with the \'pdk new class\' command.'))
+          end
         rescue Errno::EACCES => e
           raise PDK::CLI::FatalError, _("Failed to move '%{source}' to '%{target}': %{message}") % {
             source:  temp_target_dir,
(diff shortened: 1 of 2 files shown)


bmjen:
Going to add unit tests.

spec/unit/pdk/generate/module_spec.rb
```diff
@@ -24,6 +24,13 @@
   end
 end
 
+shared_context 'allow post generate info to be printed to stdout' do
+  before(:each) do
+    allow($stdout).to receive(:puts).with(a_string_matching(%r{module.*generated at path.*}i))
+    allow($stdout).to receive(:puts).with(a_string_matching(%r{In your new module directory, add classes with the 'pdk new class' command}i))
```
DavidS: shouldn't that be `expect post generate info...`?

bmjen: Yep. New updates should reflect actual testing. lol.

lib/pdk/generators/module.rb
```diff
@@ -71,7 +71,10 @@ def self.invoke(opts = {})
         PDK.answers.update!('template-url' => template_url)
 
         begin
-          FileUtils.mv(temp_target_dir, target_dir)
+          if FileUtils.mv(temp_target_dir, target_dir)
+            PDK.logger.info(_('Module \'%{name}\' generated at path \'%{path}\'.') % { name: opts[:name], path: target_dir })
+            PDK.logger.info(_('In your new module directory, add classes with the \'pdk new class\' command.'))
+          end
```
scotje: Is it possible for `FileUtils.mv` to fail without raising an exception?

rodjek: Not that I've seen

scotje: Sorry, I should have clarified the point of my question which was, why do we need that conditional around the log lines?

PR #275, opened 2017-08-28
[WIP] (PDK-443) Write templated files without carriage return on windows

rodjek:


bmjen:
There is a PR to the pdk-module-template for this issue as well: https://github.com/puppetlabs/pdk-module-template/pull/32

That PR will disable style checking for carriage returns. I'm not sure what the appropriate solution is for this issue. Simply generating a base file with no carriage returns won't prevent a Windows user from adding code that adds carriage returns right?

/cc @glennsarti

glennsarti:
@bmjen Correct.  The reverse is also true.  There's nothing stopping someone on a linux-y system added carriage returns either.

PR #278, opened 2017-08-31
(PDK-450) remove stdlib dependency

DavidS:
Having the stdlib dependency in there, but not in the rspec-puppet fixtures
leads to painful expectations: stdlib functions are available, but using
them in the code leads to test failures. Adding it to the fixtures would
cause the default PDK workflow to require a online connection, so this
instead removes the unforced dependency.

### DIFF
--- a/lib/pdk/generators/module.rb
+++ b/lib/pdk/generators/module.rb
@@ -101,9 +101,7 @@ module PDK
         defaults = {
           'name'         => "#{username}-#{opts[:name]}",
           'version'      => '0.1.0',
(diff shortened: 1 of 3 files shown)


james-stocks:
Code LGTM, and the approach of removing the dependency instead of adding more to the template to support it sound good; but the latter should maybe get one more +1

DavidS:
Yeah. For reference, https://tickets.puppetlabs.com/browse/PDK-460 also needs to be done to fully "solve" this issue.

PR #281, opened 2017-09-01
(PDK-461) Make Version.git_ref more forgiving

DavidS:
When there is no target `.git` directory, no need to call git at all.

Also adds more tests to cover everything.

### DIFF
--- a/lib/pdk/util/version.rb
+++ b/lib/pdk/util/version.rb
@@ -23,9 +23,12 @@ module PDK
       end
 
       def self.git_ref
-        ref_result = PDK::CLI::Exec.git('--git-dir', File.join(File.expand_path('../../..', File.dirname(__FILE__)), '.git'), 'describe', '--all', '--long')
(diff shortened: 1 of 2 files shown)


lib/pdk/util/version.rb
```diff
@@ -23,9 +23,12 @@ def self.pkg_sha
       end
 
       def self.git_ref
-        ref_result = PDK::CLI::Exec.git('--git-dir', File.join(File.expand_path('../../..', File.dirname(__FILE__)), '.git'), 'describe', '--all', '--long')
+        source_git_dir = File.join(File.expand_path('../../..', File.dirname(__FILE__)), '.git')
 
-        ref_result[:stdout].strip if ref_result[:exit_code].zero?
+        return '' unless File.directory?(source_git_dir)
```
rodjek: Can we return nil here instead of an empty string?

DavidS: yes

PR #287, opened 2017-09-08
(PDK-465) Add vendored git to PATH for package installs

rodjek:


DavidS:
Looks good to me, why the WIP?

rodjek:
@DavidS I was waiting to run tests with packages, which has now been done :)

DavidS:
I'm wondering if we should not move that to the end of PATH, to allow a user's real git install to provide extra configuration (e.g. authentication through SSH keys).

PR #293, opened 2017-09-13
(PDK-415) Convert user-input related problems from FATAL to ERROR

rodjek:
Adds a new exception type that should be raised when dealing with user input related problems (like trying to create a class that already exists).

```
asmodean :0: pdk/foo (git:pdk-415 → origin {2} ?:2!)$ ../bin/pdk new class foo
pdk (ERROR): Unable to generate class; '/home/tsharpe/code/puppet/pdk/foo/manifests/init.pp' already exists.
```

`PDK::CLI::FatalError` should still be used when catching and re-raising an exception, as those are almost always fatal errors and should be displayed as such.

### DIFF
--- a/lib/pdk/cli.rb
+++ b/lib/pdk/cli.rb
@@ -15,6 +15,10 @@ require 'pdk/util/version'
 module PDK::CLI
   def self.run(args)
     @base_cmd.run(args)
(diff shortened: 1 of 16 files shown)


DavidS:
Merged without coverage report

PR #298, opened 2017-09-20
(PDK-502) Add git mingw paths to PATH

rodjek:
Git on Windows doesn't seem to automatically find its component binaries, so we need to add them to the path in order for the vendored git to work properly.


rodjek:
@DavidS :+1: moved the vendored git path to the end of PATH so the user's git install will take precedence if present.

PR #301, opened 2017-09-26
(PDK-470) Validation of task metadata.

bmjen:
I've refactored the task metadata validation to validate against a json-schema per recommendation/request from @garethr .

This PR also includes functionality to extend the MetadataSyntax validator to validate the tasks JSON as well.

### DIFF
--- a/lib/pdk/validators/base_validator.rb
+++ b/lib/pdk/validators/base_validator.rb
@@ -42,14 +42,14 @@ module PDK
         matched = targets.map { |target|
           if respond_to?(:pattern)
             if File.directory?(target)
-              pattern_glob = Array[pattern].flatten.map { |p| Dir.glob(File.join(PDK::Util.module_root, p)) }
+              pattern_glob = Array(pattern).map { |p| Dir.glob(File.join(PDK::Util.module_root, p)) }
               target_list = pattern_glob.flatten.select { |file| File.fnmatch(File.join(File.expand_path(target), '*'), file) }
               skipped << target if target_list.flatten.empty?
               target_list
             elsif File.file?(target)
               if Array(pattern).include? target
                 target
-              elsif Array[pattern].flatten.map { |p| File.fnmatch(File.expand_path(p), File.expand_path(target)) }.include? true
+              elsif Array(pattern).any? { |p| File.fnmatch(File.expand_path(p), File.expand_path(target)) }
                 target
               else
                 skipped << target
--- a/lib/pdk/validators/base_validator.rb
+++ b/lib/pdk/validators/base_validator.rb
@@ -42,13 +42,14 @@ module PDK
         matched = targets.map { |target|
           if respond_to?(:pattern)
             if File.directory?(target)
-              target_list = Array[pattern].flatten.map { |p| Dir.glob(File.join(target, p)) }
+              pattern_glob = Array[pattern].flatten.map { |p| Dir.glob(File.join(PDK::Util.module_root, p)) }
+              target_list = pattern_glob.flatten.select { |file| File.fnmatch(File.join(File.expand_path(target), '*'), file) }
               skipped << target if target_list.flatten.empty?
               target_list
             elsif File.file?(target)
-              if target.eql? pattern
+              if Array(pattern).include? target
                 target
-              elsif Array[pattern].flatten.map { |p| File.fnmatch(p, File.expand_path(target)) }.include? true
+              elsif Array[pattern].flatten.map { |p| File.fnmatch(File.expand_path(p), File.expand_path(target)) }.include? true
                 target
               else
                 skipped << target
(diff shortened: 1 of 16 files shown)


bmjen:
I still need to fix-up the remaining spec tests as well as update the Acceptance tests to reflect the new changes to the info/debug output strings.

lib/pdk/validators/util/task_metadata_schema.rb
```diff
@@ -0,0 +1,65 @@
+module PDK
```
glennsarti: Is it possible to have the schema as a separate file instead of embedded?  That way other tools can consume it

glennsarti: Also means the schema file itself can be validated

scotje: There's been some discussion on tech-discuss and I've proposed hosting a JSON Schema for task (and module) metadata on the Forge. I've got a PR for that with just a placeholder so we could move this schema into that PR:

https://github.com/puppetlabs/puppet-forge-api/pull/371

glennsarti: Alternatively we could publish at http://schemastore.org/json/

lib/pdk/validators/base_validator.rb
```diff
@@ -45,7 +45,7 @@ def self.parse_targets(options)
               skipped << target if target_list.flatten.empty?
               target_list
             elsif File.file?(target)
-              if target.eql? pattern
+              if Array(pattern).flatten.include? target
```
rodjek: No need to call `flatten` after `Array()` unless `pattern` is going to be a multidimensional array (`Array([1, 2, 3])` => `[1, 2, 3]`)

lib/pdk/validators/util/task_metadata_schema.rb
glennsarti: Because two sources of truth is rarely a good thing.  Pick one.

scotje: Then I'll easily pick the one we have direct control over.

Seems like the best approach would be to host it ourselves and just add it to the schemastore catalog.

glennsarti: Done.

lib/pdk/validators/util/task_metadata_schema.rb
```diff
@@ -0,0 +1,65 @@
...
+  module Validate
+    module TaskMetadataSchema
+      SCHEMA = '{
```
DavidS: Please point here  to the canonical version of this schema, so that there can be no doubt about which copy is the primary.

lib/pdk/validators/metadata/task_metadata_lint.rb
```diff
@@ -0,0 +1,128 @@
...
+          schema = File.read(File.join(PDK::Util.package_cachedir, 'task.json'))
+        else
+          PDK.logger.debug(_('Task Metadata Schema was not found in the cache. Now downloading from the forge.'))
+          begin
+            schema = open('https://forgeapi.puppet.com/schemas/task.json').read
+          rescue
+            raise PDK::CLI::FatalError, _('Unable to download Task Metadata Schema file. Please check internet connectivity and retry this action.')
+          end
+        end
+        JSON.parse(schema)
```
scotje: Consider rescuing JSON parse error here (and logging a nicer error message) in case the schema is corrupted somehow

pdk.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'tty-spinner', '0.5.0'
   spec.add_runtime_dependency 'tty-prompt', '0.13.1'
   spec.add_runtime_dependency 'json_pure', '~> 2.1.0'
+  spec.add_runtime_dependency 'json-schema', '2.8.0'
```
scotje: Is there a ticket or PR for adding this to the vanagon repo?

scotje: (and it's runtime dependencies)

scotje:
Left a couple small comments but overall looking good @bmjen !

lib/pdk/validators/metadata/task_metadata_lint.rb
```diff
@@ -63,9 +63,20 @@ def self.schema_file
...
       def self.download_schema_from_forge
         PDK.logger.debug(_('Task Metadata Schema was not found in the cache. Now downloading from the forge.'))
+        ssl_failures ||= 0
         open(FORGE_SCHEMA_URL).read
-      rescue Exception => e
-        raise PDK::CLI::FatalError, _("Unable to download Task Metadata Schema file. Please check internet connectivity and retry this action. %{error}") % { error: e }
+      rescue OpenSSL::SSL::SSLError => e
+        ssl_failures += 1
+
+        unless ssl_failures < 2 && Gem.win_platform? && PDK::Util.gem_install?
```
bmjen: Does this mean that on a Windows system with a package install getting SSL errors, we'll retry forever?

scotje: See `ssl_retries` variable. Although we might need to do an explicit "begin" context instead of rescuing the whole method.

bmjen: yep, I see the counter. I misread the conditional itself.

PR #340, opened 2017-11-06
(PDK-626) Templatedir can now handle multiple directories

HelenCampbell:
Required for the new 'convert' functionality - The aim of this update is so that files needed for convert are kept in a default directory,
while new requires additional directories that will be found in the 'moduleroot_init'.

### DIFF
--- a/lib/pdk/generate/module.rb
+++ b/lib/pdk/generate/module.rb
@@ -70,7 +70,7 @@ module PDK
 
         template_url = opts.fetch(:'template-url', default_template_url)
 
-        PDK::Module::TemplateDir.new(template_url, metadata.data) do |templates|
+        PDK::Module::TemplateDir.new(template_url, metadata.data, true) do |templates|
           templates.render do |file_path, file_content|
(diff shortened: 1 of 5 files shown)


HelenCampbell:
This PR won't pass acceptance tests while https://github.com/puppetlabs/pdk-module-template/pull/56 isn't merged.

Both of these PRs will need merged at the same time as their functionality depends on one another.

DavidS:
The new code should not fail just because there is no `moduleroot_init` directory in the current template.

Once that is changed, the next question raised is how to test the additions without depending on "the" template. It'll cause more work now, but make the project more nimble, if you add a local example template for integration testing, that can be modified to exercise the code without external dependencies.

Finally, we still do not test the whole workflow end-to-end with pushing a new module to github/travis/appveyor. But that, again, is a new ticket.

HelenCampbell:
@DavidS The issue is that if we don't fail on no moduleroot_init being present, there is the potential that only half the files will be generated (whatever ends up being in moduleroot, assuming we are using the new template style), but I get your point.

Issue #336, opened 2017-10-31
Provide official docker image with pdk


gdubicki:
As various CI solution, such as GitLab CI, are using Docker containers to run the jobs inside them, it would be nice if you could provide official Docker image with pdk. Preferably it should be based on Alpine for minimal size.

DavidS:
Hi, we're tracking this in https://tickets.puppetlabs.com/browse/PDK-629 ; see https://github.com/sharumpe/puppet-pdk for some unofficial work in that area.

JacobHenner:
I created https://hub.docker.com/r/jacobhenner/puppet-test/ and https://hub.docker.com/r/jacobhenner/puppet-pdk/ based off of the Docker ruby images. I intend to tag them for alternative base images (e.g. Alpine), perhaps these are useful?

zoidbergwill:
@JacobHenner having images with tagged versions of pdk would be awesome.

DavidS:
There's also https://github.com/ccaum/puppet-dev-in-docker which has some of our tooling, and could use some love.

ccaum:
If anyone is interested in trying it out, I have a newer version of the puppet-dev-tools image here: https://hub.docker.com/r/ccaum/puppet-dev-tools/

It's based on CentOS so it can better support PDK.

zoidbergwill:
Is there a repo for that?

ccaum:
For now the repo is here: https://github.com/ccaum/puppet-dev-in-docker
I'll move it into the puppetlabs namespace once I'm happy with it.

I might split it into two images. One with just PDK, onceover, etc. and another built on it for the CD for PE specific bits. Feedback welcome.

zoidbergwill:
Ahh, had found that repo, but didn't think `ccaum/puppet-dev-tools` was from there for some reason. I was being a potato. That looks awesome.

bittner:
Is it so hard to provide an "official" Docker image, something we pull with [puppet/pdk](https://hub.docker.com/r/puppet/pdk)? This would really help standardize running PDK in CI environments.

A partial solution would be to provide a Docker image from the Vox Pupuli namespace (i.e. [voxpupuli/pdk](https://hub.docker.com/r/voxpupuli/pdk)) as a community effort, but then again PDK isn't really a @voxpupuli project.

TJM:
[...]

PR #342, opened 2017-11-08
Dockerize pdk

LongLiveCHIEF:
This PR adds a [docker](https://docs.docker.com) based offering of `pdk` for users, as an alternative to natively installing pdk (and all the pre-reqs that come with it). Some benefits of this are listed in the changes to the `README.md` included in this PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2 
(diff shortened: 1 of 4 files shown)


LongLiveCHIEF:
I've included the `dockerspec` gem for testing the dockerized build, but I'm not familiar enough to testing with rspec/ruby to know the best way to add tests. Should I add docker-based tests to each of the existing unit tests? (Some of this would be redundant). 

Or, should the docker distributed tests be acceptance tests?

DavidS:
There are already a number of existing pdk docker images (e.g. https://hub.docker.com/r/chriscowley/puppet-pdk/ , https://hub.docker.com/r/terzom/pdk/ , and https://hub.docker.com/r/mathow/centos-puppet-pdk/ ). While they, and your approach here, are fine as a stand-in, we want to provide a full PDK-in-docker experience using the complete deliverables (including a custom built ruby matching the agent, and all the cached content we ship in https://github.com/puppetlabs/pdk-vanagon) as official images.

The work for this is being tracked in https://tickets.puppetlabs.com/browse/PDK-629 . You can subscribe there to get notified of updates, when we get around to schedule it.

/cc #336

PR #343, opened 2017-11-08
(maint) Downgrade Bundler on TravisCI to 1.15.4

rodjek:
Travis updated bundler in their build environments to 1.16.0, but we specify Bundler ~> 1.15 as a dependency and this discrepancy causes Bad Things (tm) to happen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: false
 language: ruby
 bundler_args: "--without development"
+before_install:
+  - gem -v
+  - gem update --system 2.6.14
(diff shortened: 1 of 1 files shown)


DavidS:
> https://github.com/rubygems/rubygems/pull/2065/files

and 

> bundler 1.16.0 can't generate binstubs due to a bug in rubygems not installing the templates (see above) and we can't downgrade bundler on travis anymore because they're installing it as a default gem

rodjek:
RubyGems 2.7.2 was released ~10 hours ago and it looks like TravisCI has already updated their RVM builds :tada:

PR #354, opened 2017-11-16
(PDK-668) Templatedir now reads .sync.yml for config when rendering t…

HelenCampbell:
…emplates.

### DIFF
--- a/lib/pdk/module/templatedir.rb
+++ b/lib/pdk/module/templatedir.rb
@@ -224,10 +224,6 @@ module PDK
       # Generate a hash of data to be used when rendering the specified
       # template.
       #
-      # Read `config_defaults.yml` from the root of the template directory (if
-      # it exists) build a hash of values by merging the value of the `:global`
-      # key with the value of the key that matches `dest_path`.
-      #
       # @param dest_path [String] The destination path of the file that the
       # data is for, relative to the root of the module.
       #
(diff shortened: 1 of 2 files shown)


lib/pdk/module/templatedir.rb
```diff
@@ -236,26 +232,44 @@ def self.files_in_template(dirs)
...
-              @config = YAML.safe_load(File.read(config_path), [], [], true)
-            rescue StandardError => e
-              PDK.logger.warn(_("'%{file}' is not a valid YAML file: %{message}") % { file: config_path, message: e.message })
-              @config = {}
-            end
-          else
-            @config = {}
-          end
+        if @config.nil?
+          confdefaults = read_config(config_path)
```
bmjen: Tiny nitpick here, but just for consistency sake, `confdefaults` and `syncconfig` should be underscore separated like the other variables are.

lib/pdk/module/templatedir.rb
```diff
@@ -236,26 +232,44 @@ def self.files_in_template(dirs)
...
+          syncconfig = read_config(sync_config_path)
+          @config = confdefaults
+          @config.merge!(syncconfig) unless syncconfig.nil?
         end
-
         file_config = @config.fetch(:global, {})
         file_config['module_metadata'] = @module_metadata
         file_config.merge!(@config.fetch(dest_path, {})) unless dest_path.nil?
+        file_config.merge!(@config)
         file_config
```
bmjen: This line is no longer needed since the last line is no longer conditional.

spec/unit/pdk/module/template_dir_spec.rb
```diff
@@ -174,4 +174,47 @@
...
+       EOF
+      end
+      let(:yaml_hash) do
+        YAML.load(yaml_text) # rubocop:disable Security/YAMLLoad
+      end
+      let(:config_hash) do
+        YAML.load(config_defaults) # rubocop:disable Security/YAMLLoad
+      end
+
+      before(:each) do
```
bmjen: The `config_defaults` related allows should be moved up to the top level `before` block for maintenance purposes since they are required to be mocked with or without a `sync.yml` file.

lib/pdk/module/templatedir.rb
```diff
@@ -236,26 +232,44 @@ def self.files_in_template(dirs)
       #
       # @api private
       def config_for(dest_path)
```
bmjen: It would be cool if the path to `sync.yml` could be passed in as an optional parameter here, and if it is nil, then the code below can default the path to the `<dest_path>/sync.yml`. This will allow us to potentially add support for externally provided `sync.yml` config in the future.

lib/pdk/module/templatedir.rb
```diff
@@ -236,26 +232,44 @@ def self.files_in_template(dirs)
...
-        if @config.nil?
-          config_path = File.join(@path, 'config_defaults.yml')
+        sync_config_path = File.join(@path, '/.sync.yml')
```
bmjen: The `/` isn't necessary on a File.join. It will actually cause problems on Windows.

HelenCampbell:
@bmjen Updated with your changes!

PR #356, opened 2017-11-22
(maint) Update PDK::Test::Unit.parallel_with_no_tests? for PSH #216 changes

rodjek:
The behavour of the parallel_spec rake task in puppetlabs_spec_helper changed in https://github.com/puppetlabs/puppetlabs_spec_helper/pull/216. This PR updates our logic to detect no tests vs fail in `pdk test unit --parallel`.

### DIFF
--- a/lib/pdk/tests/unit.rb
+++ b/lib/pdk/tests/unit.rb
@@ -31,8 +31,8 @@ module PDK
 
       def self.parallel_with_no_tests?(ran_in_parallel, json_result, result)
         ran_in_parallel && json_result.empty? &&
-          !result[:exit_code].zero? &&
-          result[:stderr].strip =~ %r{Pass files or folders to run$}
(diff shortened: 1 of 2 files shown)


lib/pdk/tests/unit.rb
```diff
@@ -31,8 +31,8 @@ def self.rake(task, spinner_text, environment = {})
 
       def self.parallel_with_no_tests?(ran_in_parallel, json_result, result)
         ran_in_parallel && json_result.empty? &&
-          !result[:exit_code].zero? &&
-          result[:stderr].strip =~ %r{Pass files or folders to run$}
+          ((!result[:exit_code].zero? && result[:stderr].strip =~ %r{Pass files or folders to run$}) ||
+           result[:stderr].strip =~ %r{No files for parallel_spec to run against$})
```
scotje: If this gets much more complicated it may merit some small predicate helper methods to increase legibility

rodjek: Yeah, it's getting a bit nasty isn't it? :)

PR #368, opened 2017-12-01
(PDK-715) Use correct module template branch/ref

bmjen:
For gem installs, we need to make sure they use a compatible
module-template version, similar to what we do for package installs
in our build process. For development, we still want to use the
default git branch.

### DIFF
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -58,8 +58,13 @@ module PDK
     end
     module_function :package_install?
 
+    def development_mode?
+      !PDK::Util::Version.git_ref.nil?
+    end
+    module_function :development_mode?
+
     def gem_install?
-      !package_install?
+      !(package_install? || development_mode?)
     end
     module_function :gem_install?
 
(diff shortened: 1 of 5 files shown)


lib/pdk/util.rb
```diff
@@ -58,8 +58,13 @@ def package_install?
     end
     module_function :package_install?
 
+    def development_mode?
+      PDK::VERSION.end_with? 'pre'
+    end
+    module_function :development_mode?
```
scotje: Should this maybe test if the code is part of a git checkout or something instead?

bmjen: It could, but that could complicate matters if we (or others) attempt to run a tag from source for a repro or something? I think ultimately, the simplicity of just checking whether we're on an "official" version or a dev/preview version is more explicit. I could go either way.

scotje:
This seems like it breaks anyone using a custom template repo that doesn't have the hard coded version/ref. Does just adding and plumbing through a `--template-ref` option solve things more cleanly?

PR #376, opened 2017-12-05
(PDK-725) Add timestamp to PDK Convert Report

bmjen:


lib/pdk/module/convert.rb
```diff
@@ -95,6 +95,7 @@ def self.print_summary(update_manager)
 
       def self.fullreport(update_manager)
         File.open('convert_report.txt', 'w')
+        File.open('convert_report.txt', 'a') { |f| f.write("/* Convert Report generated by PDK at #{Time.now.getutc} */") }
         update_manager.changes[:modified].each do |_, diff|
           File.open('convert_report.txt', 'a') { |f| f.write("\n\n\n" + diff) }
         end
```
scotje: Is there a reason this needs to open and close file handles so much?

```ruby
File.open('convert_report.txt', 'w') do |f|
  f.write("/* Convert Report generated by PDK at #{Time.now.getutc} */")
  
  update_manager.changes[:modified].each do |_, diff|
    f.write("\n\n\n" + diff)
  end
end
```

lib/pdk/module/convert.rb
```diff
@@ -95,6 +95,7 @@ def self.print_summary(update_manager)
 
       def self.fullreport(update_manager)
         File.open('convert_report.txt', 'w')
+        File.open('convert_report.txt', 'a') { |f| f.write("/* Convert Report generated by PDK at #{Time.now.getutc} */") }
```
scotje: Is UTC preferable to local time in this case?

PR #377, opened 2017-12-05
(PDK-728) Add default_template_ref handler.

bmjen:


lib/pdk/module/templatedir.rb
```diff
@@ -49,13 +49,7 @@ def initialize(path_or_url, module_metadata = {}, init = false)
...
           temp_dir = PDK::Util.make_tmpdir_name('pdk-module-template')
 
-          clone_result = if PDK::Util.package_install?
-                           PDK::Util::Git.git('clone', path_or_url, temp_dir)
-                         elsif PDK::Util.gem_install?
-                           PDK::Util::Git.git('clone', path_or_url, '--branch', PDK::TEMPLATE_VERSION, temp_dir)
-                         else # TODO: Once we decide on a branching strategy for the pdk-module-template, this needs to be updated.
-                           PDK::Util::Git.git('clone', path_or_url, '--branch', 'convert', temp_dir)
-                         end
+          clone_result = PDK::Util::Git.git('clone', path_or_url, '--branch', PDK::Util.default_template_ref, temp_dir)
```
scotje: We should consider making this a `clone` and then `reset --hard` to the desired ref. `clone --branch` only supports branches and tags. This is probably fine for now though.

bmjen: Probably a good idea. Going to hold off on merging this and potentially breaking the vanagon build.

rodjek: :+1: to clone && reset

PR #379, opened 2017-12-06
(PDK-719) Check for module directory layout when converting

rodjek:
Falls back to checking for `manifests/`, `lib/`, `tasks/`, `functions/`, and
`types/` if the module does not contain a `metadata.json`. This logic is not
enabled by default as we should only need to do this during convert (after
that, a `metadata.json` file should be present).

`templates/`, `files/`, `spec/` are not checked for, as it's unlikely they'll
exist without one of the other directories being present.

### DIFF
--- a/lib/pdk/cli/convert.rb
+++ b/lib/pdk/cli/convert.rb
@@ -12,7 +12,7 @@ module PDK::CLI
 
     run do |opts, _args, _cmd|
       require 'pdk/module/convert'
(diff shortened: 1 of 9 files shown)


bmjen:
Tested this PR and on a module that does not have a metadata.json, pdk convert will go through the interview questionnaire, then exit with this error:

```
pdk (ERROR): Unable to open 'metadata.json' for reading
```

PR #397, opened 2017-12-19
(PDK-736) Improve handling of old template-url and template-ref

scotje:
Previously, if the user had the old puppetlabs/pdk-module-template URL
saved in their answers.json, PDK would still try to use the old repo but
attempt to reset to the `1.3.0` tag which doesn't exist in the old repo.
This commit starts treating a value matching the old repo URL the same
as if there is no template URL specified so that it uses the new
template URL instead.

Furthermore, if the user had a real custom template repo, we were still
attempting to reset to the `1.3.0` tag which probably won't exist in
their custom repo. After this commit, pdk will only try to reset to the
pinned tag when using the official template repo (or packaged repo).

### DIFF
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -173,11 +173,15 @@ module PDK
     module_function :targets_relative_to_pwd
 
     def default_template_url
-      if !PDK.answers['template-url'].nil?
-        PDK.answers['template-url']
-      else
-        puppetlabs_template_url
-      end
+      answer_file_url = PDK.answers['template-url']
+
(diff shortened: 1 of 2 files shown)


HelenCampbell:
Tested this with "template-url": "https://github.com/puppetlabs/pdk-module-template" in the answers file. 
Seems to pick up the correct version of the template fine. 
Did a convert on ntp and was successful with "template-url": "https://github.com/puppetlabs/pdk-templates" and "template-ref": "1.3.0-0-g632fa63" showing in the metadata.
Will create a separate ticket for the cached url being incorrect. Aside from that, merging this.
Thanks Jessie!

PR #400, opened 2018-01-08
(PDK-739) Fall back to default template if necessary

rodjek:
Checks if the repo pointed to by the saved template-url from the answer file contains valid refs and fall back if it doesn't. `git ls-remote` will return non-zero (or specifically 2) if the remote does not exist or does not contain the specified ref (if supplied).

### DIFF
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -181,6 +181,11 @@ module PDK
       return puppetlabs_template_url if answer_file_url == 'https://github.com/puppetlabs/pdk-module-template'
       return puppetlabs_template_url if answer_file_url == puppetlabs_template_url
 
+      unless PDK::Util::Git.repo_exists?(answer_file_url)
+        PDK.logger.warn(_("Unable to access the previously used template '%{template}', using the default template instead.") % { template: answer_file_url })
+        return puppetlabs_template_url
+      end
+
       answer_file_url
     end
     module_function :default_template_url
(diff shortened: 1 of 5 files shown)


bmjen:
@rodjek I've pushed an update to prevent the warning from getting spammed during a module creation b/c 'default_template_url' is called a few times.

```
[~/proj/sdk/test_modules]$ pdk new module foo
pdk (WARN): Unable to access the previously used template 'http://github.com/mjen/pdk-templates', using the default template instead.
pdk (WARN): Unable to access the previously used template 'http://github.com/mjen/pdk-templates', using the default template instead.
pdk (WARN): Unable to access the previously used template 'http://github.com/mjen/pdk-templates', using the default template instead.
pdk (WARN): Unable to access the previously used template 'http://github.com/mjen/pdk-templates', using the default template instead.
pdk (WARN): Unable to access the previously used template 'http://github.com/mjen/pdk-templates', using the default template instead.
pdk (WARN): Unable to access the previously used template 'http://github.com/mjen/pdk-templates', using the default template instead.
pdk (INFO): Creating new module: foo
```

I've modified it to update the answers file to remove the template-url the first time we detect an invalid url.

This is the new output.

```
[~/proj/sdk/test_modules]$ pdk new module foo
pdk (WARN): Unable to access the previously used template 'http://github.com/mjen/pdk-templates', using the default template instead.
pdk (INFO): Creating new module: foo
``

PR #402, opened 2018-01-12
(PDK-575) Run puppet parser validate with an dummy empty puppet.conf

rodjek:
To prevent any user or system level puppet.conf settings from interfering with the behaviour of the PDK, pass `/dev/null` or `NUL` (depending on the platform) as the puppet.conf when running `puppet parser validate`.

### DIFF
--- a/spec/unit/pdk/validate/puppet_syntax_spec.rb
+++ b/spec/unit/pdk/validate/puppet_syntax_spec.rb
@@ -24,6 +24,10 @@ describe PDK::Validate::PuppetSyntax do
     let(:options) { {} }
     let(:targets) { %w[target1 target2.pp] }
 
+    before(:each) do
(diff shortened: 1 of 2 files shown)


spec/unit/pdk/validate/puppet_syntax_spec.rb
```diff
@@ -36,7 +40,7 @@
       let(:options) { { auto_correct: true } }
 
       it 'has no effect' do
-        expect(command_args).to eq(%w[parser validate].concat(targets))
+        expect(command_args).to eq(%w[parser validate --config /dev/null].concat(targets))
```
scotje: Is the purpose of this existing `auto_correct` test assertion to make sure that some option is not being added? If so should we maybe just assert that the option is absent rather than having to update this test every time the invocation changes?

Also should we have a new test that explicitly checks for the `--config` param?

Issue #399, opened 2017-12-21
convert fails non-latest Puppet version


jeremy-asher:
The `pdk convert` command crashes when using versions of Puppet other than the latest.

I'm not aware if there is an alternative way to set the Puppet version when using PDK, but I have been using the standard environment variable.  For example:

```
[root@4e57747fe881 ntp]# export PUPPET_GEM_VERSION=5.3.2
```

This results in a crash when using `pdk convert` due to the cached `Gemfile.lock` which is currently built using Puppet 5.3.3.  This could probably fixed with a `bundle update` instead of a `bundle install`.

<details>
<summary>pdk convert crash (running against the NTP module)</summary>

```
[root@4e57747fe881 ntp]# pdk convert --debug
pdk (DEBUG): Using 'private/git/bin/git' from PDK package.
pdk (DEBUG): Using 'private/git/bin/git' from PDK package.
pdk (DEBUG): Executing '/opt/puppetlabs/pdk/private/git/bin/git clone file:///opt/puppetlabs/pdk/share/cache/pdk-templates.git /tmp/pdk-templates20171221-1393-1cpjebe'
pdk (DEBUG): Execution of '/opt/puppetlabs/pdk/private/git/bin/git clone file:///opt/puppetlabs/pdk/share/cache/pdk-templates.git /tmp/pdk-templates20171221-1393-1cpjebe' complete (duration: 0.084376892s; exit code: 0)
pdk (DEBUG): Using 'private/git/bin/git' from PDK package.
pdk (DEBUG): Using 'private/git/bin/git' from PDK package.
pdk (DEBUG): Executing '/opt/puppetlabs/pdk/private/git/bin/git -C /tmp/pdk-templates20171221-1393-1cpjebe reset --hard 1.3.0'
pdk (DEBUG): Execution of '/opt/puppetlabs/pdk/private/git/bin/git -C /tmp/pdk-templates20171221-1393-1cpjebe reset --hard 1.3.0' complete (duration: 0.006936146s; exit code: 0)
pdk (DEBUG): Using 'private/git/bin/git' from PDK package.
pdk (DEBUG): Using 'private/git/bin/git' from PDK package.
pdk (DEBUG): Executing '/opt/puppetlabs/pdk/private/git/bin/git --git-dir /tmp/pdk-templates20171221-1393-1cpjebe/.git describe --all --long --always'
[...]

PR #413, opened 2018-01-29
 Avoid environment variables that could confuse bundler during conversion

DavidS:


DavidS:
As described in #399, this will only paper over the underlying issues, and in the most important case, `test unit`, it will be of no use at all.

I'll re-upload the two infrastructure commits here separately.

PR #417, opened 2018-02-06
(PDK-754) Interview for missing or Forge only metadata before build

rodjek:
When the user triggers a module build via the CLI, check for any missing metadata fields (forge specific things like `source`, `issues_url`, etc) and interview the user for these specific values.


bmjen:
I opened #422 with a rebase and updates to accommodate --force. 

I don't have push access to your fork. :)

PR #418, opened 2018-02-06
Revert "(maint) pin pdk-templates version ref to workaround puppet 5.…

bmjen:
…3.4 issue"

This reverts commit e9ad92557da58097891a3d09a721eaf5dc3f5963.


scotje:
Should we create an upstream issue in puppetlabs/puppet to make `parser validate` respect `--render-as json`?

rodjek:
@scotje PDK-800 has already been created to track that :)

PR #425, opened 2018-02-12
(PDK-804) Fixes error in build without ignore file

bmjen:
pdk will now make sure that if there is no ignore file and if the pkg
target-dir is in the module_dir, then it will add the target-dir
to the ignores list.

### DIFF
--- a/lib/pdk/module/build.rb
+++ b/lib/pdk/module/build.rb
@@ -16,8 +16,8 @@ module PDK
       attr_reader :force
 
       def initialize(options = {})
-        @module_dir = options[:module_dir] || Dir.pwd
-        @target_dir = options[:target_dir] || File.join(module_dir, 'pkg')
+        @module_dir = File.expand_path(options[:module_dir] || Dir.pwd)
+        @target_dir = File.expand_path(options[:'target-dir'] || File.join(module_dir, 'pkg'))
       end
 
       # Read and parse the values from metadata.json for the module that is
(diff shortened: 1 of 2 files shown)


lib/pdk/module/build.rb
```diff
@@ -196,6 +196,13 @@ def ignored_files
 
                              PathSpec.new(data)
                            end
+
+        # Also ignore the target directory if it is in the module dir and not already ignored
+        if Find.find(@module_dir).include?(@target_dir) && !@ignored_files.match(File.basename(@target_dir) + '/')
```
rodjek: Rather than calling `Find.find` here (which might get expensive), could we get away with something like:

```ruby
if @ignored_files.specs_matching(target_dir).empty?
  @ignored_files.add("\/#{File.basename(target_dir)\/")
end
```

lib/pdk/module/build.rb
```diff
@@ -196,6 +196,13 @@ def ignored_files
...
+          @ignored_files = @ignored_files.add("\/#{File.basename(@target_dir)}\/")
```
rodjek: Should use `target_dir` reader method rather than referencing the instance variable directly (here and in the line above), makes it easier to mock when unit testing.

lib/pdk/module/build.rb
bmjen: That wouldn't guard against the scenario where `target-dir` wasn't a sub directory of the `module_dir`. If `target-dir` was `/tmp/pkg/` then `@ignored_files.specs_matching('/tmp/pkg/').empty?` would be `true`, and `/pkg/` would get added to the ignored_files a second time.

PR #430, opened 2018-02-16
Update validation regex and error message for module name question

ardrigh:
PDK does not correctly validate module names if a module name is not supplied to `pdk new module`.

The user question prompts for a module name but fails to pass underscores as valid.

- `pdk new module test_name` will pass
- `pdk new module` will fail on `test_name` given as the module name.

This update uses the regex [referred to in the docs](https://puppet.com/docs/puppet/5.4/modules_fundamentals.html#module-names) for validating module names. 

I have included a similar error message to the options validation line, while referring to "numbers" as the docs do.


bmjen:
Great find @ardrigh and thanks for the contribution!

PR #448, opened 2018-03-08
(PDK-785) Add --puppet-version and --pe-version CLI options

rodjek:
The important bit: It adds the `--puppet-version` and `--pe-version` options to the `pdk validate` and `pdk test unit` commands (it's a lot of changed lines for such a simple change, but this PR lays a lot of the groundwork for the rest of the puppet & ruby version switching behaviour).

 * PE versions are mapped to Puppet versions in a static hash (copied from the forge), though this will eventually be made dynamic.
 * `GEM_PATH` and `GEM_HOME` have been extracted from `PDK::CLI::Exec` into a separate class, allowing them to be reused when scanning for available Puppet versions (and eventual ruby version switching).
 * Available Puppet versions are determined by scanning `GEM_HOME` and `GEM_PATH` for puppet gemspec files and parsing their contents.
 * Puppet version selection will prefer an exact match if available, otherwise it'll use the latest patch release for the specified version.

I would recommend not being too worried about the internals of `PDK::Util::PuppetVersion` and `PDK::Util::RubyVersion` at this stage, as they will change as this feature evolves in further PRs.

### DIFF
--- a/lib/pdk/cli/test/unit.rb
+++ b/lib/pdk/cli/test/unit.rb
@@ -7,6 +7,7 @@ module PDK::CLI
     usage _('unit [options]')
     summary _('Run unit tests.')
 
+    PDK::CLI.puppet_version_options(self)
     flag nil, :list, _('List all available unit test files.')
     flag nil, :parallel, _('Run unit tests in parallel.'), hidden: true
     flag :v, :verbose, _('More verbose output. Displays examples in each unit test file.')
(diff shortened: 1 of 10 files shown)


lib/pdk/cli/test/unit.rb
```diff
@@ -21,6 +22,11 @@ module PDK::CLI
     run do |opts, _args, _cmd|
       require 'pdk/tests/unit'
 
+      if opts[:'puppet-version'] && opts[:'pe-version']
+        PDK.logger.error _('You can not specify both --puppet-version and --pe-version at the same time.')
+        exit 1
```
scotje: Should this raise `PDK::CLI::ExitWithError` instead?

lib/pdk/cli/validate.rb
```diff
@@ -32,6 +33,11 @@ module PDK::CLI
         exit 0
       end
 
+      if opts[:'puppet-version'] && opts[:'pe-version']
+        PDK.logger.error _('You can not specify both --puppet-version and --pe-version at the same time.')
+        exit 1
```
scotje: Should this raise `PDK::CLI::ExitWithError` instead?

lib/pdk/cli/test/unit.rb
rodjek: Good call

PR #457, opened 2018-03-22
(MAINT) Add Ruby 2.5 to Travis and Appveyor config

scotje:
Also moves "primary" Ruby version in Travis to 2.4.3

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,29 +1,48 @@
 # version: 1.0.{build}-{branch}
 
 environment:
+  LOG_SPEC_ORDER: true
   matrix:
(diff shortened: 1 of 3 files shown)


scotje:
Ruby 2.5 cells will be red until #459 is merged.

appveyor.yml
```diff
@@ -2,6 +2,8 @@
 
 environment:
   matrix:
```
Iristyle: can we add `DEBUG_RESOLVER=1` to the environment?

That should show us a whole ton of output for the dependency resolution

scotje:
Need to split the `path_or_url` stuff out into it's own PR and clean up a couple other things, but this is mostly working now!

scotje:
Should go green after #462 is merged.

PR #461, opened 2018-03-27
(PDK-840) Add PDK::Util::PuppetVersion.from_module_metadata

rodjek:
For use when selecting the default Puppet & Ruby version if the user doesn't specify one

### DIFF
--- a/lib/pdk/util/puppet_version.rb
+++ b/lib/pdk/util/puppet_version.rb
@@ -6,7 +6,7 @@ module PDK
       class << self
         extend Forwardable
 
-        def_delegators :instance, :find_gem_for, :from_pe_version
+        def_delegators :instance, :find_gem_for, :from_pe_version, :from_module_metadata
 
         attr_writer :instance
 
@@ -20,17 +20,17 @@ module PDK
         version = Gem::Version.new(version_str)
 
         exact_requirement = Gem::Requirement.create(version)
-        gem_version = find_gem(exact_requirement)
-        return gem_version.version unless gem_version.nil?
+        found_gem = find_gem(exact_requirement)
+        return found_gem unless found_gem.nil?
 
         latest_requirement = Gem::Requirement.create("#{version.approximate_recommendation}.0")
-        gem_version = find_gem(latest_requirement)
-        unless gem_version.nil?
+        found_gem = find_gem(latest_requirement)
+        unless found_gem.nil?
           PDK.logger.info _('Unable to find Puppet %{requested_version}, using %{found_version} instead') % {
             requested_version: version_str,
-            found_version:     gem_version.version,
+            found_version:     found_gem[:gem_version].version,
           }
-          return gem_version.version
+          return found_gem
         end
 
         raise ArgumentError, _('Unable to find a Puppet version matching %{requirement}') % {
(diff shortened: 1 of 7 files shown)


lib/pdk/util/ruby_version.rb
```diff
@@ -6,26 +6,62 @@ class RubyVersion
...
+            @active_ruby_version = version
+          else
+            raise ArgumentError, _('Unknown Ruby version "%{ruby_version}"') % {
+              ruby_version: version,
+            }
+          end
+        end
+
+        def scan_for_packaged_rubies
+          { '2.4.3' => '2.4.0' }
```
DavidS: The contents do not match the name of the function. What gives?

spec/unit/pdk/util/puppet_version_spec.rb
```diff
@@ -24,9 +24,22 @@
   shared_context 'is a package install' do
     before(:each) do
       allow(PDK::Util).to receive(:package_install?).and_return(true)
+      allow(PDK::Util::RubyVersion).to receive(:versions).and_return('2.1.9' => '2.1.0', '2.4.3' => '2.4.0')
```
DavidS: This does not match actual behaviour of this method

spec/unit/pdk/util/puppet_version_spec.rb
```diff
@@ -24,9 +24,22 @@
...
 
-      mock_response = cache_versions.map { |r| Gem::Version.new(r) }
-      allow(PDK::Util::RubyVersion).to receive(:available_puppet_versions).and_return(mock_response)
+      PDK::Util::RubyVersion.use('2.1.9')
+      instance219 = PDK::Util::RubyVersion.instance
+      PDK::Util::RubyVersion.use('2.4.3')
+      instance243 = PDK::Util::RubyVersion.instance
```
DavidS: This feels like it could benefit from a `.instance(version)`, instead of requiring two calls?

spec/unit/pdk/util/puppet_version_spec.rb
```diff
@@ -167,38 +202,45 @@
       include_context 'is not a package install'
       include_context 'with a mocked rubygems response'
 
+      def result(version)
+        {
+          gem_version:  Gem::Version.new(version),
+          ruby_version: PDK::Util::RubyVersion.default_ruby_version,
```
DavidS: Is this a test of the `default_ruby_version`, too, or are you using it here only as a shortcut?

lib/pdk/util/puppet_version.rb
```diff
@@ -59,6 +59,36 @@ def from_pe_version(version_str)
         find_gem_for(gem_version[:gem_version])
       end
 
+      def from_module_metadata(metadata)
```
DavidS: shouldn't this parsing be located in the `PDK::Module::Metadata` class?

lib/pdk/util/ruby_version.rb
bmjen: +1

lib/pdk/util/puppet_version.rb
rodjek: It could go in either really, but I opted to contain it with all the other puppet version finding logic for clarity (imho). I can move it to be a method of `PDK::Module::Metadata` if that's the group consensus though.

spec/unit/pdk/util/puppet_version_spec.rb
rodjek: `PDK::Util::RubyVersion.default_ruby_version` will return the details of the running ruby version for non-packaged installs (gem or git clone), so I guess shortcut?

spec/unit/pdk/util/puppet_version_spec.rb
rodjek: :+1:

lib/pdk/util/ruby_version.rb
rodjek: Yeah, that was method was just moved around in the file, I was going to implement it in the next PR.

spec/unit/pdk/util/puppet_version_spec.rb
rodjek: It does now (that method was just a stub that was going to be implemented in the next PR).

lib/pdk/util/puppet_version.rb
rodjek: I've moved the validation logic into `PDK::Module::Metadata` and left the version finding logic in `PDK::Util::PuppetVersion`

PR #462, opened 2018-03-27
(MAINT) Refactor templatedir path_or_url calculation

scotje:
May be worth adding unit tests for the new helper methods...


scotje:
Maybe the newly broken-out methods should move into `PDK::Util::Git`?

PR #465, opened 2018-04-04
(maint) Update bundler before build because Ruby 2.5

DavidS:


DavidS:
Unbreak build for everyone.

PR #331, opened 2017-10-23
(maint) expose `git` as an experimental subcommand

DavidS:
Fixes #328

PR #466, opened 2018-04-04
(maint) Allow `pdk bundle` to work without `--`

DavidS:
By upgrading to the new version of CRI, the `skip_option_parsing`
call can be used to avoid the previous confusion of bundle vs pdk
options.

### DIFF
--- a/pdk.gemspec
+++ b/pdk.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.1.9'
 
   spec.add_runtime_dependency 'bundler', '~> 1.15'
-  spec.add_runtime_dependency 'cri', '~> 2.9.1'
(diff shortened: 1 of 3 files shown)


DavidS:
This is a prerequisite for #331

pdk.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.1.9'
 
   spec.add_runtime_dependency 'bundler', '~> 1.15'
-  spec.add_runtime_dependency 'cri', '~> 2.9.1'
+  spec.add_runtime_dependency 'cri', '~> 2.10.1'
```
bmjen: This is throw-in. I'd prefer if this was taken out and a new ticket created to update CRI, so we can update the vanagon component at the same time.

PR #472, opened 2018-04-11
(PDK-921) Update PDK::Util::Bundler helpers to support gem switching

scotje:
Also refactored the PDK::Util::Bundler unit tests while I was in there, that work is mostly in the second commit but could be smooshed together if it doesn't seem worth keeping separate.


scotje:
Yeah, sorry about that. It helps a bit to look at each of the commits separately. Thanks for reviewing!

PR #480, opened 2018-04-18
(PDK-842) Wire puppet-version and pe-version options into subcommands

scotje:
Unless otherwise stated, packaged install stuff requires packages built from the HEAD of master on pdk-vanagon.

### DIFF
--- a/lib/pdk/util/bundler.rb
+++ b/lib/pdk/util/bundler.rb
@@ -9,8 +9,11 @@ module PDK
     module Bundler
       class BundleHelper; end
 
-      def self.ensure_bundle!(gem_overrides = {})
+      def self.ensure_bundle!(gem_overrides = nil)
         bundle = BundleHelper.new
+
+        # This will default ensure_bundle! to re-resolving everything to latest
+        gem_overrides ||= { puppet: nil, hiera: nil, facter: nil }
 
         if already_bundled?(bundle.gemfile, gem_overrides)
           PDK.logger.debug(_('Bundler managed gems already up to date.'))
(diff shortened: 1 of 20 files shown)


rodjek:
@scotje I've made some changes to get the existing unit and acceptance tests to pass with these changes. The only major behavioural change that you might want to review is in 284c693

lib/pdk/util/bundler.rb
```diff
@@ -22,19 +25,33 @@ def self.ensure_bundle!(gem_overrides = {})
...
-        unless bundle.installed?(gem_overrides)
+        # Update puppet-related gem dependencies by re-resolving them specifically.
+        # If there are additional dependencies that aren't available locally, allow
+        # `bundle lock` to reach out to rubygems.org
+        bundle.update_lock!(gem_overrides, local: all_deps_available)
+
+        # If there were missing dependencies when we checked above, let `bundle install`
+        # go out and get them. For packaged installs, this should only be true if the user
+        # has added custom gems that we don't vendor.
+        unless bundle.installed?
```
scotje: Hmm, that's odd. If `all_deps_available` was true then `bundle lock --update` shouldn't reach out to rubygems for updates since we are passing the `--local` flag. This shouldn't hurt anything though but it is curious.

rodjek:
So, this is what it'd look like with `--puppet-version`/`--pe-version` as global flags for use in `pdk bundle`. This command being special and skipping argument parsing would require the option to come between `pdk` and `bundle` (as everything after `bundle` is considered an argument).

```
semirhage :0: pdk/foo (git:842_its_aliveeeee → scotje U:4 ?:1!)$ ../bin/pdk --puppet-version 4.7.0 bundle exec puppet --version
pdk (INFO): Using Ruby 2.1.10
pdk (INFO): Using Puppet 4.7.0
[✔] Installing missing Gemfile dependencies.
4.7.0

semirhage :0: pdk/foo (git:842_its_aliveeeee → scotje U:4 ?:1!)$ ../bin/pdk --puppet-version 5.5.0 bundle exec puppet --version
pdk (INFO): Using Ruby 2.1.10
pdk (INFO): Using Puppet 5.5.0
5.5.0
```

I'm honestly not sure if it's worth it to have the specific option ordering for this command vs using an env var to set the desired puppet version for `pdk bundle`... I guess consistency would be the main point in its favour. I'll leave it for you @scotje and @bmjen to discuss and squash or drop the commit as appropriate :)

scotje:
My vote would be to recognize values from `PDK_PUPPET_VERSION` and `PDK_PE_VERSION` environment variables and *not* make the `--puppet-version` and `--pe-version` options global.

My reasoning:

- Long term I think we should be aiming to remove the `bundle` subcommand. It's an escape hatch for use cases we haven't yet considered or addressed but part of the whole reason for PDK existing is to make it unnecessary for people to even be aware of Bundler.
- Making option strings more positional is bad for UX.
- Having those options show up in the help output for subcommands where they have no impact is also bad for UX.

lib/pdk/util/bundler.rb
```diff
@@ -93,42 +115,45 @@ def installed?(gem_overrides = {})
...
-            # Update the vendored lock with any overrides
-            update_lock!(gem_overrides, local: true) unless gem_overrides.empty?
           else
             argv = ['lock']
 
             cmd = bundle_command(*argv).tap do |c|
-              c.add_spinner(_('Resolving Gemfile dependencies.'))
-              c.update_environment(gemfile_env(gem_overrides)) unless gem_overrides.empty?
+              c.add_spinner(_('Resolving default Gemfile dependencies.'))
+              # c.update_environment(gemfile_env(gem_overrides)) unless gem_overrides.empty?
```
bmjen: This commented needs to be removed or a reason for leaving it should be documented.

scotje: Oh yeah I've been meaning to remove this line.

PR #483, opened 2018-04-24
(maint) Remove static PE version map from PDK::Util::PuppetVersion

rodjek:
This ended up an uglier change than I expected as the format of the JSON object from the forge API changed between the initial implementation and now. I've tried to keep the changes logically separated by commit though for easier reviewing.

### DIFF
--- a/lib/pdk/util/vendored_file.rb
+++ b/lib/pdk/util/vendored_file.rb
@@ -1,0 +1,83 @@
+require 'pdk/util'
+require 'net/https'
+require 'openssl'
+require 'fileutils'
+
+module PDK
+  module Util
+    class VendoredFile
+      class DownloadError < StandardError; end
+
+      HTTP_ERRORS = [
+        EOFError,
+        Errno::ECONNRESET,
+        Errno::EINVAL,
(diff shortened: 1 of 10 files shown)


scotje:
Had some concerns about requiring stable network access for the unit tests and also the explicit indirect testing of the private method, so I vendored a copy of the version mapping and moved the test cases into the existing suites for `.from_pe_version`

lib/pdk/util/vendored_file.rb
```diff
@@ -0,0 +1,83 @@
...
+          raise DownloadError, _('Unable to download %{url}. %{code}: %{message}.') % {
+            url:     url,
+            code:    response.code,
+            message: response.message,
+          }
+        end
+
+        response.body
+      rescue *HTTP_ERRORS => e
+        raise DownloadError, _('Unable to download %{url}. Check internet connectivity and try again. %{error}') % {
```
bmjen: Missing the `url` argument in the subsequent hash.

scotje: fixed

PR #494, opened 2018-05-01
(PDK-802) Work around OpenSSL multi-threading errors when needed

scotje:
For post-1.5.0 bug bashing consideration.

See https://bugs.ruby-lang.org/issues/11033 for background.

I think this is the cause of our "cert already in hash table" transients in Appveyor as well as some of the issues @glennsarti was encountering in https://tickets.puppetlabs.com/browse/PDK-802

Ruby issue linked above was backported to Ruby 2.3 branch and released in 2.3.5, however Appveyor is still on 2.3.3 for Ruby23.

Inside PDK packages it should only be an issue on the 2.1.9 bundler invocations.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,15 +7,19 @@ environment:
     # - RUBY_VERSION: 25-x64
     #   USE_MSYS: true
     #   SUITES: "spec"
+    #   BUNDLE_JOBS: 4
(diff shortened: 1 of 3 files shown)


scotje:
Oops, my unit test expectations are wrong, will fix tomorrow.

PR #504, opened 2018-05-08
(PDK-956) Remove empty array values from metadata.json when rendering

rodjek:
Just like we remove `nil` values before rendering. In the reverse direction, these values will automatically be converted back to empty arrays as `.from_file` populates the object with the defaults merged with the values read from the file.

### DIFF
--- a/lib/pdk/module/metadata.rb
+++ b/lib/pdk/module/metadata.rb
@@ -70,7 +70,10 @@ module PDK
       end
 
       def to_json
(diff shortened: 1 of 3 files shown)


scotje:
Unfortunately `dependencies` is a required key in the metadata. (https://puppet.com/docs/puppet/5.2/modules_metadata.html#available-metadatajson-keys)

scotje:
Decided to close this for now until there is an agreed upon solution to the issues raised in https://tickets.puppetlabs.com/browse/PUP-6964

PR #515, opened 2018-05-29
(PDK-636) Make fixture cleaning optional

rodjek:
To make the iterative workflow faster, this (along with puppetlabs/puppetlabs_spec_helper#242) changes the behaviour of `pdk test unit` so that it doesn't automatically remove the downloaded fixtures after the test run. For CI purposes (or if you really want to slow things down) `--clean-fixtures` can be passed to `pdk test unit` to restore the old behaviour.

```
semirhage :0: pdk/foo (git:master → origin U:2 ?:1!)$ rm -rf spec/fixtures/modules/
semirhage :0: pdk/foo (git:master → origin U:2 ?:1!)$ time ../bin/pdk test unit
pdk (INFO): Using Ruby 2.4.4
pdk (INFO): Using Puppet 5.5.1
[✔] Preparing to run the unit tests.
[✔] Running unit tests.
  Evaluated 4 tests in 0.149583489 seconds: 0 failures, 0 pending.

real	0m32.575s
user	0m8.906s
sys	0m1.328s
semirhage :0: pdk/foo (git:master → origin U:2 ?:1!)$ time ../bin/pdk test unit
pdk (INFO): Using Ruby 2.4.4
pdk (INFO): Using Puppet 5.5.1
[✔] Preparing to run the unit tests.
[✔] Running unit tests.
  Evaluated 4 tests in 0.147905862 seconds: 0 failures, 0 pending.

real	0m9.732s
user	0m6.779s
sys	0m0.782s
```

### DIFF
--- a/lib/pdk/tests/unit.rb
+++ b/lib/pdk/tests/unit.rb
@@ -7,7 +7,7 @@ module PDK
   module Test
     class Unit
       def self.cmd(tests, opts = {})
-        rake_args = opts.key?(:parallel) ? 'parallel_spec' : 'spec'
+        rake_args = opts.key?(:parallel) ? 'parallel_spec_standalone' : 'spec_standalone'
         rake_args += "[#{tests}]" unless tests.nil?
         rake_args
       end
@@ -89,7 +89,7 @@ module PDK
 
         result[:exit_code]
       ensure
-        tear_down
+        tear_down if options[:'clean-fixtures']
       end
 
       def self.parse_output(report, json_data)
--- a/lib/pdk/tests/unit.rb
+++ b/lib/pdk/tests/unit.rb
@@ -57,6 +57,8 @@ module PDK
 
         return if result[:exit_code].zero?
 
+        tear_down
+
         PDK.logger.error(_('The spec_prep rake task failed with the following error(s):'))
         print_failure(result, _('Failed to prepare to run the unit tests.'))
       end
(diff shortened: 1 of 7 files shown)


scotje:
Looks like there are some spec tests that are expecting the message about fixture cleanup.

lib/pdk/tests/unit.rb
```diff
@@ -89,7 +89,7 @@ def self.invoke(report, options = {})
 
         result[:exit_code]
       ensure
-        tear_down
+        tear_down if options[:'clean-fixtures']
```
hunner: Users forgetting tear-down and then running `pdk build` would end up bundling their fixtures. Until `pdk release prep` can clean the fixtures, I would suggest having `pdk build` run a `spec_clean` or `tear_down` or whatever to avoid bundling fixtures.

rodjek: @hunner good call!

scotje:
CI will be blocked on meta-gems release I think?

rodjek:
Yep

lib/pdk/tests/unit.rb
bmjen: @rodjek has this been ticketed so we don't forget to update pdk build? I'm wondering if we should add the fixtures directory to the .pdkignore vs calling something that will change a tagged repo/directory before building.

brettjacobson:
Please get this Released soon!  I have 19 unit tests for my module, which take 10 seconds to execute.  But the entire execution takes about 4 minutes, because of prep time downloading the 6 dependent modules.  I really want to write more unit tests, but this process is not giving me "fast feedback".

PR #512, opened 2018-05-24
fix templating deep directory structure

jeremy-asher:
This fixes an ordering bug with the file enumeration that caused directories to
be unintentionally be included.  This allows arbitrary subdirectory structures
to be templated as expected.

Issue #445, opened 2018-03-01
Support for deep directory structure in templates


rtib:
I tried to create a template containing `moduleroot_init/spec/nodesets` directory with some nodes definitions for an acceptance test wireframe. The `pdk new module` command ended up with being unable to create directory structures deeper than one level:
```
pdk (DEBUG): Rendering 'spec/nodesets'...
pdk (FATAL): Failed to render template 'spec/nodesets'
ArgumentError: '/var/folders/f2/s1z9f_qd24n417_4w8q8c2fr0000gn/T/pdk-templates20180301-3694-h8fiix/moduleroot_init/spec/nodesets' is not a readable file
pdk (DEBUG): ArgumentError: '/var/folders/f2/s1z9f_qd24n417_4w8q8c2fr0000gn/T/pdk-templates20180301-3694-h8fiix/moduleroot_init/spec/nodesets' is not a readable file
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/template_file.rb:62:in `template_content'
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/template_file.rb:86:in `render_plain'
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/template_file.rb:43:in `render'
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/module/templatedir.rb:131:in `block in render'
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/module/templatedir.rb:126:in `each'
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/module/templatedir.rb:126:in `render'
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/generate/module.rb:58:in `block in invoke'
pdk (DEBUG): /opt/local/lib/ruby2.4/gems/2.4.0/gems/pdk-1.4.1/lib/pdk/module/templatedir.rb:82:in `initialize'
[...]

PR #524, opened 2018-06-14
Correct template path filter logic to only include regular files

nabertrand:
It looks like the original logic in PDK::Module::TemplateDir.files_in_template was created to filter paths that weren't regular files, but that logic was altered in commit https://github.com/puppetlabs/pdk/commit/7facba3fa940ec3a2015deacad9420481a005ad3#diff-f1961fa3f901b19ee585c1962e910d5eL193 and the filter no longer works as exptected. If a directory other than "spec" appears in the moduleroot or moduleroot_init folders, `pdk new module` fails with the error
```
pdk (FATAL): Failed to render template '<directory>'
ArgumentError: '/tmp/pdk-templates20180614-14695-18x9jy1/moduleroot_init/<directory>' is not a readable file
```

This patch restores the original behavior.


nabertrand:
It looks like this is same issue as #445 but solves the problem slightly differently than #512. #512 will create a possibly unbalanced dirlocs list @ https://github.com/puppetlabs/pdk/pull/512/files#diff-f1961fa3f901b19ee585c1962e910d5eL216

thomas-illiet:
Thanks  !!!

PR #528, opened 2018-06-21
fix templating deep directory structure

DavidS:
This refactors `files_in_templates` to fix an ordering bug with the file
enumeration that caused directories to be unintentionally be included.  This
allows arbitrary subdirectory structures to be templated as expected.

### DIFF
--- a/lib/pdk/module/templatedir.rb
+++ b/lib/pdk/module/templatedir.rb
@@ -104,8 +104,8 @@ module PDK
       #
       # @api public
       def render
-        PDK::Module::TemplateDir.files_in_template(@dirs).each do |template_file, template_loc|
-          template_file = template_file.to_s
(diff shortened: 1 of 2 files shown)


DavidS:
@jeremy-asher I've re-pushed your #512 PR here as a new one as travis seemed to not want to cooperate on the other PR. It's probably best for you to continue your work in a new PR (thanks to github ACLs you won't be able to update this one here).

Please leave #512 open, as I'll contact travis' support about that one.

DavidS:
Of course, a soon as I say it, travis catches up on #512. Seems like we all need more :coffee: :-D

So, @jeremy-asher , apologies for all the notifications, but it seems like we can continue as normal on #512, and I'll clean up here as soon as travis is finished.

scotje:
Dupe of #512 which was resolved by #524

PR #540, opened 2018-07-06
Allow use hieradata and manifests folders in tpl

thomas-illiet:
To allow add a default manifest & hieradata folders in custom "pdk-module-template", and also allows to add default configuration files in these folders.


thomas-illiet:
Hello @rodjek, this request has been well supported ?

PR #554, opened 2018-07-23
(PDK-1077) Expand the package acceptance test suite

rodjek:
To cover the manual tests that we run on packages leading up to a release.

Also, this adds a `acceptance:testenv` rake task that quickly spins up an environment on vmpooler with PDK pre-installed.


scotje:
@rodjek Is there a way to have the `acceptance:testenv` task output the hostname of the VM before it exits?

Everything else looks good!

scotje:
Updated to cleanup the `rake acceptance:testenv` output a bit

PR #559, opened 2018-08-03
(PDK-1096, PDK-1097, PDK-1098) Add puppet-dev flag to validate and test unit

bmjen:
Adds the `--puppet-dev` flag to `validate` and `test unit`. This flag will allow users to get Puppet from github source.

### DIFF
--- a/lib/pdk/cli/util.rb
+++ b/lib/pdk/cli/util.rb
@@ -90,7 +90,6 @@ module PDK
         begin
           puppet_env =
             if use_puppet_dev
-              PDK::Util::PuppetVersion.fetch_puppet_dev
               PDK::Util::PuppetVersion.puppet_dev_env
             elsif desired_puppet_version
               PDK::Util::PuppetVersion.find_gem_for(desired_puppet_version)
--- a/lib/pdk/cli/util.rb
+++ b/lib/pdk/cli/util.rb
@@ -132,6 +132,21 @@ module PDK
         pe_ver_specs << '--pe-version option' if opts[:'pe-version']
         pe_ver_specs << 'PDK_PE_VERSION environment variable' if ENV['PDK_PE_VERSION'] && !ENV['PDK_PE_VERSION'].empty?
 
+        puppet_dev_specs = []
+        puppet_dev_specs << '--puppet-dev flag' if opts[:'puppet-dev']
+        puppet_dev_specs << 'PDK_PUPPET_DEV environment variable' if ENV['PDK_PUPPET_DEV'] && !ENV['PDK_PUPPET_DEV'].empty?
+
+        puppet_dev_specs.each do |pup_dev_spec|
+          [puppet_ver_specs, pe_ver_specs].each do |offending|
+            next if offending.empty?
+
+            raise PDK::CLI::ExitWithError, _('You cannot specify a %{first} and %{second} at the same time') % {
+              first: pup_dev_spec,
+              second: offending.first,
+            }
+          end
+        end
+
         puppet_ver_specs.each do |pup_ver_spec|
           next if pe_ver_specs.empty?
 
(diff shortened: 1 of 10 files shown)


spec/unit/pdk/cli/validate_spec.rb
```diff
@@ -221,4 +221,32 @@
       }.to exit_zero
     end
   end
+
+  context 'with --puppet-dev' do
+    it 'activates resolved puppet version' do
+      expect {
+        PDK::CLI.run(['validate', '--puppet-dev'])
+      }.to exit_zero
+    end
```
scotje: This test isn't actually asserting anything about activating the right puppet version yet, I assume this that part hasn't been implemented in this PR. But maybe it would be better to leave this out or mark it as pending so that we don't accidentally leave it this way later?

bmjen: Yeah, good point. I was going to address it in the next PR when I do get the puppet-version selection parts tested.

lib/pdk/util/puppet_version.rb
```diff
@@ -16,6 +17,19 @@ def instance
       end
 
       PE_VERSIONS_URL = 'https://forgeapi.puppet.com/private/versions/pe'.freeze
+      DEFAULT_PUPPET_DEV_URL = 'https://github.com/puppetlabs/puppet'
```
scotje: `.freeze` the dev URL as well?

lib/pdk/util/puppet_version.rb
```diff
@@ -16,6 +17,19 @@ def instance
...
+
+      def puppet_dev_env
+        {
+          gem_version: 'file://%{path}' % { path: puppet_dev_path },
+          ruby_version: PDK::Util::RubyVersion.default_ruby_version,
+        }
+      end
+
+      def puppet_dev_path
+        '%{cache}/src/puppet' % { cache: PDK::Util.cachedir }
```
scotje: Should this use `File.join`?

lib/pdk/util/puppet_version.rb
```diff
@@ -27,6 +41,25 @@ def latest_available
...
 
+      def fetch_puppet_dev
+        unless PDK::Util::Git.remote_repo? puppet_dev_path
+          FileUtils.mkdir_p puppet_dev_path
+          clone_result = PDK::Util::Git.git('clone', DEFAULT_PUPPET_DEV_URL, puppet_dev_path)
+          unless clone_result[:exit_code].zero?
+            PDK.logger.error clone_result[:stdout]
+            PDK.logger.error clone_result[:stderr]
+            raise PDK::CLI::FatalError, _("Unable to clone git repository at '%{repo}'.") % { repo: DEFAULT_PUPPET_DEV_URL }
+          end
```
scotje: Might be worth returning early after the clone since the subsequent `fetch` is unlikely to bring in any new commits.

lib/pdk/cli/util.rb
```diff
@@ -83,12 +83,16 @@ def module_version_check
...
       def puppet_from_opts_or_env(opts)
+        use_puppet_dev = (opts || {})[:'puppet-dev'] || ENV['PDK_PUPPET_DEV']
         desired_puppet_version = (opts || {})[:'puppet-version'] || ENV['PDK_PUPPET_VERSION']
         desired_pe_version = (opts || {})[:'pe-version'] || ENV['PDK_PE_VERSION']
 
         begin
           puppet_env =
-            if desired_puppet_version
+            if use_puppet_dev
+              PDK::Util::PuppetVersion.fetch_puppet_dev
```
scotje: I don't think it's optimal for `puppet_from_opts_or_env` to have the side effect of cloning/fetching the repo every time it's called. Could this be exposed separately and maybe this method can raise if the local repo doesn't exist?

scotje: Yeah, it currently is a 1-1 correlation with the places you want to call it, but I think it's just unclear that calling that method will have that side effect, so I'd rather see explicit calls in those places.

bmjen: Ah, that's fair. I can make that change and at this part, I'll just call `.puppet_dev_env` to get the hash.

PR #566, opened 2018-08-24
Update lib/pdk/module/metadata.rb with latest OS versions / Archlinux

bastelfreak:
This is based on a Twitter conversation: https://twitter.com/dev_el_ops/status/1032895068707930112

### DIFF
--- a/lib/pdk/module/metadata.rb
+++ b/lib/pdk/module/metadata.rb
@@ -6,8 +6,12 @@ module PDK
     class Metadata
       attr_accessor :data
 
+<<<<<<< HEAD
       include PDK::Util::Filesystem
 
+=======
+      # rubocop:disable Style/WordArray
+>>>>>>> (FIXUP) Resolve rubocop issues
       OPERATING_SYSTEMS = {
         'RedHat based Linux' => [
           {
@@ -58,9 +62,10 @@ module PDK
           'operatingsystemrelease' => ['2008 R2', '2012 R2', '10'],
         },
         'Arch Linux' => {
-          'operatingsystem'        => 'Archlinux'
-        }
+          'operatingsystem'        => 'Archlinux',
+        },
       }.freeze
+      # rubocop:enable Style/WordArray
 
       DEFAULT_OPERATING_SYSTEMS = [
         'RedHat based Linux',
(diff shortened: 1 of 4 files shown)


lib/pdk/module/metadata.rb
```diff
@@ -54,12 +54,16 @@ class Metadata
...
+        'Archlinux' => {
+          'operatingsystem'        => 'Archlinux'
+        }
       }.freeze
 
       DEFAULT_OPERATING_SYSTEMS = [
         'RedHat based Linux',
         'Debian based Linux',
         'Windows',
+        'Archlinux',
```
scotje: I'd like to leave the default list as is for now, in the near future you will be able to customize this with local config.

lib/pdk/module/metadata.rb
```diff
@@ -54,12 +54,16 @@ class Metadata
           'operatingsystem'        => 'windows',
           'operatingsystemrelease' => ['2008 R2', '2012 R2', '10'],
         },
+        'Archlinux' => {
```
scotje: The user friendly name appears to be "Arch Linux" so we should use that as the key here.

lib/pdk/module/metadata.rb
bastelfreak: Well I had to try to sneak it in :D

scotje:
@bastelfreak thanks for the contribution to PDK!

Looks like the CI failures are due to some hard coded expectations in the unit tests around what the OS support metadata should look like. I can probably take a look at those next week if needed.

lib/pdk/module/metadata.rb
bastelfreak: updated

bastelfreak:
can you recheck? You should already have write perms to my feature branch.

scotje:
Oops, OK I'll see what I can do :)

scotje:
OK, hopefully that commit fixes up the unit test failures.

SimonHoenscheid:
Windows checks are failing, but it seems like test preconditions are not met. @scotje any idea how to fix this?

scotje:
We are going to take a pass at updating the defaults here as part of PDK-1366 which will likely integrate many of these changes.

PR #568, opened 2018-08-30
(PDK-1056) Update default ruby for packaged versions

bmjen:
Simple change here to ensure that the packaged version of PDK will default to 2.4.4, but also adds a helper function to identify the newest version of Ruby in the package.

### DIFF
--- a/lib/pdk/util/ruby_version.rb
+++ b/lib/pdk/util/ruby_version.rb
@@ -45,7 +45,14 @@ module PDK
         end
 
         def default_ruby_version
+          # For now, the packaged versions will be using default of 2.4.4.
+          return '2.4.4' if PDK::Util.package_install?
+
           # TODO: may not be a safe assumption that highest available version should be default
+          latest_ruby_version
+        end
+
+        def latest_ruby_version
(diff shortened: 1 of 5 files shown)


lib/pdk/util/ruby_version.rb
```diff
@@ -45,10 +45,17 @@ def scan_for_packaged_rubies
         end
 
         def default_ruby_version
+          # For now, the packaged versions will not be using default of 2.4.4.
+          '2.4.4' if PDK::Util.package_install?
```
rodjek: I think you meant to `return '2.4.4'` here

lib/pdk/util/ruby_version.rb
```diff
@@ -45,10 +45,17 @@ def scan_for_packaged_rubies
...
+
           # TODO: may not be a safe assumption that highest available version should be default
           versions.keys.sort { |a, b| Gem::Version.new(b) <=> Gem::Version.new(a) }.first
```
rodjek: Should change this line to just call `latest_ruby_version` now to get rid of the duplicated code

PR #575, opened 2018-10-03
(PDK-1100) Exclude known artifacts from build instead of cleaning

rodjek:
By handling it in this manner, we can remove the need to call the PDK shipped `spec_clean` task. This is the only dependency on the PDK runtime environment when calling `PDK::Module::Build.invoke` from outside of PDK (as needs to be done in puppetlabs_spec_helper's `build` rake task).

### DIFF
--- a/lib/pdk/module/build.rb
+++ b/lib/pdk/module/build.rb
@@ -8,6 +8,16 @@ require 'pdk/tests/unit'
 module PDK
   module Module
     class Build
+      DEFAULT_IGNORED = [
+        '/pkg/',
+        '.*',
+        '~*',
+        '/coverage',
+        '/checksums.json',
+        '/REVISION',
+        '/spec/fixtures/modules/',
+      ].freeze
+
       def self.invoke(options = {})
         new(options).build
       end
@@ -37,7 +47,6 @@ module PDK
       #
       # @return [String] The path to the built package file.
       def build
-        cleanup_module
         create_build_dir
 
         stage_module_in_build_dir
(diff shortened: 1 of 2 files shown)


lib/pdk/module/build.rb
```diff
@@ -215,6 +213,8 @@ def ignored_files
               ignored = ignored.add("\/#{File.basename(target_dir)}\/")
             end
 
+            DEFAULT_IGNORED.each { |r| ignored.add(r) }
```
bmjen: This will add duplicate entries to the ignore list if there are common ignores between the pmt/pdk/git ignore files and the defaults.

lib/pdk/module/build.rb
```diff
@@ -83,17 +92,6 @@ def cleanup_build_dir
         FileUtils.rm_rf(build_dir, secure: true)
       end
 
-      # Clean up any files created during use of the PDK that shouldn't be part
-      # of the built module (e.g. test fixtures).
-      #
-      # @return nil
-      def cleanup_module
-        PDK::Util::Bundler.ensure_bundle!
-        PDK::Util::Bundler.ensure_binstubs!('rake')
```
scotje: Does this need to re-appear somewhere?

lib/pdk/module/build.rb
rodjek: Duplicates shouldn't cause an issue (just as you can have duplicate entries in a `.gitignore`).

lib/pdk/module/build.rb
rodjek: Nope, the purpose of this change is to remove the need for the in-module bundler environment when building release tarballs

lib/pdk/module/build.rb
bmjen: True, we don't currently clear dupes from the ignore file either.

PR #579, opened 2018-10-24
(PDK-1200) Fix bundle env handling with puppet-dev

bmjen:
Bundler requires that you continue to pass in PUPPET_GEM_VERSION
when doing check, install, and exec. Previously we were only setting
the puppet gem env when doing a bundle check and updating the lock file.

### DIFF
--- a/spec/acceptance/version_changer_spec.rb
+++ b/spec/acceptance/version_changer_spec.rb
@@ -26,9 +26,47 @@ describe 'puppet version selection' do
       end
     end
 
-    describe command('pdk validate --puppet-dev') do
-      its(:stderr) { is_expected.to match(%r{Using Puppet file://}i) }
-      its(:exit_status) { is_expected.to eq(0) }
+    describe 'puppet-dev uses the correct puppet env' do
+      before(:all) do
+        File.open(File.join('manifests', 'init.pp'), 'w') do |f|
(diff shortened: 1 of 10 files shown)


spec/acceptance/version_changer_spec.rb
```diff
@@ -30,5 +30,45 @@
...
+describe 'version_select', :if => Gem::Specification.find_by_name('puppet').source.options.key?('path') do
+  on_supported_os.each do |os, os_facts|
+    context 'on #{os}' do
+      let(:facts) { os_facts }
+
+      it { is_expected.to compile }
+    end
+  end
+end
+          TESTFILE
```
scotje: This method seems a little round-about to me...

Could we update the test manifest to have two specific tests that assert either the presence or absence of the `path` key and then below use `pdk test unit --tests=` to only invoke the specific test and assert that it passed?

rodjek:
Should this change also be applied to `pdk validate` so that `PDK::Validate::PuppetSyntax` can run against the dev version as well?

spec/acceptance/version_changer_spec.rb
bmjen: I've updated the test to expect the path... so the acceptance test was simplified to expect a PDK test success or failure.

PR #580, opened 2018-10-24
(MAINT) Replace `gem which bundler` with Gem::Specification lookup

scotje:


scotje:
Looks like this needs more tweaking...

scotje:
Seems to so far be something specific to the Travis/RVM setup, still trying to replicate the issue on a VM.

PR #584, opened 2018-10-29
(PDK-1093) Replace null values and empty data structures in metadata when converting

rodjek:
Currently when we're updating a module's `metadata.json` file during
convert, we only replace values if they are completely absent from the
file. Unfortunately, this means if - for example - the existing metadata
defines `dependencies` as `null` or an empty array then it will not be
updated when the module is converted and it will subsequently fail to
validate cleanly.

This PR changes the metadata merging logic so that values defined as
`null` will be replaced with the default metadata value. Additionally, if the `requirements` array exists but is empty, it will also be updated with the default value.

### DIFF
--- a/lib/pdk/module/convert.rb
+++ b/lib/pdk/module/convert.rb
@@ -118,7 +118,7 @@ module PDK
             metadata = PDK::Module::Metadata.from_file(metadata_path)
             new_values = PDK::Module::Metadata::DEFAULTS.select do |key, _|
               !metadata.data.key?(key) || metadata.data[key].nil? ||
-                (metadata.data[key].respond_to?(:empty?) && metadata.data[key].empty?)
+                (key == 'requirements' && metadata.data[key].empty?)
             end
             metadata.update!(new_values)
           rescue ArgumentError
--- a/lib/pdk/module/convert.rb
+++ b/lib/pdk/module/convert.rb
@@ -111,7 +111,10 @@ module PDK
           if File.readable?(metadata_path)
             begin
               metadata = PDK::Module::Metadata.from_file(metadata_path)
-              new_values = PDK::Module::Metadata::DEFAULTS.reject { |key, _| metadata.data.key?(key) }
+              new_values = PDK::Module::Metadata::DEFAULTS.select do |key, _|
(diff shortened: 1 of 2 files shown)


lib/pdk/module/convert.rb
```diff
@@ -108,19 +108,22 @@ def template_url
...
+          unless File.readable?(metadata_path)
             raise PDK::CLI::ExitWithError, _('Unable to update module metadata; %{path} exists but it is not readable.') % {
               path: metadata_path,
             }
           end
+
+          begin
+            metadata = PDK::Module::Metadata.from_file(metadata_path)
+            new_values = PDK::Module::Metadata::DEFAULTS.select do |key, _|
+              !metadata.data.key?(key) || metadata.data[key].nil? ||
```
bmjen: Hmm. This PR makes me think about how we handle `operatingsystem_support`. Currently.. if the key doesn't exist, we put the defaults in, which may be inaccurate. But if the key does exist and is empty, we leave it alone. With this PR, in both of the cases, we will update the metadata with the default operating system support, which might be undesirable.

This is out of scope of this PR, but IMO, I think we need to remove operating system support from the defaults, and only check the key (and prompt user to select OS) if `--full-interview` is specified or if they are doing a `pdk build` and we detect the key is missing.

scotje: Yeah, I'm 👍 to replacing values that are `null` but I think replacing empty arrays with the defaults could be more problematic.

rodjek: Fair enough, but we'll have to add an exception for `requirements` as PDK will fail to validate as it expects to find a puppet version requirement.

scotje: Should we just have a pre-condition somewhere that bails out if `requirements` is an empty array?

rodjek: I don't think so. If the requirements array exists and is empty, it is a pretty harmless change to add a puppet requirement and saves the user from having to make a manual change (either removing the array which would then be re-added with the default value anyway during convert, or manually creating the puppet requirement).

scotje: Yeah, that sounds reasonable!

PR #586, opened 2018-10-31
(PDK-1194) Exclude plans/**/*.pp from PDK::Validate::PuppetSyntax

rodjek:
Plan syntax validation works slightly differently to regular manifest validation. It will probably be cleaner to implement this as a separate validator rather complicate the existing Puppet syntax validator with conditional behaviour.

### DIFF
--- a/lib/pdk/validate/base_validator.rb
+++ b/lib/pdk/validate/base_validator.rb
@@ -52,7 +52,8 @@ module PDK
                 end
               end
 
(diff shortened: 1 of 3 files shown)


MikaelSmith:
I took a stab at adding it to puppet-syntax at https://github.com/voxpupuli/puppet-syntax/pull/96, but the logic for choosing manifests or plans seems problematic.

PR #589, opened 2018-11-01
(PDK-1172) Call PDK::Util::Bundler.ensure_bundle! after module creation

rodjek:


lib/pdk/generate/module.rb
```diff
@@ -84,6 +84,8 @@ def self.invoke(opts = {})
 
         begin
           if FileUtils.mv(temp_target_dir, target_dir)
+            Dir.chdir(target_dir) { PDK::Util::Bundler.ensure_bundle! }
```
scotje: Do we want to do this regardless of whether or not it's a package install?

PR #596, opened 2018-11-20
(maint) Fix package tests

rodjek:
* PDK now lays down a `Gemfile.lock` when initialising a module, so we can't expect it to be absent.
 * Remove the `DEPENDENCIES` section of the `Gemfile.lock` before comparing. This section only lists the specified depedencies whereas the `GEMS` section is more important as it contains all the fully resolved dependencies. The `DEPENDENCIES` section can cause false failures in cases where different ways of specifying the dependency resolve to the same version e.g. `puppet (~> 6.0)` != `puppet (= 6.0.4)` even though they resolve to the same version.
 * Either the SUITE_VERSION env var changed for OSX or the path vanagon creates changed since the last release. In any case, the path to the `.pkg` in the `.dmg` has changed and we need to update our OSX package install logic.

### DIFF
--- a/package-testing/spec/package/add_gem_to_module_spec.rb
+++ b/package-testing/spec/package/add_gem_to_module_spec.rb
@@ -7,10 +7,6 @@ describe 'C100545 - Generate a module, add a gem to it, and validate it' do
     its(:exit_status) { is_expected.to eq(0) }
   end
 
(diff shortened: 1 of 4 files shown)


rodjek:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/PDK/job/platform_pdk_pdk-int-sys-testing_master/370/

PR #610, opened 2019-01-04
Update bundler dependency

akerl:
Bundler 2.0.0 is out, this kicks PDK up to use it. Based on https://github.com/bundler/bundler/blob/v2.0.0/CHANGELOG.md, the only impact is that it's a breaking change to the minimum Ruby / RubyGems versions.


rodjek:
@akerl Thanks for the contribution! Unfortunately, we decided not to merge it as we can't drop Ruby 2.1 support at this time. This issue has been resolved in https://github.com/puppetlabs/pdk/pull/611 though, so future versions of PDK will work with Bundler 2.x

PR #613, opened 2019-01-09
(maint) Fix package tests to remove hardcoding

bmjen:
The vendored Gemfile name had a hardcoded value with the "latest ruby"
version. This caused breakage in package tests each time packaged ruby
versions were updated. This fix updates the tests to detect the latest version
from the pdk private cache.


bmjen:
Ran the package tests changes against SHA "ea596e938f28de92f884e2abfd1356256641cd66" on redhat7 and windows2012.

PR #624, opened 2019-02-11
(PDK-1273) Whitelist Ruby symbols in YAML validator

rodjek:
Ruby symbols are the only additional classes allowed in Puppet's YAML loader, so this should be the only one we need to add.


david22swan:
Changing the PR name in so that it doesn't contain any offensive terminology that is then added to the changelog.
Apologies for any offense.

PR #633, opened 2019-03-02
(PDK-1298) acceptance:local test suite optimisation

rodjek:
Drops the execution time of the `acceptance:local` test suite to ~16 minutes on Travis and ~35 minutes on Appveyor.

### Optimisations
 * Combine duplicate commands that only specify different report formats into single commands with multiple reports specified (e.g. `pdk validate` & `pdk validate --format junit` => `pdk validate --format text:stdout --format junit:report.xml`)
 * Rewrite the majority of tests so that they tear down properly rather than the much slower module destroy and recreate process so that most of the test files only create a single module each.
 * Splits the `spec` and `acceptance:local` tasks in the Appveyor config into separate tasks.
 * Combine `bundle_spec.rb` and `bundle_management_spec.rb` tests into a single file, allowing them to share the same test module.

### Maintenance
 * Add a new `have_no_output` matcher to Englishify the many `match(%r{\A\Z})` expectations.
 * Add a new `contain_valid_junit_xml` matcher to DRY up the validation of the report output against the XSD.
 * General refactoring of the specs so that similar tests are consistently styled between files.


appveyor.yml
```diff
@@ -11,8 +11,12 @@ environment:
     #   BUNDLE_JOBS: 4
     - RUBY_VERSION: 24-x64
       USE_MSYS: true
-      SUITES: "spec acceptance:local"
+      SUITES: "acceptance:local"
       BUNDLE_JOBS: 4
+    - RUBY_VERSION: 24-x64
+      USE_CYGWIN: true
```
bmjen: To keep the same environment and coverage as before, shouldn't this line be `USE_MSYS: true` ?

rodjek: :+1: fixed

PR #638, opened 2019-03-08
(FM-7579, PDK-1236) bump the version of CRI used

tphoney:


scotje:
I believe this is blocked until we can drop Ruby 2.1.9 support from PDK?

tphoney:
thanks @scotje , i have a new commit had allows a range of CRI versions. hopefully removing the need to deprecate ruby 2.1.9. 
the point of this is to remove the conflict between the bolt and ruby gems

PR #639, opened 2019-03-15
(PDK-1193) Saves packaged template-url in metadata as a keyword

bmjen:
Currently, for packaged installations the default templates template-url
is saved in the metadata.json as the file path. This breaks when a
module that is developed with PDK on Windows is moved to a Linux installation
of PDK, since the Windows path doesn't exist. With this fix, the default
packaged template path will be determined by PDK using the keyword
'pdk-default' in that same metadata.json field and the approriate
platform specific template location will be used.

Requires #434

### DIFF
--- a/lib/pdk/util/template_uri.rb
+++ b/lib/pdk/util/template_uri.rb
@@ -5,6 +5,15 @@ module PDK
     class TemplateURI
       SCP_PATTERN = %r{\A(?!\w+://)(?:(?<user>.+?)@)?(?<host>[^:/]+):(?<path>.+)\z}
 
+      PACKAGED_TEMPLATE_KEYWORD = 'pdk-default'.freeze
+      DEPRECATED_TEMPLATE_URL = 'https://github.com/puppetlabs/pdk-module-template'.freeze
+
+      LEGACY_PACKAGED_TEMPLATE_PATHS = {
+        'windows' => 'file:///C:/Program Files/Puppet Labs/DevelopmentKit/share/cache/pdk-templates.git',
+        'macos'   => 'file:///opt/puppetlabs/pdk/share/cache/pdk-templates.git',
+        'linux'   => 'file:///opt/puppetlabs/pdk/share/cache/pdk-templates.git',
+      }.freeze
+
       # XXX Previously
       # - template_uri used to get the string form of the uri when generating the module and written to pdk answers and metadata
       # - template_path or deuri_path used for humans to see and commands to run
@@ -26,9 +35,16 @@ module PDK
       #
       def initialize(opts_or_uri)
         # If a uri string is passed, skip the valid uri finding code.
-        @uri = if opts_or_uri.is_a?(String) || opts_or_uri.is_a?(self.class)
+        @uri = if opts_or_uri.is_a?(self.class)
+                 opts_or_uri.uri
+               elsif opts_or_uri.is_a?(String)
                  begin
-                   Addressable::URI.parse(opts_or_uri)
+                   uri, ref = opts_or_uri.split('#', 2)
+                   if self.class.packaged_template?(uri)
+                     self.class.default_template_uri(ref).uri
+                   else
+                     Addressable::URI.parse(opts_or_uri)
+                   end
                  rescue Addressable::URI::InvalidURIError
                    raise PDK::CLI::FatalError, _('PDK::Util::TemplateURI attempted initialization with a non-uri string: {string}') % { string: opts_or_uri }
                  end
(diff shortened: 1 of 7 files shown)


lib/pdk/util.rb
```diff
@@ -8,6 +8,14 @@
 
 module PDK
   module Util
+    PACKAGED_TEMPLATE_KEYWORD = 'pdk-default'.freeze
```
DavidS: I'd use `'https://github.com/puppetlabs/pdk-module-template'` as `PACKAGED_TEMPLATE_KEYWORD`. This allows users to understand what's happening and where to go, while to the PDK it's just another string.

lib/pdk/util.rb
```diff
@@ -8,6 +8,14 @@
...
+
+    PACKAGED_TEMPLATE_PATHS = {
+      'windows' => 'file://C:/Program Files/Puppet Labs/DevelopmentKit/share/cache/pdk-templates.git',
```
DavidS: Reminder to mention "non-default install locations on Windows need manual intervention by the user to fix this" in the Release Notes.

lib/pdk/util.rb
bmjen: I think using `https://github.com/puppetlabs/pdk-module-template` (an old, deprecated repo for templates) would be confusing for users. I also didn't want a string that was clearly a URI or filepath, because we need to disambiguate the custom templates, gem installs, and package installs.

lib/pdk/util.rb
bmjen: We currently don't support changing the install location on Windows, so we know that anything in a version of PDK prior to the release of this fix will be at that static file path. Any modules converted or created after the introduction of this fix will appropriately show the static string.

lib/pdk/util.rb
DavidS: :man_facepalming: yeah that was a copy/paste error. I meant to put in the current templates URL

DavidS: now I'm curios: what difference does it make to the module if the PDK binary is from a gem or a package?

bmjen: More related to the work done in #434 , but one example is when performing an `update` on a package install that uses the packaged templates, we will need to overlook the `tag` in template-ref and proceed with an update to the latest tag. On a gem install, 1) the package templates path will not exist and 2) any `tag` ref will need to be honored as a static pin and only changes to .sync.yml will be processed until the user changes the `template-ref` to something else.

DavidS: I see now. Does that come from overloading `template-ref` in the `metadata.json` with "the ref used for the last update" and "the ref that should be used for updating"? Splitting that up and making it explicit (e.g. using `last-update-ref` and `next-update-ref`) might make those issues go away. #434 seems to be a good place to make that change.

DavidS: After chatting offline about this, I now get your point: When running from the native packages, and using the built-in pdk-default templates, this needs to specifically NOT be handled like a cache, but a separate source, because the pdk-default templates are pinned to a very specific version.

bmjen: In retrospect, I'm wondering if the `PACKAGED_TEMPLATE_KEYWORD` should be `pdk-packaged` instead. To avoid confusion about defaults.

/cc @DavidS @rodjek @scotje

DavidS: "packaged-templates" ? #bikeshed

lib/pdk/util/template_uri.rb
```diff
@@ -5,6 +5,15 @@ module Util
...
       SCP_PATTERN = %r{\A(?!\w+://)(?:(?<user>.+?)@)?(?<host>[^:/]+):(?<path>.+)\z}
 
+      PACKAGED_TEMPLATE_KEYWORD = 'pdk-default'.freeze
+      DEPRECATED_TEMPLATE_URL = 'https://github.com/puppetlabs/pdk-module-template'.freeze
+
+      PACKAGED_TEMPLATE_PATHS = {
+        'windows' => 'file:///C:/Program Files/Puppet Labs/DevelopmentKit/share/cache/pdk-templates.git',
+        'macos'   => 'file:///opt/puppetlabs/pdk/share/cache/pdk-templates.git',
+        'linux'   => 'file:///opt/puppetlabs/pdk/share/cache/pdk-templates.git',
+      }.freeze
```
scotje: Would it be possible to compute these from the relative path of the current file? I'm thinking about our hope to someday have the packages be relocatable and/or distributing a `.zip` package.

scotje: Oh, I think this is what Bryan and David were already talking about, this is only for recognizing the package template paths before this change goes out since afterwards this will get converted to a keyword in the metadata. Does that sound correct?

If so, maybe we just rename this const to like `LEGACY_PACKAGED_TEMPLATE_URLS` or something similar?

Issue #636, opened 2019-03-04
pdk update modifies the metadata.json such that it is missing a newline on the final line


ghoneycutt:
**Describe the bug**
When using `pdk update` the resulting `metadata.json` is missing a newline at the end of the file. 

Here's an example.

https://github.com/ghoneycutt/puppet-module-rpcbind/pull/40/files#diff-490694a9db8f7a371538da1abe484314R77

**To Reproduce**
Run `pdk update` on a module.

**Expected behavior**
There is a new line character.

**Additional context**
- Your PDK installation method: native packages
- Your PDK version: 1.9.0
- Your operating system / platform: macos mojave

scotje:
Tracked in JIRA as https://tickets.puppetlabs.com/browse/PDK-1308

scotje:
Resolved by #640

ghoneycutt:
Thanks @scotje and @rodjek

PR #640, opened 2019-03-19
(PDK-1308) Ensure PDK-written non-templated files have trailing newline

scotje:
There are some files that PDK writes that are not derived from templates
in the pdk-templates repo. The most notable example of this is the
metadata.json file which is constructed as a Ruby hash and then
serialized to JSON and written to a file. When PDK writes these files it
was not including a trailing newline and files having trailing newlines
is a best practice.

This change updates the PDK::Util::Filesystem.write_file method to
ensure that it always writes files with a trailing newline.

Some of the unit tests around writing templated files also needed to be
updated because of the way the tests were implemented did not simulate
template files that already contained trailing newlines.

### DIFF
--- a/lib/pdk/util/filesystem.rb
+++ b/lib/pdk/util/filesystem.rb
@@ -4,7 +4,13 @@ module PDK
       def write_file(path, content)
         raise ArgumentError unless path.is_a?(String) || path.respond_to?(:to_path)
 
-        File.open(path, 'wb') { |f| f.write(content.encode(universal_newline: true)) }
+        # Harmonize newlines across platforms.
+        content = content.encode(universal_newline: true)
+
+        # Make sure all written files have a trailing newline.
+        content += "\n" unless content[-1] == "\n"
+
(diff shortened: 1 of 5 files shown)


lib/pdk/util/filesystem.rb
```diff
@@ -4,6 +4,9 @@ module Filesystem
       def write_file(path, content)
         raise ArgumentError unless path.is_a?(String) || path.respond_to?(:to_path)
 
+        # Make sure all written files have a trailing newline.
+        content += "\n" unless content[-1] == "\n"
```
scotje: Perhaps we should do the `content.encode` before checking the last character? What do you folks think?

scotje:
Will resolve #636

lib/pdk/util/filesystem.rb
rodjek: Yeah, just for safety we should reencode first.

scotje:
@rodjek updated

PR #643, opened 2019-03-25
(PDK-1086) pdk build --force to warn if necessary and continue

### LINKED ISSUE #634: requirements for metadata.json does not match actual requirements
**Describe the bug**
The latest metadata.json documentation is in https://puppet.com/docs/puppet/6.3/modules_metadata.html it states which fields are required and which are optional.  For example operatingsystem_support is an optional field.

pdk build however refuse to build this perfectly valid module and will not accept a empty list either.

**To Reproduce**
Remove operatingsystem_support from metadata.json, try to build

**Expected behavior**
Modules written according to official specification can be built.

**Additional context**
PDK 1.9.0, EL7, rpm install

rodjek:
If there is missing fields in the `metadata.json` when running `pdk build --force`, display an warning and continue as is expected when running with `--force`.

```
$ pdk new module foo --skip-interview
pdk (INFO): Creating new module: foo
[✔] Resolving default Gemfile dependencies.
pdk (INFO): Module 'foo' generated at path '/home/tsharpe/code/puppetlabs/pdk/foo', from template 'https://github.com/puppetlabs/pdk-templates'.
pdk (INFO): In your module directory, add classes with the 'pdk new class' command.
$ cd foo
$ pdk build --force
pdk (WARN): This module is missing the following fields in the metadata.json: summary, source, project_page, issues_url. These missing fields may affect the visibility of the module on the Forge.
pdk (INFO): Building tsharpe-foo version 0.1.0
pdk (INFO): Build of tsharpe-foo has completed successfully. Built package can be found here: /home/tsharpe/code/puppetlabs/pdk/foo/pkg/tsharpe-foo-0.1.0.tar.gz
```

Fixes #634

### DIFF
--- a/lib/pdk/cli/build.rb
+++ b/lib/pdk/cli/build.rb
@@ -28,8 +28,12 @@ module PDK::CLI
       #
       unless module_metadata.forge_ready?
         if opts[:force]
-          PDK.logger.error _('This module is missing required fields in the metadata.json. Re-run the build command without --force to add this information.')
-          exit 1
+          PDK.logger.warn _(
+            'This module is missing the following fields in the metadata.json: %{fields}. ' \
+            'These missing fields may affect the visibility of the module on the Forge.',
+          ) % {
+            fields: module_metadata.missing_fields.join(', '),
+          }
         else
           module_metadata.interview_for_forge!
           module_metadata.write!('metadata.json')
--- a/lib/pdk/module/metadata.rb
+++ b/lib/pdk/module/metadata.rb
@@ -153,12 +153,12 @@ module PDK
         end
       end
 
-      private
-
       def missing_fields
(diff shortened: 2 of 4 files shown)


lib/pdk/cli/build.rb
```diff
@@ -28,8 +28,12 @@ module PDK::CLI
       #
       unless module_metadata.forge_ready?
         if opts[:force]
-          PDK.logger.error _('This module is missing required fields in the metadata.json. Re-run the build command without --force to add this information.')
-          exit 1
+          PDK.logger.warn _(
+            'This module is missing the following fields in the metadata.json: %{fields}. ' \
+            'These missing fields may affect the visibility of the module on the Forge.',
```
scotje: Are any of the fields things that would cause the Forge to reject it? If so maybe this should say "may affect your ability to publish this module on the Forge" instead?

PR #644, opened 2019-03-26
(PDK-1333) command_spec rake task

rodjek:
Initial stab at introspecting Cri to generate a JSON spec of the PDK command structure.

/cc @glennsarti @jpogran

### DIFF
--- a/rakelib/command_spec.rake
+++ b/rakelib/command_spec.rake
@@ -1,0 +1,25 @@
+task :command_spec do
+  require 'pdk/cli'
+  require 'json'
+
+  base_command = PDK::CLI.instance_variable_get('@base_cmd')
+  result = describe_command(base_command)
+  puts JSON.pretty_generate(result)
+end
+
+def describe_command(cri_command)
+  {
+    'name'        => cri_command.name,
+    'aliases'     => cri_command.aliases.to_a,
+    'description' => cri_command.description,
+    'hidden'      => cri_command.hidden,
+    'summary'     => cri_command.summary,
+    'usage'       => cri_command.summary,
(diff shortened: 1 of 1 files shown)


glennsarti:
Hrmmm small issue and I'm not sure how you'd solve it

e.g. `pdk new class` 

As per https://github.com/puppetlabs/pdk/blob/master/lib/pdk/cli/new/class.rb#L15 it requires a class name however this is not expressed in the spec output

```
        {
          "name": "class",
          "aliases": [

          ],
          "description": null,
          "hidden": null,
          "summary": "Create a new class named <class_name> using given options",
          "usage": "Create a new class named <class_name> using given options",
          "options": [

          ],
          "subcommands": [

          ]
        },
```

Apart from string interpolating for `<xxxxxx>` but that seems fragile.

rodjek:
cri 2.12.x introduced explicitly defined parameters, so we can make use of that to define the parameters and then pull them out when introspecting. As we still currently have to support Ruby 2.1.9 which is limited to cri 2.10.x, we'll need to guard these definitions. 

Something like this should work:

```ruby
param :class_name if respond_to?(:param)
```

We'd also need to change how we access the parameters as they'd change from an array to a hash (maybe `args = args.values if args.is_a?(Hash)`)

scotje:
I wonder if we could conditionally monkey patch basic `param` support into Cri < 2.12.x so that the actual PDK command definitions and argument handling code doesn't get littered with new conditionals?

glennsarti:
Maybe... This is just metadata at the end of the day,....

glennsarti:
I think if we cleaned up the commit message, this is ready for merge.  As it's purely for development of (PDK) and not user facing, this won't require the same rigour or have the same blast radius if it's wrong.

PR #652, opened 2019-04-04
(PDK-1338) Initial import of analytics code from Bolt

rodjek:


lib/pdk/telemetry.rb
```diff
@@ -0,0 +1,41 @@
...
+        Client::GoogleAnalytics.new(config['user-id'])
+      end
+    rescue StandardError => e
+      PDK.logger.debug "Failed to initialize telemetry client, telemetry will be disabled: #{e}"
+      Client::Noop.new
+    end
+
+    def self.load_config(filename)
+      if File.exist?(filename)
+        YAML.load_file(filename)
```
glennsarti: Should this be using safe load?

lib/pdk/telemetry.rb
```diff
@@ -0,0 +1,41 @@
+require 'securerandom'
+require 'pdk/telemetry/util'
+require 'pdk/telemetry/client/google_analytics'
+require 'pdk/telemetry/client/noop'
+
+module PDK
+  module Telemetry
+    def self.build_client
+      config_file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
```
glennsarti: nit: Could you add a note here that we'll be reviewing this filename/location.  It's not exactly intuitive to disable PDK telemetry in a directory called bolt.

lib/pdk/telemetry/client/google_analytics.rb
```diff
@@ -0,0 +1,132 @@
...
+    module Client
+      class GoogleAnalytics
+        PROTOCOL_VERSION  = 1
+        APPLICATION_NAME  = 'pdk'.freeze
+        TRACKING_ID       = 'UA-xxxx-1'.freeze
+        TRACKING_URL      = 'https://google-analytics.com/collect'.freeze
+        CUSTOM_DIMENSIONS = {
+          operating_system: :cd1,
+          inventory_nodes:  :cd2,
+          inventory_groups: :cd3,
```
glennsarti: Can probably remove this bolt specific custom dimensions

pdk.gemspec
```diff
@@ -36,6 +36,8 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'ffi', '~> 1.9.0'
   spec.add_runtime_dependency 'hitimes', '1.3.0'
   spec.add_runtime_dependency 'net-ssh', '~> 4.2.0'
+  spec.add_runtime_dependency 'httpclient', '~> 2.8.3'
+  spec.add_runtime_dependency 'concurrent-ruby', '~> 1.1.5'
```
glennsarti: IIRC this is compiled gem. No issues with this?

spec/unit/pdk/telemetry_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    ENV.delete('PDK_DISABLE_TELEMETRY')
+
+    # Ensure these tests will never read or write a local config
+    allow(described_class).to receive(:load_config).and_return(default_config)
+    allow(described_class).to receive(:write_config)
+  end
+
+  it 'creates a Noop client if telemetry is disabled' do
+    default_config.replace('disabled' => true)
+    expect(described_class).not_to receive(:write_config)
```
glennsarti: Not sure you need this expectation.  This seems like an implementation detail, not a behaviour.

pdk.gemspec
scotje: It will get compiled as part of the vanagon build process and the results cached into the packages, so it should be OK.

spec/unit/pdk/telemetry_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+describe PDK::Telemetry do
+  let(:default_config) { {} }
+
+  before(:each) do
+    # We use a hard override to disable analytics for tests, but that obviously
+    # interferes with these tests...
+    ENV.delete('PDK_DISABLE_TELEMETRY')
```
scotje: Does the ENV get reset somewhere between tests?

lib/pdk/telemetry.rb
scotje: Yeah, probably worth at least a `TODO` here and maybe a reference to the ticket where we plan to move the config.

lib/pdk/telemetry/util.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Telemetry
+    module Util
+      def self.compute_os
+        require 'concurrent/configuration'
+        require 'concurrent/future'
+
+        Concurrent::Future.execute(executor: :io) do
+          require 'facter'
+          os = Facter.value('os')
+          "#{os['name']} #{os.dig('release', 'major')}"
```
scotje: Did we figure out if `facter` is available at runtime to PDK itself?

rodjek: The simplest thing to do is adding facter 2.5.x as a dependency of PDK as it doesn't introduce any new dependencies.

spec/unit/pdk/telemetry_spec.rb
rodjek: Yep, in `spec/spec_helper.rb`

spec/unit/pdk/telemetry_spec.rb
rodjek: Yep, those specs were just copy-pastad in from Bolt and hadn't been checked through yet.

lib/pdk/telemetry/client/google_analytics.rb
rodjek: Yeah, I was going to leave that to be handled in a follow up PR, but for now I've removed the dimensions that we definitely won't be using.

lib/pdk/telemetry.rb
rodjek: Yep, I've added a TODO with the ticket ID for the task to rewrite the config loading.

lib/pdk/telemetry.rb
rodjek: Yep

lib/pdk.rb
```diff
@@ -7,4 +8,10 @@
 require 'pdk/validate'
 require 'pdk/version'
 
-module PDK; end
+module PDK
+  def self.analytics
+    @analytics = PDK::Analytics.build_client(
+      logger: PDK.logger,
+    )
+  end
```
scotje: Should this be an `||=` situation so we only instantiate one client?

Also I think for `PDK.logger` we add this method to the `PDK` module in `lib/pdk/logger.rb`, should probably be consistent one way or the other?

rodjek: Good call on `||=`, don't know how I missed that!

The reason I've put the accessor in `lib/pdk.rb` is that I'm trying to arrange things so that if/when it does get extracted the whole whole `lib/analytics` tree can be yoinked as is and need nothing more than a simple `sed` to rename it to something generic. To that end, I've refactored it to pass the logger in to the constructor rather than calling `PDK.logger` directly in the methods. When we refactor the configuration loading next, I'll do something similar so that the analytics constructor takes the loaded configuration data rather than handle it internally itself.

scotje: Cool, that makes sense!

PR #654, opened 2019-04-06
Allow a newer childprocess

ekohl:
73505faa181376727fbe42d1ac1ba35202bb98b3 pinned this to 0.7.1 but this means it's limited to the 0.7 branch. It should work with 0.9 as well.  It was bumped for a Windows 7 bugfix, but that was included in 0.6.3 so 0.7.0 should be safe as well.

### DIFF
--- a/pdk.gemspec
+++ b/pdk.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'bundler', '>= 1.15.0', '< 3.0.0'
   spec.add_runtime_dependency 'cri', ' >= 2.10.1', '< 2.16.0'
-  spec.add_runtime_dependency 'childprocess', '~> 0.7.1'
+  spec.add_runtime_dependency 'childprocess', '>= 0.7.1', '< 2.0.0'
   spec.add_runtime_dependency 'gettext-setup', '~> 0.24'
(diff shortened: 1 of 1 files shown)


scotje:
Hi @ekohl !

It seems like the 1.0.x releases are also fine, so maybe we could just do a compound requirement something like:

```
spec.add_runtime_dependency 'childprocess', '>= 0.7.1', '< 1.1.0'
```

what do you think?

ekohl:
That sounds like an even better solution though I wonder if we should assume they respect semver and allow `< 2`

scotje:
Yeah, I wasn't sure if we wanted to assume it would be strictly semver going forward. I'm OK with making that assumption until proven wrong though. :) Let me ping some other team members and see how they feel.

scotje:
@rodjek do you feel comfortable with `< 2.0.0` being the top of the range for `childprocess`?

scotje:
Hmm, the native extensions for 1.0.1 are failing to build on Windows with Ruby 2.1 for some reason...

scotje:
Going to close this for now, latest childprocess seems to have introduced a new native gem dependency on Windows that is failing in our Ruby 2.1/Windows CI environment and we don't currently have a strong motivation to debug. If there is an important bug fix, etc. in a newer childprocess release we can revisit this later.

PR #656, opened 2019-04-09
(PDK-1348) remove unused constants throwing warns

tphoney:


glennsarti:
Might be best to rescope the constants to `PDK::Util::Windows` and use that instead of using raw values.  Calling Win32 APIs is notoriously "odd" and the variable names help to keep it consistent across codebases (and sanity)

tphoney:
@glennsarti  is this ok ?

PR #660, opened 2019-05-03
(PDK-1300) Ensure `test unit --list` uses correct Puppet/Ruby env

scotje:
Previously `pdk test unit --list` would always run under PDK's default
Ruby runtime, which could cause a Bundler error if the module had never
had `ensure_bundle` called for that runtime.


scotje:
Spent a bunch of time this afternoon debugging the acceptance test failure, but now it's showing up on `master` as well so I think it's unrelated to this change. Still can't replicate the issue locally though. (Has something to do with `pdk validate --puppet-dev` exiting non-zero.)

PR #662, opened 2019-05-07
(maint) Add spawned process stdout & stderr to debug log

rodjek:


scotje:
There are some instances in the bundle helper at least where stdout/stderr is logged to debug when the process exits non-zero so we should probably go in and remove those.

scotje:
I guess that actually logs to `fatal` instead of `debug`, so that content will show up even if the user hasn't passed `--debug` which may be desirable for those failures. I think to avoid the duplication I can just check `PDK.logger.debug?` on those. I'll push up a commit.

PR #663, opened 2019-05-10
(PDK-1264) Display a nicer error when tarring long paths

rodjek:
Ideally we'd upgrade minitar to 0.8 which supports long paths, but the rest of the Puppet ecosystem is using minitar 0.6.1, which would result in building packages that r10k or PMT couldn't install. Also, upgrading minitar to 0.8 causes dependency hell with all the gems that have a `~> 0.6.1` minitar dependency.

Instead, we now display an error explaining the problem and asking the user to rename the file or exclude the file from the build before trying again.

### DIFF
--- a/lib/pdk/module/build.rb
+++ b/lib/pdk/module/build.rb
@@ -120,8 +120,14 @@ module PDK
         elsif File.symlink?(path)
           warn_symlink(path)
         else
+          validate_ustar_path!(relative_path.to_path)
           FileUtils.cp(path, dest_path, preserve: true)
         end
+      rescue ArgumentError => e
+        raise PDK::CLI::ExitWithError, _(
+          '%{message} Please rename the file or exclude it from the package ' \
+          'by adding it to the .pdkignore file in your module.',
+        ) % { message: e.message }
       end
 
       # Check if the given path matches one of the patterns listed in the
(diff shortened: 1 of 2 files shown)


lib/pdk/module/build.rb
```diff
@@ -152,6 +158,58 @@ def warn_symlink(path)
...
       end
 
+      # Checks if the path length will fit into the POSIX.1-1998 (ustar) tar
+      # header format.
+      #
+      # POSIX.1-2001 (which allows paths of infinite length) was adopted by GNU
+      # tar in 2004 and is supported by minitar 0.7 and above. Unfortunately
+      # much of the Puppet ecosystem still uses minitar 0.6.1.
+      #
+      # POSIX.1-2001 tar format does not allow for paths greater than 256 bytes,
```
scotje: Should this comment refer to the older `POSIX.1-1998` format?

lib/pdk/module/build.rb
```diff
@@ -120,8 +120,14 @@ def stage_path(path)
         elsif File.symlink?(path)
           warn_symlink(path)
         else
+          validate_ustar_path!(relative_path.to_path)
           FileUtils.cp(path, dest_path, preserve: true)
         end
+      rescue ArgumentError => e
+        raise PDK::CLI::ExitWithError, _(
+          '%{message} Please rename the file or exclude it from the package ' \
+          'by adding it to the .pdkignore file in your module.',
```
scotje: I think the subtle distinction between the two failure cases is probably not worth surfacing to the user in this message and we can just say that the path is too long here. We can log the distinction at DEBUG perhaps?

rodjek: I was thinking that, but say there is a file in the root of the module called `this_is_a_file_name_that_goes_over_100_characters_oh_no.md`, it would fail but would clearly be < 256 bytes which would make a generic error confusing.

lib/pdk/module/build.rb
rodjek: Yes, yes it should

scotje:
Forced pushed an amended commit with the comment fixed up. Will merge after CI passes.

PR #668, opened 2019-05-28
(PDK-1342) Submit PDK analytics events

rodjek:
OK, I think this implements analytics submission for all the items that we're interested in according to our spec.

Common values that are submitted for every analytics event:
 * A random non-identifying user ID (shared with Bolt analytics if installed and enabled)
 * PDK version
 * PDK installation method (`package` or `gem`)
 * Operating system

We submit a "screen view" for every successful CLI command invocation that records:
 * The command name
 * The version of Ruby used to execute the PDK command
 * The anonymised command options and arguments
  * All arguments are redacted
  * All non-boolean option values are redacted (except `--puppet-version` and `--pe-version`)
 * The output formats for the command
 * `PDK_*` environment variables and their values (if set)

Invalid commands are submitted as a distinct analytics events with the arguments and option values redacted.

Whenever a template repo is used, we submit an event recording if the template is `default` or `custom` (n.b. we do not record the path to the template repo itself). Additionally, if the template repo being used is the default one, we submit events for each file rendered, recording if the file is `unmanaged`, `deleted`, `customized` or `default` (n.b. in the event of a `customized` file, we do not record what was changed, only that it was changed via `.sync.yml` overrides).

Given the concerns of GDPR compliance you note that we're very carefully and strictly testing the analytics calls to ensure that no unexpected data is accidentally passed in.

### DIFF
--- a/lib/pdk/cli.rb
+++ b/lib/pdk/cli.rb
@@ -13,8 +13,44 @@ require 'pdk/report'
 require 'pdk/util/version'
 require 'pdk/util/puppet_version'
 
+class Cri::Command::CriExitException
+  def initialize(is_error:)
+    @is_error = is_error
+    PDK.analytics.event('CLI', 'invalid command', label: PDK::CLI.anonymised_args.join(' ')) if error?
+  end
+end
+
 module PDK::CLI
+  # Attempt to anonymise the raw ARGV array if the command parsing failed.
+  #
+  # If an item does not start with '-' but is preceeded by an item that does
+  # start with '-', assume that these items are an option/value pair and redact
+  # the value. Any additional values that do not start with '-' that follow an
+  # option/value pair are assumed to be arguments (rather than subcommand
+  # names) and are also redacted.
+  #
+  # @example
+  #   # Where PDK::CLI.args => ['new', 'plan', '--some', 'value', 'plan_name']
+  #
+  #   PDK::CLI.anonymised_args
+  #     => ['new', 'plan', '--some', 'redacted', 'redacted']
+  #
+  # @return Array[String] the command arguments with any identifying values
+  #   redacted.
+  def self.anonymised_args
+    in_args = false
+    @args.map do |arg|
+      if arg.start_with?('-')
+        in_args = true
+        arg
+      else
+        in_args ? 'redacted' : arg
+      end
+    end
+  end
+
   def self.run(args)
+    @args = args
     PDK::Config.analytics_config_interview! unless PDK::Config.analytics_config_exist?
     @base_cmd.run(args)
   rescue PDK::CLI::ExitWithError => e
--- a/lib/pdk/module/templatedir.rb
+++ b/lib/pdk/module/templatedir.rb
@@ -75,6 +75,9 @@ module PDK
 
         @module_metadata = module_metadata
 
+        template_type = uri.default? ? 'default' : 'custom'
+        PDK.analytics.event('TemplateDir', 'initialize', label: template_type)
+
         yield self
       ensure
         # If we cloned a git repo to get the template, remove the clone once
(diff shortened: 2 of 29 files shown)


spec/unit/pdk/cli/build_spec.rb
```diff
@@ -67,13 +73,27 @@
           expect(mock_metadata_obj).to receive(:interview_for_forge!)
           expect(mock_metadata_obj).to receive(:write!).with('metadata.json')
         end
+
+        it 'submits the command to analytics' do
+          allow(mock_metadata_obj).to receive(:interview_for_forge!)
+          allow(mock_metadata_obj).to receive(:write!)
+
+          expect(analytics).to receive(:screen_view).with('build', cli_options: 'target-dir=redacted', output_format: 'default', ruby_version: RUBY_VERSION)
```
scotje: I haven't really worked with rspec expectations and keyword arguments before, is the default behavior that additional keywords arguments are accepted or if we add another keyword arg later will that cause this to fail?

spec/unit/pdk/cli/convert_spec.rb
```diff
@@ -11,50 +11,111 @@
...
   context 'when run from inside a module' do
     before(:each) do
       allow(PDK::Util).to receive(:module_root).and_return('/path/to/test/module')
+      allow(PDK::Module::Convert).to receive(:invoke)
     end
 
     context 'and provided no flags' do
+      after(:each) do
+        PDK::CLI.run(['convert'])
+      end
```
scotje: Would these be better as `subject` blocks?

spec/unit/pdk/cli/build_spec.rb
scotje: Also maybe adopt the multi-line syntax you use in `convert_spec.rb` for these as well.

rodjek: Adding additional arguments later will cause this to fail (we'd have to use something like `hash_including` if we wanted it to be permissive). This does increase the maintenance burden a bit but increases the confidence of not letting any private data slip in to analytics.

spec/unit/pdk/cli/convert_spec.rb
rodjek: Not really, as the CLI run isn't the subject of the tests, it's just something that we have to do to trigger the logic being tested.

lib/pdk/cli.rb
```diff
@@ -13,8 +13,28 @@
...
 
+class Cri::Command::CriExitException
+  def initialize(is_error:)
+    @is_error = is_error
+    PDK.analytics.event('CLI', 'invalid command', label: PDK::CLI.anonymised_args.join(' ')) if error?
+  end
+end
+
 module PDK::CLI
+  def self.anonymised_args
```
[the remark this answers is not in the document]
rodjek: Done

lib/pdk/module/templatedir.rb
```diff
@@ -256,14 +259,28 @@ def config_for(dest_path, sync_config_path = nil)
...
           @config = conf_defaults
-          @config.deep_merge!(sync_config, knockout_prefix: '---') unless sync_config.nil?
+          @config.deep_merge!(@sync_config, knockout_prefix: '---') unless @sync_config.nil?
         end
         file_config = @config.fetch(:global, {})
         file_config['module_metadata'] = @module_metadata
         file_config.merge!(@config.fetch(dest_path, {})) unless dest_path.nil?
-        file_config.merge!(@config)
+        file_config.merge!(@config).tap do |c|
+          unless uri.default?
```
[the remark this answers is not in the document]
rodjek: Yep

spec/unit/pdk/cli/convert_spec.rb
scotje: I don't feel super strongly about the `subject` thing but I do think this happening in the `after` hook obfuscates things a bit.

PR #670, opened 2019-05-30
(PDK-1345) Disable analytics during package tests

rodjek:


package-testing/spec/spec_helper_package.rb
```diff
@@ -6,6 +6,8 @@
   let(:path) { windows_node? ? nil : "#{install_dir}/bin:$PATH" }
 end
 
+set :env, PDK_DISABLE_ANALYTICS: 'true'
```
scotje: Took me a minute to figure out that `set` came from the `specinfra` gem.

rodjek: Yeah, the specinfra gem is pretty terribly documented

PR #671, opened 2019-05-30
(PDK-1337) Warn and unset any of the legacy *_GEM_VERSION env vars

rodjek:


lib/pdk/logger.rb
```diff
@@ -29,6 +30,13 @@ def initialize
       self.level = ::Logger::INFO
     end
 
+    def warn_once(*args)
+      hash = args.inspect.hash
+      return if @sent_messages.key?(hash)
+      @sent_messages[hash] = true
```
scotje: Should we keep track of these per log level? Or does that already get captured by `args.inspect.hash`?

rodjek: If down the line we implement other `*_once` methods then yeah we'll need to track them by severity

rodjek: Pushed a quick update to track by severity

PR #674, opened 2019-06-01
Be less strict when using #hash_including with defaults

ddfreyne:
This PR introduces #hash_with_defaults_including, which is similar to RSpec’s built-in #hash_including, but does not require the keys of `expected` to be present in `actual` -- what matters is that for all keys in `expected`, the value in `actual` and `expected` is the same.

```ruby
hash = Hash.new { |hash, key| 9000 }
moo(hash)
```

This passes:

```
expect(something)
  .to receive(:moo)
  .with(hash_with_defaults_including(stuff: 9000))
```

This does not pass:

```ruby
expect(something)
  .to receive(:moo)
  .with(hash_including(stuff: 9000))
```

This is in preparation of landing ddfreyne/cri#99, which makes use of these `Hash` defaults to specify default values for options.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -95,3 +95,35 @@ RSpec.shared_context :validators do
     ]
   end
 end
+
+# Similar to RSpec’s built-in #hash_including, but does not require the keys of
+# `expected` to be present in `actual` -- what matters is that for all keys in
+# `expected`, the value in `actual` and `expected` is the same.
+#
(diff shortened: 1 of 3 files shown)


scotje:
Need to make sure this is updated everywhere and we should remove the Cri pin as part of the same PR.

scotje:
See also https://tickets.puppetlabs.com/browse/PDK-1384

scotje:
Superseded by #731

PR #675, opened 2019-06-04
(maint) Pin cri to <= 2.15.6

rodjek:
cri 2.15.7 changed the behaviour of the options hash so that unspecified
flags are defaulted to an explicit `false` rather than being unset. From
the look of https://github.com/ddfreyne/cri/issues/94 though this
behaviour may be changing in the next release so lets pin cri to
a maximum of 2.15.6 for now until the behaviour stabilises and then we
can make the necessary changes.

/cc #674

### DIFF
--- a/pdk.gemspec
+++ b/pdk.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'bundler', '>= 1.15.0', '< 3.0.0'
   spec.add_runtime_dependency 'childprocess', '~> 0.7.1'
-  spec.add_runtime_dependency 'cri', ' >= 2.10.1', '< 2.16.0'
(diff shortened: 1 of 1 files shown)


ddfreyne:
This PR makes sense, and I apologise for the trouble caused. I am working towards explicitly specifying Cri’s expected behavior, to minimize future disruption.

For reference, here are the release notes for Cri 2.15.8 (unreleased): https://github.com/ddfreyne/cri/pull/100.

Upgrading to Cri 2.15.8 will break the `pdk` tests (though not the `pdk` behavior), and so I have opened #674 (as you already mentioned) which should fix that.

scotje:
@ddfreyne Thank you for your quick response to this issue, we really appreciate your diligence as the maintainer of Cri!

PR #677, opened 2019-06-07
(PDK-1354) Default template ref for custom templates should always be master

rodjek:


lib/pdk/util/template_uri.rb
```diff
@@ -207,14 +207,21 @@ def self.templates(opts)
...
+
         if PDK::Util.development_mode?
           'master'
         else
-          PDK::TEMPLATE_REF
+          uri.default? ? PDK::TEMPLATE_REF : 'master'
         end
       end
 
+      def default_template_ref
```
glennsarti: Hrmmm well..this is a little confusing.  class vs instance name with different method signatures :-(

rodjek: Yeah, we don't really need the instance method, so i've removed it

lib/pdk/util/template_uri.rb
```diff
@@ -207,11 +207,14 @@ def self.templates(opts)
       end
 
       # @returns String
-      def self.default_template_ref
+      def self.default_template_ref(uri = nil)
```
glennsarti: This seems a little complex. Maybe this is better (not sure)?

``` ruby
def self.default_template_ref(uri = nil)
  return 'master' if PDK::Util.development_mode?
  return PDK::TEMPLATE_REF if uri.nil?  # The default url always uses PDK::TEMPLATE_REF as the ref
  uri = new(uri) unless uri.is_a?(self)
  uri.default? ? PDK::TEMPLATE_REF : 'master'
end

File: spec/unit/pdk/util/template_uri_spec.rb [hunk 3]
```diff
@@ -161,7 +161,7 @@
 
           it 'returns the specified template' do
             allow(Gem).to receive(:win_platform?).and_return(true)
-            expect(template_uri.to_s).to eq("C:\\cli-templates##{described_class.default_template_ref}")
+            expect(template_uri.to_s).to eq('C:\cli-templates#master')
```
glennsarti: nit: Although it's not required, I've found just always using double slash makes it far less confusing e.g. `\'` and `\\` is still are valid escape sequences in single quoted strings.

spec/unit/pdk/util/template_uri_spec.rb
```diff
@@ -302,37 +302,45 @@
   end
 
   describe '.default_template_ref' do
-    subject { described_class.default_template_ref }
+    subject { described_class.default_template_ref(uri) }
```
glennsarti: I don't see any tests for `described_class.default_template_ref()` anywhere.  Should these be added?

lib/pdk/util/template_uri.rb
scotje: Yeah I think @glennsarti 's guard clause version is clearer.

spec/unit/pdk/util/template_uri_spec.rb
rodjek: I can add an explicit `nil` test if you want, but it's just going to be duplicating the tests here https://github.com/puppetlabs/pdk/blob/e8797f535cb55b2fc29b195f80e54db7f5bcf8eb/spec/unit/pdk/util/template_uri_spec.rb#L331-L349

glennsarti: The only thing I would be looking for is:
`expect(described_class.default_template_ref()).to eq(PDK::TEMPLATE_REF)`

PR #678, opened 2019-06-07
(PDK-1365) Use dynamic ruby detection for default ruby instance

glennsarti:
Previously the default version of ruby selected in packaged PDK instances was
hardcoded to 2.4.5 which caused issues when updating ruby runtimes and did not
reflect that the default should really by ruby 2.5 and Puppet 6.  This commit
modifies the default ruby detection to instead use the ruby environment for the
latest Puppet Gem in the PDK package install.  Non-pkged installs are not
affected.

This commit also memoizes the default value to slightly speed up default ruby
selection.

### DIFF
--- a/lib/pdk/util/ruby_version.rb
+++ b/lib/pdk/util/ruby_version.rb
@@ -48,12 +48,19 @@ module PDK
         end
 
         def default_ruby_version
-          # For now, the packaged versions will be using default of 2.4.5
-          # FIXME: make this dynamic
-          return '2.4.5' if PDK::Util.package_install?
-
-          # TODO: may not be a safe assumption that highest available version should be default
-          latest_ruby_version
+          @default_ruby_version ||= if PDK::Util.package_install?
+                                      # Default to the ruby that supports the latest puppet gem. If you wish to default to a
+                                      # specific Puppet Gem version use the following example;
+                                      #
+                                      # PDK::Util::PuppetVersion.find_gem_for('5.5.10')[:ruby_version]
+                                      #
+                                      PDK::Util::PuppetVersion.latest_available[:ruby_version]
+                                    else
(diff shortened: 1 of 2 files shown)


lib/pdk/util/ruby_version.rb
```diff
@@ -48,12 +48,19 @@ def scan_for_packaged_rubies
...
+          @default_ruby_version ||= if PDK::Util.package_install?
+                                      # Default to the ruby that supports the latest puppet gem. If you wish to default to a
+                                      # specific Puppet Gem version use the following example;
+                                      #
+                                      # PDK::Util::PuppetVersion.find_gem_for('5.5.10')[:ruby_version]
+                                      #
+                                      PDK::Util::PuppetVersion.latest_available[:ruby_version]
+                                    else
+                                      # TODO: may not be a safe assumption that highest available version should be default
+                                      # WARNING Do NOT use PDK::Util::PuppetVersion.*** methods as it can recurse into this
```
rodjek: This conflicts with the code in the true side of this if statement where `PDK::Util::PuppetVersion.latest_available` is used.

glennsarti: Which bit conflicts?  The warning?

glennsarti: To clarify, when not in a packaged install there exists this callpath which can cause the recursion loop

PDK::Util::PuppetVersion.find_gem ->
PDK::Util::PuppetVersion.find_in_rubygems ->
PDK::Util::RubyVersion.default_ruby_version

Packaged installs use a deterministic code path using the on disk files so the recursion doesn't happen.

glennsarti: Because this behaviour may change in the future I've left the comment as a warning for future-us

PR #681, opened 2019-06-13
(PDK-1362) Warn user if updating module with older PDK version

rodjek:


lib/pdk/cli/update.rb
```diff
@@ -26,6 +26,18 @@ module PDK::CLI
         raise PDK::CLI::ExitWithError, _('You can not specify --noop and --force when updating a module')
       end
 
+      if Gem::Version.new(PDK::VERSION) < Gem::Version.new(PDK::Util.module_pdk_version)
+        PDK.logger.warn _(
+          'This module is compatible with PDK %{module_pdk_version} which ' \
+          'is newer than your PDK version (%{user_pdk_version}).',
```
scotje: I would suggest we make this a bit stronger and more declarative about the fact that the module has already been updated rather than just being "compatible" with the newer version.

Something like:

> This module has previously been updated to PDK {new_version} which is newer than your current PDK version ({current_version}), proceed with caution!

We should get @jbondpdx's feedback when she is back from PTO.

lib/pdk/cli/update.rb
```diff
@@ -26,6 +26,18 @@ module PDK::CLI
...
+        ) % {
+          module_pdk_version: PDK::Util.module_pdk_version,
+          user_pdk_version:   PDK::VERSION,
+        }
+
+        raise PDK::CLI::ExitWithError, _('Please update your PDK installation and try again.') unless opts[:force]
```
scotje:
```suggestion
        raise PDK::CLI::ExitWithError, _('Please update your PDK installation and try again. You may also use the --force flag to override this.') unless opts[:force]
```

lib/pdk/cli/update.rb
rodjek: :+1:

lib/pdk/cli/update.rb
rodjek: :+1:

PR #682, opened 2019-06-13
(PDK-1366) Update default operatingsystem versions

rodjek:


SimonHoenscheid:
The original PR was originally created to add Arch Linux to the List. That should be covered. 
cc @bastelfeak

lib/pdk/module/metadata.rb
```diff
@@ -12,50 +12,50 @@ class Metadata
         'RedHat based Linux' => [
           {
             'operatingsystem'        => 'CentOS',
-            'operatingsystemrelease' => ['7'],
+            'operatingsystemrelease' => ['8'],
```
SimonHoenscheid: This is a nonexistent future release.

lib/pdk/module/metadata.rb
```diff
@@ -12,50 +12,50 @@ class Metadata
...
           },
           {
             'operatingsystem'        => 'RedHat',
-            'operatingsystemrelease' => ['7'],
+            'operatingsystemrelease' => ['8'],
           },
           {
             'operatingsystem'        => 'Scientific',
-            'operatingsystemrelease' => ['7'],
+            'operatingsystemrelease' => ['8'],
```
SimonHoenscheid: The will be no Scientific Linux 8. See https://listserv.fnal.gov/scripts/wa.exe?A2=SCIENTIFIC-LINUX-ANNOUNCE;11d6001.1904

lib/pdk/module/metadata.rb
```diff
@@ -12,50 +12,50 @@ class Metadata
         'RedHat based Linux' => [
           {
             'operatingsystem'        => 'CentOS',
-            'operatingsystemrelease' => ['7'],
+            'operatingsystemrelease' => ['8'],
           },
           {
             'operatingsystem'        => 'OracleLinux',
-            'operatingsystemrelease' => ['7'],
+            'operatingsystemrelease' => ['8'],
```
SimonHoenscheid: Still a beta release

SimonHoenscheid:
Is there a reason to just reference latest stable release? All currently supported versions should be referenced.

rodjek:
@SimonHoenscheid Thanks for the information around the other EL8 based distros!

> The original PR was created to add Arch Linux to the List. That should be covered.

When creating and updating the list of default operating systems we have to balance all the available options with the need to not overload the users with a bunch of options. For that reason we only offer the most common supported operating systems & distributions in the defaults and unfortunately (and I'm speaking as a happy Arch user here) Arch is just not popular enough in the module development community to justify adding it at this time.

> Is there a reason to just reference latest stable release?

We only reference the latest stable release rather than all supported releases because when someone is developing a new module, 95% of the time they are writing it for use on the latest release. It may work on older releases of an OS, but that decision should be made by module author and explicitly specified. This helps keep the Forge clean by making it less likely that modules are published with incorrect metadata.

I hope that all makes sense :)

lib/pdk/module/metadata.rb
```diff
@@ -30,32 +30,32 @@ class Metadata
...
+          'operatingsystemrelease' => ['15'],
         },
         'Solaris' => {
           'operatingsystem'        => 'Solaris',
           'operatingsystemrelease' => ['11'],
         },
         'Windows' => {
           'operatingsystem'        => 'windows',
-          'operatingsystemrelease' => ['2008 R2', '2012 R2', '10'],
+          'operatingsystemrelease' => %w[2019 10],
```
glennsarti: This one is a bit tricky. The most common Server OS right now is either 2016 or 2012R2, but 2019 is the latest.

PR #689, opened 2019-06-28
(PDK-1414) Detect common CI environments and set non-interactive

glennsarti:
Previously PDK detected whether a session was interactive by using the isatty
method on STDERR. However it is still possible to create TTY based STDERR in
CI environments such as Travis.  This commit adds additional Environment
Variable checks for common CI systems to ensure that PDK is non-interactive in
CI.

### DIFF
--- a/lib/pdk/cli/util.rb
+++ b/lib/pdk/cli/util.rb
@@ -48,9 +48,30 @@ module PDK
       end
       module_function :prompt_for_yes
 
+      # Uses environment variables to detect if the current process is running in common
+      # Continuous Integration (CI) environments
+      # @return [Boolean] Whether the PDK is in a CI based environment
+      def ci_environment?
+        [
+          'APPVEYOR_BUILD_FOLDER', # AppVeyor CI
+          'GITLAB_CI',             # GitLab CI
+          'JENKINS_URL',           # Jenkins
+          'BUILD_DEFINITIONNAME',  # Azure Pipelines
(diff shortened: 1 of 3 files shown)


scotje:
Created PDK-1414 to track and amended the commit message to include ticket number.

lib/pdk/cli/util.rb
```diff
@@ -48,9 +48,30 @@ def prompt_for_yes(question_text, opts = {})
...
+        [
+          'APPVEYOR_BUILD_FOLDER', # AppVeyor CI
+          'GITLAB_CI',             # GitLab CI
+          'JENKINS_URL',           # Jenkins
+          'BUILD_DEFINITIONNAME',  # Azure Pipelines
+          'TEAMCITY_VERSION',      # Team City
+          'BAMBOO_BUILDKEY',       # Bamboo
+          'GOCD_SERVER_URL',       # Go CD
+          'TRAVIS',                # Travis CI
+          'GITHUB_WORKFLOW'        # GitHub Actions
```
smortex: Travis-CI also set some ["generic" environment variables](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables):

* `CI=true`
* `CONTINUOUS_INTEGRATION=true`

Does it make sense to add these to the list hopping that other CI systems use such variables?

scotje: Yeah wouldn't hurt, I'll amend those in.

PR #695, opened 2019-07-02
(maint) Update beaker in package tests

rodjek:
Updates beaker and friends to 4.10.0 so that we can run tests against the latest platforms like RHEL8. The gem versions are all pinned because I'm paranoid about the pipeline breaking due to unrelated changes to any of these packages.

I'll link the result of the Jenkins run here once it has finished.

### DIFF
--- a/package-testing/Gemfile
+++ b/package-testing/Gemfile
@@ -13,11 +13,11 @@ end
 gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '= 4.10.0')
 gem 'beaker-abs', *location_for(ENV['BEAKER_ABS_VERSION'] || '= 0.5.0')
 gem 'beaker-hostgenerator', *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || '= 1.1.31')
-gem 'beaker-rspec', '= 6.2.4'
-gem 'rake', '= 12.3.2'
-gem 'nokogiri', '= 1.10.3'
-gem 'beaker-vmpooler', '= 1.3.3'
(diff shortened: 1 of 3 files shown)


rodjek:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/PDK/view/adhoc/job/platform_pdk_adhoc_pdk-int-sys-testing_adhoc/29/

package-testing/Gemfile
```diff
@@ -10,11 +10,14 @@ def location_for(place, fake_version = nil)
...
 
-gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '~> 3.34')
-gem 'beaker-abs', *location_for(ENV['BEAKER_ABS_VERSION'] || '~> 0.2')
-gem 'beaker-hostgenerator', *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || '~> 1.1.8')
-gem 'beaker-rspec'
-gem 'rake', '~> 10.1'
+gem 'beaker', *location_for(ENV['BEAKER_VERSION'] || '= 4.10.0')
+gem 'beaker-abs', *location_for(ENV['BEAKER_ABS_VERSION'] || '= 0.5.0')
+gem 'beaker-hostgenerator', *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || '= 1.1.31')
+gem 'beaker-puppet', '= 1.18.5'
```
glennsarti: Can't say I'm a fan, but I completely understand why.  I've been bitten way too many times as well.

PR #699, opened 2019-07-03
(PDK-1333) Fix command_spec rake task for newer CRI versions

glennsarti:
Previously the command_spec rake task only worked for older CRI version
(<= 2.11.0) because the cri_options used to be passed as a hash, but is now it's
own object type. This commit updates the conversion process to optionally
convert the object to hash on newer CRI versions.


glennsarti:
Thought versions would be more "correct".  But comparing on object type is ok too.

PR #700, opened 2019-07-03
(maint) Fix typo in gitignore

glennsarti:
The /ext/packaging directory was not being excluded correctly as there appeared
to be a typo in the text.  This commit removes the `+ ` text.

[ci skip]


glennsarti:
Merging without the need for CI.  It's a gitignore change.

PR #703, opened 2019-07-07
Extend version range for net-ssh

leoarnold:
There is a dependency conflict between `pdk` and `beaker`
since beaker `>= 4.7.0` requires `net-ssh ~> 5.0` and
`pdk` is still tied to `~> 4.2.0`.

All tests still pass when extending the version range:

https://travis-ci.org/leoarnold/pdk/jobs/555337752

### DIFF
--- a/pdk.gemspec
+++ b/pdk.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'json-schema', '2.8.0'
   spec.add_runtime_dependency 'json_pure', '~> 2.1.0'
   spec.add_runtime_dependency 'minitar', '~> 0.6.1'
-  spec.add_runtime_dependency 'net-ssh', '~> 4.2.0'
+  spec.add_runtime_dependency 'net-ssh', '>= 4.2.0', '< 6.0.0'
   spec.add_runtime_dependency 'pathspec', '~> 0.2.1'
   spec.add_runtime_dependency 'tty-prompt', '0.13.1'
   spec.add_runtime_dependency 'tty-spinner', '0.5.0'


scotje:
Hi @leoarnold, thank you for this contribution!

Unfortunately, I don't think we can merge this as-is due to wanting to retain Ruby 2.1.9 support. (I don't think Rubygems is smart enough to fall back to the compatible version automatically when someone does `gem install pdk`).

However, the good news is that this dependency isn't actually directly used by PDK, I believe it is here to help force things into using a version of beaker that is compatible with Ruby 2.1.9 but we are going to look into other ways to handle this so that everyone can use the latest version of beaker that is supported by their Ruby version.

leoarnold:
Gem is unskilled as ever:

```
~
$ rvm use 2.1.9
Using /home/larnold/.rvm/gems/ruby-2.1.9
~
$ gem list

*** LOCAL GEMS ***

bigdecimal (default: 1.2.4)
bundler (default: 1.17.3)
bundler-unload (1.0.2)
coderay (1.1.2)
executable-hooks (1.6.0)
gem-wrappers (1.4.0)
io-console (default: 0.4.3)
json (default: 1.8.1)
method_source (0.9.2)
minitar (0.8)
minitest (5.11.3, default: 4.7.5)
pry (0.12.2)
psych (default: 2.0.5)
rake (default: 10.1.0)
rdoc (default: 4.1.0)
rubygems-bundler (1.4.5)
rvm (1.11.3.9)
test-unit (default: 2.1.9.0)
~
$ gem install beaker
ERROR:  Error installing beaker:
        There are no versions of byebug (~> 11.0) compatible with your Ruby & RubyGems. Maybe try installing an older version of the gem you're looking for?
        byebug requires Ruby version >= 2.3.0. The current ruby version is 2.1.9.490.
~
$ gem install net-ssh:'~> 4.2.0'
Fetching net-ssh-4.2.0.gem
Successfully installed net-ssh-4.2.0
Parsing documentation for net-ssh-4.2.0
Installing ri documentation for net-ssh-4.2.0
Done installing documentation for net-ssh after 1 seconds
1 gem installed
~
$ gem install beaker
ERROR:  Error installing beaker:
        There are no versions of byebug (~> 11.0) compatible with your Ruby & RubyGems. Maybe try installing an older version of the gem you're looking for?
        byebug requires Ruby version >= 2.3.0. The current ruby version is 2.1.9.490.
```

Only Bundler is able to handle this correctly. Then again, with the change suggested here, and given a Gemfile with both `pdk` and `beaker` in it, Bundler will be smart enough to pick the latest version of `beaker` suitable for the given Ruby version, then infer which version of `net-ssh` is compatible with the both of them and pick that one. So, as far as I see, there should be no problem.

scotje:
The plan here is to add `net-ssh` as a dependency to the `puppet-module-system-posix-r2.1` metagem and then remove it entirely from the PDK gemspec.

scotje:
This work is now tracked as https://tickets.puppetlabs.com/browse/PDK-1465

PR #704, opened 2019-07-10
(MAINT) Add debug logging of yaml files being validated

npwalker:


glennsarti:
Is the report not enough to see what's going on?  Just curious as to why you need to the debugging.

npwalker:
@glennsarti when I run `pdk validate yaml` I just get this output.  

```
pdk validate yaml
pdk (INFO): Using Ruby 2.5.1
pdk (INFO): Using Puppet 6.6.0
[✔] Checking YAML syntax (["**/*.yaml", "*.yaml", "**/*.yml", "*.yml"]).
```

scotje:
Amended commit with an update to externalize the log string.

PR #705, opened 2019-07-11
(PDK-1434) Gracefully handle unparsable bolt analytics config

rodjek:
This issue was found because the classroom Windows instances created the
`~/puppetlabs/bolt/analytics.yaml` manually with UTF-16LE encoding which
Psych on Windows is unable to parse, causing
`PDK::Config.bolt_analytics_config` to raise `PDK::Config::LoadError`.

The classroom bootstrap scripts have been fixed to generate the file as
UTF-8 now, but we should still gracefully handle this case and have PDK
behave as if the file doesn't exist and use the default values.

### DIFF
--- a/lib/pdk/config.rb
+++ b/lib/pdk/config.rb
@@ -34,7 +34,14 @@ module PDK
     end
 
     def self.bolt_analytics_config
-      PDK::Config::YAML.new(file: File.expand_path('~/.puppetlabs/bolt/analytics.yaml'))
+      file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
+      PDK::Config::YAML.new(file: file)
+    rescue PDK::Config::LoadError => e
+      PDK.logger.debug _('Unable to load %{file}: %{message}') % {
+        file:    file,
+        message: e.message,
+      }
+      PDK::Config::YAML.new
     end
 
     def self.analytics_config_path
--- a/spec/unit/pdk/config_spec.rb
+++ b/spec/unit/pdk/config_spec.rb
@@ -8,6 +8,7 @@ describe PDK::Config do
   let(:user_config_content) { '{}' }
   let(:analytics_config_content) { nil }
   let(:bolt_analytics_content) { nil }
+  let(:bolt_analytics_path) { '~/.puppetlabs/bolt/analytics.yaml' }
 
   def mock_file(path, content)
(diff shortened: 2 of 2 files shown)


lib/pdk/config.rb
```diff
@@ -34,7 +34,14 @@ def user
     end
 
     def self.bolt_analytics_config
-      PDK::Config::YAML.new(file: File.expand_path('~/.puppetlabs/bolt/analytics.yaml'))
+      file = File.expand_path('~/.puppetlabs/bolt/analytics.yaml')
+      PDK::Config::YAML.new(file: file)
```
glennsarti: I'm not sure if this matters, but the return types here look different.  The happy path returns a `PDK::Config::YAML` object while the unhappy path returns a `PDK::Config::Namespace` object

glennsarti: Perhaps we need a helper there for `PDK::Config::YAML.empty` 

so in `pdk/config/yaml.rb`

``` ruby
class << self
  def empty
    PDK::Config::Namespace.new
  end
end
```

OR

just `PDK::Config::YAML.new` ?

spec/unit/pdk/config_spec.rb
```diff
@@ -82,6 +96,20 @@ def mock_file(path, content)
...
+
+        context 'and the bolt configuration is unparsable' do
+          before(:each) do
+            allow(PDK::Config::YAML).to receive(:new).and_call_original
+            allow(PDK::Config::YAML).to receive(:new)
+              .with(file: File.expand_path(bolt_analytics_path))
+              .and_raise(PDK::Config::LoadError)
+          end
+
+          it 'generates a new UUID' do
```
glennsarti: Hrmmm...this I'm not sure about.  If we know there's a bolt config but it's unparsable, should we disable our analytics or generate new ID?

lib/pdk/config.rb
rodjek: `PDK::Config::Namespace` is just `PDK::Config::YAML` without the yaml logic, but yeah I'll change it to `PDK::Config::YAML.new` for consistency.

spec/unit/pdk/config_spec.rb
rodjek: I think we should treat it as if the file doesn't exist, so the default values will be `disabled => true` and a new UUID.

glennsarti: I guess that's the default behaviour so yeah.

PR #710, opened 2019-07-19
support new module creation with removed files in configuration

bond-os:
When running `pdk new module` against template that has a file set with `delete: true` in `config_defaults.yml` module creation fails as destination status returned from template is ignored and pdk tries to read non-existent file.

### DIFF
--- a/lib/pdk/generate/module.rb
+++ b/lib/pdk/generate/module.rb
@@ -58,10 +58,12 @@ module PDK
 
         begin
           PDK::Module::TemplateDir.new(template_uri, metadata.data, true) do |templates|
-            templates.render do |file_path, file_content|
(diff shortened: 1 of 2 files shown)


seanmil:
Independently I had put up PR #723 before stumbling across this PR which I initially deemed to be more correct and complete. But then I realized that this PR does string comparisons to the status value, not symbol comparisons. The values yielded by `PDK::Module::TemplateDir.render()` here https://github.com/puppetlabs/pdk/blob/master/lib/pdk/module/templatedir.rb#L125-L130 are symbols. I have put up PR #725 which hopefully includes the best of this PR and my other PR.

scotje:
I agree that I think #725 is a more correct fix for this, but thank you for getting the ball rolling @bond-os !

PR #712, opened 2019-07-24
(PDK-641) Make `pdk bundle` fully interactive

scotje:
Previously, the subprocess invoked by `pdk bundle` had its output
streams buffered as a side effect of the subprocess infrastructure used
for other commands. Additionally, the input stream was not connected to
the subprocess at all.

This commit adds a new subclass of `PDK::CLI::Exec::Command` called
`PDK::CLI::Exec::InteractiveCommand` that bypasses all of the output
buffering, spinners, etc. that the parent class implements and instead
uses Ruby's built-in `Kernel.system` method to run the subprocess fully
interactively. The `pdk bundle` subcommand is also updated to use this
new class for invoking Bundler.

If this approach proves successful, it will likely be conditionally
extended to certain invocations of `pdk test unit` in a future PR.

### DIFF
--- a/lib/pdk/cli/exec/command.rb
+++ b/lib/pdk/cli/exec/command.rb
@@ -1,0 +1,233 @@
+require 'bundler'
+require 'childprocess'
+require 'English'
+require 'tempfile'
+require 'tty-spinner'
+require 'tty-which'
+
+require 'pdk/util'
+require 'pdk/util/git'
+require 'pdk/util/ruby_version'
+
+module PDK
+  module CLI
+    module Exec
+      class Command
+        attr_reader :argv
+        attr_reader :context
+        attr_accessor :timeout
+        attr_accessor :environment
+        attr_writer :exec_group
+
+        def initialize(*argv)
+          @argv = argv
+
+          @process = ChildProcess.build(*@argv)
+          @process.leader = true
+
+          @stdout = Tempfile.new('stdout').tap { |io| io.sync = true }
+          @stderr = Tempfile.new('stderr').tap { |io| io.sync = true }
+
+          @process.io.stdout = @stdout
+          @process.io.stderr = @stderr
+
+          # Default to running things in the system context.
+          @context = :system
+
+          # Extra environment vars to add to base set.
+          @environment = {}
+
+          # Register the ExecGroup when running in parallel
+          @exec_group = nil
+        end
+
+        def context=(new_context)
+          unless [:system, :module, :pwd].include?(new_context)
+            raise ArgumentError, _("Expected execution context to be :system or :module but got '%{context}'.") % { context: new_context }
+          end
+
+          @context = new_context
+        end
+
+        def register_spinner(spinner, opts = {})
+          return unless PDK::CLI::Util.interactive?
+          @success_message = opts.delete(:success)
+          @failure_message = opts.delete(:failure)
+
+          @spinner = spinner
+        end
+
+        def add_spinner(message, opts = {})
+          return unless PDK::CLI::Util.interactive?
+          @success_message = opts.delete(:success)
+          @failure_message = opts.delete(:failure)
+
+          @spinner = TTY::Spinner.new("[:spinner] #{message}", opts.merge(PDK::CLI::Util.spinner_opts_for_platform))
+        end
+
+        def update_environment(additional_env)
+          @environment.merge!(additional_env)
+        end
+
+        def execute!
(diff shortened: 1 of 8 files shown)


scotje:
Still need to add some unit tests for the new class but wanted to get this up for some conceptual review.

Tested the approach on macOS, Linux, and Windows and it seems to work. (Tested by adding a `pry` breakpoint into a spec file and then running `pdk bundle exec rake spec`.)

lib/pdk/cli/exec/command.rb
```diff
@@ -0,0 +1,233 @@
...
+        attr_accessor :environment
+        attr_writer :exec_group
+
+        def initialize(*argv)
+          @argv = argv
+
+          @process = ChildProcess.build(*@argv)
+          @process.leader = true
+
+          @stdout = Tempfile.new('stdout').tap { |io| io.sync = true }
```
glennsarti: Warning: Tempfile doesn't set binary mode, so CRLF will be used on Windows which may not be what you want.

lib/pdk/cli/exec/command.rb
```diff
@@ -0,0 +1,233 @@
...
+            # Make sure invocation of Ruby prefers our private installation.
+            package_binpath = PDK::Util.package_install? ? File.join(PDK::Util.pdk_package_basedir, 'bin') : nil
+
+            resolved_env['PATH'] = [
+              PDK::Util::RubyVersion.bin_path,
+              File.join(resolved_env['GEM_HOME'], 'bin'),
+              PDK::Util::RubyVersion.gem_paths_raw.map { |gem_path| File.join(gem_path, 'bin') },
+              package_binpath,
+              PDK::Util.package_install? ? PDK::Util::Git.git_paths : nil,
+              ENV['PATH'],
```
glennsarti: Warning: ENV has known issues with non-ASCII characters on Windows environment variables.  https://github.com/puppetlabs/puppet/blob/eb19bedb221959bd202e598401a3d29ea77950c9/spec/unit/util_spec.rb#L99-L218

Typically this comes from non-English Usernames e.g. `C:\Users\jmüller`

glennsarti:
Manually tested with cmd.exe, powershell.exe, conemu and new windows terminal.  Confirmed I can use pry interactively within a `bundle exec` but not within `pdk test unit

lib/pdk/cli/exec/command/interactive.rb
```diff
@@ -0,0 +1,101 @@
...
+            PDK.logger.debug(_('Command environment:'))
+            @resolved_env.each do |var, val|
+              PDK.logger.debug("  #{var}: #{val}")
+            end
+          end
+
+          start_time = Time.now
+
+          # Use the string form of command to ensure command is invoked via a shell
+          system(@resolved_env, command_string)
```
rodjek: Do we gain much from using `Kernel.system` here instead of using `Kernel.exec` to just replace the running process after setting up the environment?

scotje: Am I correct that `exec` never returns to the original process? Once we extend this beyond `pdk bundl` I think it fits better in the current model to have it return to the PDK process so we have a chance to do stuff like cleaning up fixtures, etc. before exiting.

scotje: Also preserves the option of running multiple interactive commands sequentially.

lib/pdk/cli/exec/command.rb
scotje: The stuff in `lib/pdk/cli/exec/command.rb` was extracted out into it's own file but pretty much unchanged from the current implementation. So this is a good thing to know but shouldn't be a regression from current behavior.

lib/pdk/cli/exec/command.rb
scotje: Similar to above, good thing to know but shouldn't be a regression since this is the existing implementation.

lib/pdk/cli/exec/command/interactive.rb
rodjek: Fair enough, if we're planning on using it for more than `pdk bundle` then `exec` is not suitable

scotje:
Updated with unit tests, I'm wondering how necessary acceptance/package tests are since it would mostly be testing that `Kernel#system` is behaving correctly?

scotje:
We decided to separately ticket automated acceptance tests for this, so I think this is ready to merge.

PR #713, opened 2019-07-25
(PDK-1309) Ensure file modes in built modules are sane

rodjek:
When building module tarballs for publishing, we need to sanitise the modes of the tarball entries. Directories should have at least 0755 and files should have at least 0644.

### DIFF
--- a/lib/pdk/module/build.rb
+++ b/lib/pdk/module/build.rb
@@ -227,11 +227,19 @@ module PDK
               entry_meta = {
                 name: entry,
               }
-              entry_meta[:mode] = if Minitar.dir?(entry)
-                                    File.stat(entry).mode | 0o755
-                                  else
-                                    File.stat(entry).mode | 0o644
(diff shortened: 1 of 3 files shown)


lib/pdk/module/build.rb
```diff
@@ -220,8 +220,30 @@ def build_package
...
+                name: entry,
+              }
+
+              orig_mode = File.stat(entry).mode
+              min_mode = Minitar.dir?(entry) ? 0o755 : 0o644
+
+              entry_meta[:mode] = orig_mode | min_mode
+
+              if entry_meta[:mode] != orig_mode
+                PDK.logger.debug(_('Updated permissions of packaged \'%{entry}\' to %{new_mode}') % {
```
scotje: I added this logging, I originally had it at `WARN` but the original report was someone who deliberately developed with a restrictive umask so I though a warning for every file would be annoying.

PR #714, opened 2019-07-26
(maint) Expect pdk test unit to run more than 1 test

rodjek:
Relax this expectation a bit so that changes in the default operatingsystem list doesn't cause false failures.


glennsarti:
Will merge once appv goes green

PR #715, opened 2019-07-29
(PDK-1107) Add pdk config get CLI command

glennsarti:
- [x] ~~Needs to format as json?~~ punting. too hard and not needed
- [x] Do I need acceptance tests? wouldn't have thought so
- [x] Blocked by https://github.com/puppetlabs/pdk-planning/issues/40
- [x] ~~Add validation and schema~~  Separate PR #728 

---

Previously there was no way to display what the current configuration the PDK
was using.  This commit adds a `pdk config` command, with a single `get`
sub-command.

This commit also adds the ability to filter the settings either by a specific
name or setting treename.

This commit also adds tests for the new resolve method in the Namespace object.

### DIFF
--- a/lib/pdk/cli/config.rb
+++ b/lib/pdk/cli/config.rb
@@ -1,0 +1,20 @@
+module PDK::CLI
+  @config_cmd = @base_cmd.define_command do
+    name 'config'
+    usage _('config [subcommand] [options]')
+    summary _('Configure the Puppet Development Kit.')
+    default_subcommand 'help'
+
+    run do |_opts, args, _cmd|
+      if args == ['help']
+        PDK::CLI.run(%w[config --help])
+        exit 0
+      end
+
+      PDK::CLI.run(%w[config help]) if args.empty?
+    end
+  end
+  @config_cmd.add_command Cri::Command.new_basic_help
+end
+
+require 'pdk/cli/config/get'
--- a/lib/pdk/cli/config/get.rb
+++ b/lib/pdk/cli/config/get.rb
@@ -1,0 +1,24 @@
+module PDK::CLI
+  @config_get_cmd = @config_cmd.define_command do
+    name 'get'
+    usage _('config get [name]')
+    summary _('Retrieve the configuration for <name>. If not specified, retrieve all configuration settings')
+
+    run do |_opts, args, _cmd|
+      item_name = args[0]
+      resolved_config = PDK.config.resolve(item_name)
+      # If the user wanted to know a setting but it doesn't exist, raise an error
+      if resolved_config.empty? && !item_name.nil?
+        PDK.logger.error(_("Configuration item '%{name}' does not exist") % { name: item_name })
+        exit 1
+      end
+      # If the user requested a setting and it's the only one resolved, then just output the value
+      if resolved_config.count == 1 && resolved_config.keys[0] == item_name
+        puts _('%{value}') % { value: resolved_config.values[0] }
+        exit 0
+      end
+      # Otherwise just output everything
+      resolved_config.keys.sort.each { |key| puts _('%{name}=%{value}') % { name: key, value: resolved_config[key] } }
+    end
+  end
+end
--- a/lib/pdk/config/namespace.rb
+++ b/lib/pdk/config/namespace.rb
@@ -141,6 +141,28 @@ module PDK
         end
       end
 
+      # Resolves all filtered settings, including child namespaces, fully namespaced and filling in default values.
+      #
+      # @param filter [String] Only resolve setting names which match the filter. See #be_resolved? for matching rules
+      # @return [Hash{String => Object}] All resolved settings for example {'user.module_defaults.author' => 'johndoe'}
+      def resolve(filter = nil)
+        # Explicitly force values to be loaded if they have not already
+        # done so. This will not cause them to be persisted to disk
+        (@values.keys - data.keys).each { |key_name| self[key_name] }
+        resolved = {}
+        data.each do |data_name, obj|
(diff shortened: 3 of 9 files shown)


glennsarti:
Looked at adding json formatter but it's waaay outside the scope of this PR and should not stop merging of this PR.  Punting for now

glennsarti:
Not sure about this though:

https://github.com/puppetlabs/pdk/pull/715/files#diff-583f8159ac861abd5736f8e9c611c396R35

If the user specifies a non-existant setting, should it exit with non-zero?

lib/pdk/cli/config.rb
```diff
@@ -0,0 +1,20 @@
...
+    summary _('Configure the Puppet Development Kit.')
+    default_subcommand 'get'
+
+    run do |_opts, args, _cmd|
+      if args == ['help']
+        PDK::CLI.run(%w[config --help])
+        exit 0
+      end
+
+      PDK::CLI.run(%w[config get]) if args.empty?
```
rodjek: Our usual behaviour is to have the default subcommand be `help` (with the exception of `pdk validate` which is a special case that combines the behaviour of all the subcommands). I would suggest that `help` be the default here as well for consistency.

spec/acceptance/config_get_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+describe 'pdk config get' do
+  include_context 'with a fake TTY'
+
+  context 'when run outside of a module' do
+    describe command('pdk config get') do
+      its(:exit_status) { is_expected.to eq 0 }
+      # This setting should appear in all pdk versions
+      its(:stdout) { is_expected.to match(%r{user\.analytics\.user-id=}) }
+      its(:stderr) { is_expected.to be_empty }
```
rodjek: I generally prefer to use `match(%r{\A\Z})` rather than `be_empty` as in the event of a test failure you get a diff showing the contents of the expected empty value rather than `expected empty? to be true, was false`. To that end, I have a custom `have_no_output` matcher in the acceptance suite to wrap this up in an nice DSL (`its(:stderr) { is_expected.to have_no_output }`)

glennsarti:
> If <key> is a leaf node of the configuration graph, this command will return the raw value

As there is no formal spec on what settings should appear where (i.e. what's a leaf vs, what's a tree) it get's tricky.  So instead I guess. If user specified `x.y.z` and only one setting with the name `x.y.z` is returned then, it assumes it's a leaf.

lib/pdk/cli/config.rb
glennsarti: Fixed

spec/acceptance/config_get_spec.rb
glennsarti: Fixed all instances of `empty?`

lib/pdk/config/namespace.rb
```diff
@@ -21,11 +21,12 @@ class Namespace
       # @option params [self] :parent the parent {self} that this namespace is
       #   a child of (defaults to nil).
       # @param block [Proc] a block that is evaluated within the new instance.
-      def initialize(name = nil, file: nil, parent: nil, &block)
+      def initialize(name = nil, file: nil, parent: nil, persistent_defaults: false, &block)
```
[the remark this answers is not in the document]
glennsarti: Added

glennsarti:
@scotje I've ticketted changes to the RFC to match this PR at https://github.com/puppetlabs/pdk-planning/issues/40

Once that is issue is approved and merged, this PR can be merged.

scotje:
> As there is no formal spec on what settings should appear where (i.e. what's a leaf vs, what's a tree) it get's tricky. So instead I guess. If user specified `x.y.z` and only one setting with the name `x.y.z` is returned then, it assumes it's a leaf.

Yeah, we had talked about having a config schema we could validate against to make this easier to determine as well. This section of the RFC was a initial draft of allowed keys and values at the various layers: https://github.com/puppetlabs/pdk-planning/blob/master/RFCs/0002-add-pdk-config.md#proposed-configuration-keys but that will certainly be expanded/revised as we go along. I do think we should do some sort of checking against a schema to make sure people don't typo a setting key and then wonder why it's not working.

glennsarti:
I did see that.  Probably best to introduce a schema validator now actually

glennsarti:
Back to WIP for you!

glennsarti:
Blocked on #726 merge

scotje:
RFC update has been merged

glennsarti:
@scotje Regarding `If <key> is a leaf node of the configuration graph, this command will return the raw value:`

While I don't have a schema to work with, the tests show the current logic exhibits the behaviour the RFC sets out.

lib/pdk/cli/config/get.rb
```diff
@@ -0,0 +1,24 @@
+module PDK::CLI
+  @config_get_cmd = @config_cmd.define_command do
+    name 'get'
+    usage _('config get [name]')
+    summary _('Retrieve the configuration for <name>. If not specified, retrieve all configuration settings')
+
+    run do |_opts, args, _cmd|
+      item_name = args[0]
+      resolved_config = PDK::Config.new.resolve(item_name)
```
rodjek:
```suggestion
      resolved_config = PDK.config.resolve(item_name)
```
There's a memoised accessor for `PDK::Config` that you can use instead of instantiating a new one

glennsarti: Good catch! fixed.

glennsarti:
CI Is green and approvals done.  Merging.

PR #723, opened 2019-08-06
Handle deleted/unmanaged files for new module

seanmil:
When creating a new module, only process templates for files which
are not marked as deleted or unmanaged. This permits non-standard
PDK templates with files set to delete or unmanaged to be used
for new module creation without error.


seanmil:
Note that I have tested this successfully, but was uncertain about how to add automated tests to cover it.

seanmil:
Never mind. Closing in favor of #710 which I just found and better addresses it anyways.

PR #725, opened 2019-08-07
Handle deleted template files for new module

seanmil:
When using `pdk new module` to create a new module, handle the
situation where the template used contains one or more files
with `delete: true` in `config_defaults.yml` instead of causing
an error.

### DIFF
--- a/lib/pdk/generate/module.rb
+++ b/lib/pdk/generate/module.rb
@@ -58,7 +58,8 @@ module PDK
 
         begin
           PDK::Module::TemplateDir.new(template_uri, metadata.data, true) do |templates|
-            templates.render do |file_path, file_content|
+            templates.render do |file_path, file_content, file_status|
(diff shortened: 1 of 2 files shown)


lib/pdk/generate/module.rb
```diff
@@ -58,7 +58,8 @@ def self.invoke(opts = {})
 
         begin
           PDK::Module::TemplateDir.new(template_uri, metadata.data, true) do |templates|
-            templates.render do |file_path, file_content|
+            templates.render do |file_path, file_content, file_status|
```
scotje: I must be missing something but where does `file_status` come from?

scotje: Ah, I see, you linked that in the other thread: https://github.com/puppetlabs/pdk/blob/master/lib/pdk/module/templatedir.rb#L125-L130

PR #726, opened 2019-08-08
(PDK-1107) Config fetch and [] should have no side effects

glennsarti:
Previously even querying a configuration setting would save the default value to
disk. This is undesirable, and not expected as a user, because not all defaults
should persist on disk, across multiple PDK sessions/projects/modules. This
commit:

* Modifies the Config::Namespace class to have a new parameter called
  persistent_defaults.
* Modified the [] and fetch methods in Namespace to optioally call save_data
  when a default value is queried and persistent_defaults is set to true. This
  is affected in #default_config_value method.
* Added a wrapper method for mkdir_p for the FileSystem class so it could be
  mocked correctly in tests.  Currently it was creating folder structures in the
  root of the PDK project.
* Updated the namespace and config tests to ensure that the analytics settings
  use the persistent_defaults Namespace setting.

### DIFF
--- a/lib/pdk/config.rb
+++ b/lib/pdk/config.rb
@@ -15,7 +15,7 @@ module PDK
       @user ||= PDK::Config::JSON.new('user', file: PDK::Config.user_config_path) do
         mount :module_defaults, PDK::Config::JSON.new(file: PDK.answers.answer_file_path)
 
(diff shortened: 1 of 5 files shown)


glennsarti:
@scotje Yes.  I extracted this commit from `pdk config get` so I can at least get something merged.  I also need this for the pdk config schema PR

glennsarti:
Will wait for +1 from @rodjek before merge.

PR #727, opened 2019-08-08
(docs) minor fixups to README

jbondpdx:
removing an incomplete sentence and an article in front of PDK

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,13 +107,11 @@ This command runs all available unit tests.
 
 ### `pdk bundle` command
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,13 +107,11 @@ This command runs all available unit tests.
 
 ### `pdk bundle` command
 
-This command executes arbitrary commands in a bundler context within the module you're currently working on. Arguments to this command are passed straight through to bundler. This command is experimental  and can lead to errors that can't be resolved by the pdk itself.
-
-Note that for most uses, you must use the `--` to separate bundler options from pdk options. Compare the following two commands:
```
scotje: This is a thing we probably still want to document in some way, but perhaps would be better lived in some real `pdk bundle` documentation that we could just link to.

scotje: @rodjek tells me the `--` isn't needed anymore so nevermind!

PR #728, opened 2019-08-09
{WIP}(PDK-1112) Use schema documents for configuration settings

glennsarti:
- [x] Need to add a check for  schemaificated namespaces that validate the document upon load.  
- [x] Needs tests for schemaificated namespaces
  - [x] settings not in the schema should be preserved, but not surfaced.
- [x] Needs tests the actual schema files themselves
- [ ] Need to split up the commits into modifying existing stuff, and then building on it.
- [ ] Address all TODOs
- [x] ~~Need to change the json-schema gem~~ won't do.  There doesn't seem to be a stable alternative yet for Ruby 2.1


glennsarti:
Closed in favour of #742

PR #729, opened 2019-08-12
(PDK-1363) Apply init templates during module convert

rodjek:
During `pdk convert`, if there are init-only templates for files that do not exist in the module, create them as would be done during `pdk new module`.

Coverage dropped a bit because there isn't much in the way of unit tests for `PDK::Module::TemplateDir` currently and I don't want this change to get bogged down working on that.

### DIFF
--- a/lib/pdk/module/convert.rb
+++ b/lib/pdk/module/convert.rb
@@ -74,7 +74,7 @@ module PDK
 
           if options[:noop] && new_metadata.nil?
             update_manager.add_file(metadata_path, '')
-          elsif File.file?(metadata_path)
+          elsif PDK::Util::Filesystem.file?(metadata_path)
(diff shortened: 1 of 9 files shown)


lib/pdk/module/convert.rb
```diff
@@ -68,25 +72,30 @@ def needs_bundle_update?
       def stage_changes!
         metadata_path = 'metadata.json'
 
-        PDK::Module::TemplateDir.new(template_uri, nil, false) do |templates|
+        PDK::Module::TemplateDir.new(template_uri, nil, true) do |templates|
```
scotje: The `PDK::Module::TemplateDir` constructor seems like a good candidate for keyword arguments. Not necessarily now though.

PR #731, opened 2019-08-20
(PDK-1384) Updates to be compatible with latest Cri

scotje:


scotje:
There are still a bunch of places where we call `hash_including` but the tests aren't failing, so I'm not sure if we need to update them? They may be testing cases where options are being explicitly passed.

scotje:
Tweaked `Gemfile` and some of the test expectations to hopefully be compatible with both old and new Cri back to Ruby 2.1.9.

glennsarti:
Hrmm odd error in appveyor there.

```
+puppet-syntax: C:/Ruby24-x64/lib/ruby/gems/2.4.0/gems/bundler-2.0.2/lib/bundler/spec_set.rb:87:in `block in materialize': Could not find i18n-0.9.5 in any of the sources (Bundler::GemNotFound)
````

It's in the `build install` so ... 🤷‍♂️

scotje:
Yeah, I was trying to repro that in my VM yesterday, going to just re-kick the Appveyor build and see what happens....

scotje:
Well it's happy now. 🤷‍♂

scotje:
Ping @puppetlabs/pdk for review now that this is passing.

PR #732, opened 2019-08-20
(PDK-1465) Remove net-ssh from gemspec

glennsarti:
Previously the net-ssh gem was required for beaker on older ruby versions
however this dependency is being moved to the module metagems.  This commit
removes the gem from the gemspec.


glennsarti:
Blocked on gems release of the metagem. See PDK-1465 for more information

scotje:
meta gems have been released with this change (0.4.2)

PR #733, opened 2019-08-21
(PDK-1464) Update nokogiri due to CVE-2019-5477

glennsarti:
This commit updates the minimum nokogiri version to 1.10.4 to mitigate against
CVE-2019-5477.  The PDK package will be updated in a separate commit.


glennsarti:
Confirmed in the Travis and Appveyor logs that it uses 1.10.4 except for Ruby 2.1 and 2.3 which uses 1.7.2

PR #734, opened 2019-08-22
(Performance) - lazy load the generate class

logicminds:
* on systems like windows where the execution time of pdk
    can take 2000-19000 ms performance is a bid deal.

  * lazy loading as many libraries as possible will make a
    big difference in large corporations where windows and antivirus
    file scanning dominate the memory space.

  * The pdk/module/generate class was not required in the cli class and was
    removed. After removal the template_uri class needed the addressable
    library was must have been loaded elsewhere before this patch.  The
    addressable library was added to template_uri in order to satisfy
    the dependency.

  * Users on windows can expect a slight speedup by about
    1000 ms on heavily loaded systems.  Every seconds counts.

  * 1000 ms is an eternity to a computer.


glennsarti:
Welp...my additional commit didn't help that much either. Looks like we'll need to run the acceptance tests locally to see where the issues are

logicminds:
closing due to #754

PR #735, opened 2019-08-23
(PDK-1175) pdk new unit_test

rodjek:
Introduces a new pdk command `pdk new unit_test` which can be used to generate a skeleton unit test for objects like existing classes and defined types. This is **not** intended to provide retrospec like functionality, but simply provide people with the same "it compiles" skeleton test that the existing generators do automatically.


glennsarti:
@rodjek and I had a chat via Slack, and general consensus is pretty much what @scotje said:

1. Yes the name probably needs to change

2. We lack shared context around what each Verb actually means so right now it's too easy to bikeshed

3. It's safe to ship this as "experimental"/"dark ship", and once we get a better naming standard we can move the functionality there.

glennsarti:
image

PR #742, opened 2019-09-02
(PDK-1112) Create json schema to validate pdk config file

glennsarti:
- [x] Address TODOs
- [x] Finish yard docs for methods
- [x] What happens if the file is not valid BEFORE the PDK tries to read it? We can never save it due to the validation.
---

Previously the Config Namespace class conflated a few concepts together and had
less than ideal method naming.  This commit:

* Changes the name of a namespace value to a namespace setting. This is because
  the class held more than just a value, such as the default value and
  validators. Therefore it made more sense to call it a a Namespace Setting.

* The namespace had an abstract method called parse_data which was passed both
  the data and the name to the file. This didn't make sense as the filename is
  all that's required, not the filename.  The method was also renamed to
  parse_file which is more accurate name.

* The parse_file used to return an array of objects, but this create extra array
  objects during parsing. This commit changes the method to instead yield
  namespace setting objects.  This will allow namespaces to create their own
  settings objects instead of only using the one.

* Refactors common file based namespace tests to an rspec shared example.

* Adds tests for assigning values to a setting

---

Previously there was no formal schema, or a way to validate a schema, for
configuration settings. This commit adds two new namespace types, YAML and
JSON files with JSON schema validation.

These namespaces will also take a JSON Schema file and use that to provide
defaults and validation.  Any settings NOT in the schema will not be visible in
the namespace, but will not be lost when saving the namespace to disk; That is
the namespace will only modify settings that are in the schema. This also
includes tests for any schema files, to ensure that they are indeed valid.

This commit adds a schema for the PDK analytics configuration.

This commit introduces a new setting type, JSONSchemaSetting which uses the
schema to validate and get default values for a setting from the schema.

---

This commit adds the tests for the new schema based namespaces. This commit
also refactors common tests between JSON, YAML and non-schema namespaces into
rspec shared example groups.

Note that due to a bug in json-schema object, the tests fail on non-Windows
platforms.

---

Unfortunately the json-schema gem fails to load filed based schema on Windows
due to URI encoding issues. Also the json-schema gem is abandoned so it get
fixed either, and other gems don't have support for old ruby 2.1. This commit
monkey patches the JSON Schema read_file method to munge the pathname on
Windows platforms.

This commit removes the test guards on non-Windows platforms.


lib/pdk/config/json_schema_namespace.rb
```diff
@@ -0,0 +1,123 @@
...
+      end
+
+      # Sets the list of unmanaged settings. For subclass use only
+      #
+      # @param value [Hash<String, Object]] A hashtable of all unmanaged settings which will be persisted, but not visible
+      #
+      # @private
+      def assign_unmanaged_settings(value)
+        @unmanaged_settings = value
+      end
```
scotje: Should this be `protected` if it's only for use by subclasses?

lib/pdk/config/json_schema_namespace.rb
```diff
@@ -0,0 +1,123 @@
...
+
+        if @schema_file.nil?
+          @document_schema = create_empty_schema
+          return @document_schema
+        end
+        unless PDK::Util::Filesystem.file?(@schema_file)
+          @document_schema = create_empty_schema
+          raise PDK::Config::LoadError, _('Unable to open %{file} for reading. File does not exist') % {
+            file: @schema_file,
+          }
```
scotje: Does this still need to `create_empty_schema` if it's going to `raise` right after? Is there a use case where the exception is caught and `@document_schema` is expected to be set?

lib/pdk/config/json.rb
```diff
@@ -3,16 +3,26 @@
...
+        data = load_data(filename)
+        return if data.nil? || data.empty?
 
         require 'json'
 
-        ::JSON.parse(data)
+        data = ::JSON.parse(data)
+        return if data.nil? || data.empty?
+
+        data.each { |k, v| yield k, PDK::Config::Setting.new(k, self, v) }
```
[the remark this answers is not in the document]
scotje: (Otherwise it seems like this would raise a `LocalJumpError` if no block was given?)

lib/pdk/config/json_schema_namespace.rb
scotje: (I see it does the same thing down below so I assume that side effect is intended?)

lib/pdk/config/setting.rb
```diff
@@ -1,35 +1,60 @@
...
     #     validate PDK::Config::Validator.boolean
     #     default_to { false }
     #   end
     # end
-    class Value
-      # Initialises an empty value definition.
+    class Setting
+      attr_reader :namespace
+
+      attr_writer :previous_setting
```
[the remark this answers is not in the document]
glennsarti: Only in the commit message.  You thinking comments here?

lib/pdk/config/json.rb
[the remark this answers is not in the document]
glennsarti: Umm.... given this is internal, and the rspec shared examples test for it.... I don't think there's much risk of it not having a block.  That said I'm not against it.

lib/pdk/config/json_schema_namespace.rb
glennsarti: TIL `protected`.  I made it public because I use it in tests.  Could be a code smell that the code is written as testable as it could.

lib/pdk/config/json_schema_namespace.rb
glennsarti: It is possible for namespace to not care that a schema failed. So I needed some kind of value.  That said, I could perhaps assign it better/earlier? so it doesn't look as strange.

lib/pdk/config/json.rb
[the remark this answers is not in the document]
scotje: Yeah, I would like to see either `raise unless block_given?` at the top or making the block param explicit in the argument list, just so it's clear to future us that this method requires a block.

lib/pdk/config/setting.rb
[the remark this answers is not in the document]
scotje: Yeah a comment somewhere was what I was thinking.

lib/pdk/config/json_schema_namespace.rb
scotje: If you just need to call it for test setup you can always bypass the visibility check with https://ruby-doc.org/core-2.6.4/Object.html#method-i-send

Or you could use https://ruby-doc.org/core-2.6.4/Object.html#method-i-instance_variable_set to bypass the method entirely if it doesn't do anything special.

Also should we just call this method `def unmanaged_settings=(value)` ?

lib/pdk/config/json_schema_namespace.rb
scotje: This case could potentially just be another `rescue` for whatever `JSON::Schema::Reader` raises if the file doesn't exist? (Unless of course it just raises `::JSON::Schema::JsonParseError` :/) That might clean up the rest of the flow control here.

lib/pdk/config/setting.rb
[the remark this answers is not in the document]
glennsarti: Done.

lib/pdk/config/yaml.rb
```diff
@@ -1,14 +1,22 @@
...
+        data = load_data(filename)
+        return if data.nil? || data.empty?
 
         require 'yaml'
 
-        ::YAML.safe_load(data, [Symbol], [], true)
+        data = ::YAML.safe_load(data, [Symbol], [], true)
+        return if data.nil?
+
+        data.each { |k, v| yield k, PDK::Config::Setting.new(k, self, v) }
```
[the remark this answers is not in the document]
glennsarti: Done.

lib/pdk/config/json_schema_namespace.rb
glennsarti: Done.

glennsarti: Fixed.  Uses an empty schema as the "default" value.

lib/pdk/config/json_schema_namespace.rb
glennsarti: Went for a protected `attr_writer` for unmanaged_settings.

Added a TODO that we should probably subclass instead of using `instance_variable_set` but that shouldn't stop merging.

PR #750, opened 2019-09-16
Add release command and change log generator

carabasdaniel:


lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+          end
+
+          output = File.read('CHANGELOG.md')
+
+          raise 'Uncategorized PRs; Please go label them' if output =~ %r{UNCATEGORIZED PRS; GO LABEL THEM}
+        end
+
+        def get_next_version(current_version)
+          current_version_map = current_version.split('.').map { |v| v.to_i }
+          puts '# Extracting target version from CHANGELOG.md'
```
glennsarti: Should be using the PDK logging classes, not `puts`

lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
+module PDK
+  module CLI
+    module Util
+      class ChangelogGenerator
+        def generate_changelog
```
glennsarti: Should be guarded here on whether the changelog rake task/gem is available

lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+          retries = 0
+          result = nil
+          loop do
+            result = system('bundle exec rake changelog')
+            retries += 1
+            break if result || retries > 3
```
glennsarti: This doesn't make any sense.  Running the same command three times makes it magically work?

lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+          end
+
+          unless result
+            raise 'Error generating changelog'
+          end
+
+          output = File.read('CHANGELOG.md')
```
glennsarti: Dangerous defaults.  What about line endings or encoding?

lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+
+        def get_next_version(current_version)
+          current_version_map = current_version.split('.').map { |v| v.to_i }
+          puts '# Extracting target version from CHANGELOG.md'
+
+          # Grab all lines that start with ## between two lines that start with the version to detect changes
+          data = ''
+          matches = 1
+          File.foreach('CHANGELOG.md') do |line|
```
glennsarti: Assumes `pwd` is in a module.  This is a dangerous assumption.

glennsarti:
Missing RFC or link to an existing one.

carabasdaniel:
Thanks for the quick feedback @glennsarti, the RFC is this: https://github.com/puppetlabs/pdk-planning/blob/master/RFCs/0003-add-pdk-release.md, but this is still WIP.

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: This utility class is used in the release command, where there is an ensure that the pwd is in the module, see line: https://github.com/puppetlabs/pdk/pull/750/files#diff-87ad85e8da7245ccfda95cfed0b88327R19

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: Will change this to PDK logger

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: Will add a check here

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: I guess there is no need for a retry here, therefore I will remove it. Initially had networking issues with this but the errors were not very explicit. Thanks.

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: Will fix this

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: As stated in the template here: https://github.com/puppetlabs/pdk-templates/blob/master/moduleroot/Rakefile.erb#L75 there is already a check in the Rakefile

glennsarti: That assumes the person is using the pdk-template, or has not overridden it to not have the gem.

carabasdaniel: Added exception if gem not installed

carabasdaniel:
More tests will be added in a future PR

lib/pdk/cli/release.rb
```diff
@@ -0,0 +1,295 @@
...
+        # Create new changelog with the correct version
+        changelog_generator.generate_changelog
+        PDK.logger.info _("New changelog generated with version #{new_version}")
+      end
+
+      if opts[:skip_documentation]
+        PDK.logger.info _('Skipping documentation update for this module')
+      else
+        PDK.logger.info _('Updating documentation using puppet strings')
+        result = system('bundle exec puppet strings generate --format markdown --out REFERENCE.md')
```
glennsarti: This won't work in the packaged versions of PDK.  See `lib/pdk/cli/bundle.rb` for more examples as to how to run a bundle commend in the context of a module.

lib/pdk/cli/release.rb
```diff
@@ -0,0 +1,295 @@
...
+        # run dependency-checker and output dependent modules list
+        PDK.logger.info _('Running dependency checker for %{module_name} - version %{module_version}') % {
+          module_name:    module_metadata.data['name'],
+          module_version: module_metadata.data['version'],
+        }
+
+        PDK.logger.warn _('Dependency-checker failed to execute, please make sure the gem is installed') unless system('dependency-checker metadata.json')
+      end
+
+      # TODO: Ensure forge metadata has been set, or call out to interview
```
glennsarti: nit: Remaining TODO

lib/pdk/cli/release.rb
```diff
@@ -0,0 +1,295 @@
...
+        end
+      end
+
+      builder = PDK::Module::Build.new(opts)
+
+      unless opts[:force]
+        if builder.package_already_exists?
+          PDK.logger.info _("The file '%{package}' already exists.") % { package: builder.package_file }
+
+          unless PDK::CLI::Util.prompt_for_yes(_('Overwrite?'), default: false)
```
glennsarti: Should there be an option to ignore the overwrite? Only way to avoid this is call --force which has other implications

lib/pdk/cli/release.rb
```diff
@@ -0,0 +1,295 @@
...
+        package_name: module_metadata.data['name'],
+        package_path: builder.package_file,
+      }
+
+      if opts[:apikey]
+        # TODO: Replace this code when the upload functionality is added to the forge ruby gem
+        file_data = Base64.encode64(File.open(builder.package_file, 'rb').read)
+
+        PDK.logger.info _('Uploading tarball to puppet forge...')
+        uri = URI('https://forgeapi.puppetlabs.com/v3/releases')
```
glennsarti: This should be configurable?

lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
+require 'bundler'
+
+module PDK
+  module CLI
+    module Util
+      class ChangelogGenerator
+        def generate_changelog
+          raise 'github_changelog_generator gem not installed' unless Bundler.rubygems.find_name('github_changelog_generator').any?
+          result = system('bundle exec rake changelog')
```
glennsarti: See above.  This won't work in package versions of PDK

lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+
+          unless result
+            raise 'Error generating changelog'
+          end
+
+          output = PDK::Util::Filesystem.read_file('CHANGELOG.md')
+
+          raise 'Uncategorized PRs; Please go label them' if output =~ %r{UNCATEGORIZED PRS; GO LABEL THEM}
```
glennsarti: This error message is not very communicative.  Which PRs? how do you label them? what label?

lib/pdk/cli/util/changelog_generator.rb
glennsarti: This still feels like it needs work.  What if the file doesn't exist? What if the file is empty? What if the file isn't encoded correctly? What encoding does it even use? Will it break lines correctly on Windows (CRLF) and Linux (LF) correctly?

lib/pdk/cli/util/changelog_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+            current_version_map[1] = 0
+            current_version_map[2] = 0
+          when %r{^### Added}
+            current_version_map[1] += 1
+            current_version_map[2] = 0
+          when %r{^### Fixed}
+            current_version_map[2] += 1
+          else
+            PDK.logger.info _('No notable changes to release')
+            exit 1
```
glennsarti: This seems like a VERY dangerous place to put an exit statement.  The caller should decide whether to exit, not this method.

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: Using PDK::Util::Filesystem.read_file now

lib/pdk/cli/release.rb
carabasdaniel: Fixed, thanks

lib/pdk/cli/release.rb
carabasdaniel: The easiest method is to use force for this. As this command is meant to be interactive I don't think it should be problematic.

lib/pdk/cli/release.rb
carabasdaniel: Set it as an option

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: Pointed the used to the generated file

lib/pdk/cli/util/changelog_generator.rb
carabasdaniel: Removed exit statement

glennsarti:
@carabasdaniel I've raised https://github.com/puppetlabs/pdk/pull/809 as an alternative.

- Moved the actual working methods under `PDK::Module::Release`.  So now we have separate of UI and Logic
- Moved changelog generator uder PDK::Util
- Added a whole bunch of tests.
- Used the preferred hypens in command line args instead of underscores
- Actively remove tokens being passed to analytics
- Cleaned up some of the logic errors in your original PR
- Stopped skipping CA Cert verification

carabasdaniel:
Closing PR as https://github.com/puppetlabs/pdk/pull/809 got merged

PR #753, opened 2019-09-18
(maint) - Allow no output when reading puppet_from_opts

logicminds:
* Previously the output of the puppet and ruby version used
    was displayed on every run.  This adds a new parameter to allow
    for the developer to choose when this output should be displayed.
    This is useful for interactive commands that would also display
    the same info.

  * Refactors the conditional logic used to pluck out values from the
    hash using the fetch method.

### DIFF
--- a/lib/pdk/cli/util.rb
+++ b/lib/pdk/cli/util.rb
@@ -117,10 +117,17 @@ module PDK
       end
       module_function :check_for_deprecated_puppet
 
-      def puppet_from_opts_or_env(opts)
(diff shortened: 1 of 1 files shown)


lib/pdk/cli/util.rb
```diff
@@ -150,6 +159,7 @@ def puppet_from_opts_or_env(opts)
         gemset.each do |gem, version|
           next if version.nil?
 
+          next if logging_disabled
```
rodjek: I'd put this conditional outside of the loop as there's no need to enter the loop at all if you're not logging.

logicminds: Updated in latest push.

PR #755, opened 2019-09-19
(feat) - Add new utility method for fixtures dir

logicminds:
* Without this, a developer would have to piece together
    the same code in multiple places to create the fixtures path
    for the module.  This adds a new util method called
    module_fixtures_dir that returns the ./spec/fixtures directory
    of the module.

### DIFF
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -132,6 +132,14 @@ module PDK
     end
     module_function :module_root
 
+    # The module's fixtures directory for spec testing
+    # @return [String] - the path to the module's fixtures directory
+    def module_fixtures_dir
+      dir = module_root
+      File.join(module_root, 'spec', 'fixtures') unless dir.nil?
+    end
+    module_function :module_fixtures_dir
+
     # Returns true or false depending on if any of the common directories in a module
     # are found in the current directory
     #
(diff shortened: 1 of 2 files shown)


lib/pdk/util.rb
```diff
@@ -132,6 +132,14 @@ def module_root
     end
     module_function :module_root
 
+    # The module's fixtures directory for spec testing
+    # @return [String] - the path to the module's fixtures directory
+    def module_fixtures_dir
+      dir = module_root
+      File.join(module_root, 'spec', 'fixtures') unless dir.nil?
```
glennsarti: Adding a PR note here (won't stop merging) `<root>/spec/fixtures` is hardcoded here and (probably) comes from the `puppetlabs_spec_helper` gem

https://github.com/puppetlabs/puppetlabs_spec_helper/blob/9d66ff58eba8bd4e21bb31c638a6ef6e9ac8cd14/lib/puppetlabs_spec_helper/tasks/fixtures.rb#L268-L363

IIRC There is no way to override this, so hardcoding here is fine.

glennsarti:
Curious, should this belong in the puppetlabs_spec_helper gem (and namespace?)

logicminds:
> Curious, should this belong in the puppetlabs_spec_helper gem (and namespace?)

PDK doesn't explicitly require puppetlabs_spec_helper so we can't use that method.  The pdk console command is going to use this new function to auto set the module path going forward.

Issue #759, opened 2019-09-27
error testing against 2.1.9 with public_suffix


logicminds:
**Describe the bug**
Travis uses a build cache with preset gems.  However, when using a fresh container and fresh install, bundler fails to install the gem dependencies.

```
root@7abc1d33208c:/app# bundle install --without development --path=${BUNDLE_PATH:-vendor/bundle}
Fetching gem metadata from https://rubygems.org/............
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies......
Fetching rake 10.5.0
Installing rake 10.5.0
Fetching public_suffix 4.0.1
Installing public_suffix 4.0.1
Gem::InstallError: public_suffix requires Ruby version >= 2.3.
An error occurred while installing public_suffix (4.0.1), and Bundler cannot continue.
Make sure that `gem install public_suffix -v '4.0.1' --source 'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  pdk was resolved to 1.14.0.pre, which depends on
    json-schema was resolved to 2.8.0, which depends on
      addressable was resolved to 2.7.0, which depends on
        public_suffix
```

**To Reproduce**

```
git clone https://github.com/puppetlabs/pdk
cd pdk
gem install bundler -v 1.17.3
bundle install --without development --path=${BUNDLE_PATH:-vendor/bundle}

```
**Expected behavior**
Successful bundle install

DavidS:
We're dropping ruby 2.1 support in the upcoming PDK 2.0 release. Hence, closing this wontfix.

To keep your module infrastructure compatible with older rubies, folks need to stay on the correct PDK and pdk-template versions AND maintain a clone of puppet-module-gems with continually updated pins for breaking changes like this.

PR #760, opened 2019-09-27
Fixes gh-759 ruby 2.1.9 support

logicminds:
* previously the pdk gem was able to be installed
    under ruby 2.1.9.  However, there were some recent
    upstream dependency changes that dropped support for 2.1.9
    which caused pdk to be no longer installable.

    This fixes the gem dependency issue by backing down some gems
    that were forcing a newer ruby version.

### DIFF
--- a/pdk.gemspec
+++ b/pdk.gemspec
@@ -3,25 +3,26 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'pdk/version'
 
 Gem::Specification.new do |spec|
-  spec.name    = 'pdk'
+  spec.name = 'pdk'
   spec.version = PDK::VERSION
   spec.authors = ['Puppet, Inc.']
-  spec.email   = ['pdk-maintainers@puppet.com']
+  spec.email = ['pdk-maintainers@puppet.com']
 
-  spec.summary     = 'A key part of the Puppet Development Kit, the shortest path to better modules'
+  spec.summary = 'A key part of the Puppet Development Kit, the shortest path to better modules'
   spec.description = 'A CLI to facilitate easy, unified development workflows for Puppet modules.'
-  spec.homepage    = 'https://github.com/puppetlabs/pdk'
+  spec.homepage = 'https://github.com/puppetlabs/pdk'
 
-  spec.files         = Dir['CHANGELOG.md', 'README.md', 'LICENSE', 'lib/**/*', 'exe/**/*', 'locales/**/*']
-  spec.bindir        = 'exe'
-  spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
+  spec.files = Dir['CHANGELOG.md', 'README.md', 'LICENSE', 'lib/**/*', 'exe/**/*', 'locales/**/*']
+  spec.bindir = 'exe'
+  spec.executables = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
   spec.required_ruby_version = '>= 2.1.9'
 
+  spec.add_runtime_dependency 'addressable', '2.6.0'
   spec.add_runtime_dependency 'bundler', '>= 1.15.0', '< 3.0.0'
   spec.add_runtime_dependency 'childprocess', '~> 0.7.1'
-  spec.add_runtime_dependency 'cri', '~> 2.10'
+  spec.add_runtime_dependency 'cri', '= 2.10.1'
   spec.add_runtime_dependency 'diff-lcs', '1.3'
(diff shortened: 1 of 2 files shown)


pdk.gemspec
```diff
@@ -3,25 +3,26 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
...
-  spec.bindir        = 'exe'
-  spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
+  spec.files = Dir['CHANGELOG.md', 'README.md', 'LICENSE', 'lib/**/*', 'exe/**/*', 'locales/**/*']
+  spec.bindir = 'exe'
+  spec.executables = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
   spec.required_ruby_version = '>= 2.1.9'
 
+  spec.add_runtime_dependency 'addressable', '2.6.0'
```
logicminds: wondering if this should be <= 2.6.0 ?

glennsarti: The changelog and gemspec for addressable say it _should_ work for Ruby 2.1 anyway, so perhaps we don't need it at all? `addressable` depends on `public_suffix` which dropped ruby 2.1 support in 4.x (but ok in 3.x)

logicminds: I only added addressable here to force the version as the changelog generator was too permissive for addressable gem and caused this problem.

logicminds: Also now rake updated on the 27th.   So will need to force this to less than 13.0.0.

glennsarti: The never-ending story dependency story :-(

logicminds:
Seems to be a rogue test failing.

glennsarti:
OK, there's in fix in master now. May need to rebase.

rodjek:
Yep, a rebase off master should clear up the build failure

scotje:
Looks like `rubyzip` dependency is now being declared twice for Ruby 2.1.9. (Second declaration in in the `test` group.) So we need to consolidate those.

pdk.gemspec
```diff
@@ -3,25 +3,26 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
...
   spec.add_runtime_dependency 'bundler', '>= 1.15.0', '< 3.0.0'
   spec.add_runtime_dependency 'childprocess', '~> 0.7.1'
-  spec.add_runtime_dependency 'cri', '~> 2.10'
+  spec.add_runtime_dependency 'cri', '= 2.10.1'
```
scotje: This makes PDK bundle incompatible with `r10k` which was an issue for people previously. (See also: https://github.com/puppetlabs/pdk/pull/731)

pdk.gemspec
scotje: Is "addressable" really needed at runtime by the PDK gem? Is the `Gemfile` override sufficient instead?

pdk.gemspec
logicminds: This is required for 2.1 support unless we do a <= 2.10.1

scotje: It's not _required_ since `2.10.1` satisfies `~> 2.10`, it just means you have to utilize some other means to pin `cri` since Rubygems is so dumb about how it considers Ruby version requirements in gems. I think it would be better for us to document those other means than break the co-installability of `pdk` with a growing number of other ecosystem tools for the people on modern Rubies...

logicminds:
> Is "addressable" really needed at runtime by the PDK gem? Is the Gemfile override sufficient instead?

I thought there was a specific reason I did this.   The problem is the current tests use cached gems that hide the skeletons awaiting for gem updates.

scotje:
Hmmm, maybe we can add a `bundle update` to the tests somewhere?

pdk.gemspec
logicminds: `~> 2.10` would pull in 2.15 which starts to bring in dependencies that have requirement issues with ruby 2.1

scotje: It should only pull in the latest if it's not already satisfied by an existing version of the gem.

logicminds:
Keep in mind this is only problem if: 

1. You install PDK as a gem
2. Still running ruby 2.1

I don't really know who fits in this category.   Except for old man Jenkins and his buddy Travis.

logicminds:
scotje,  Can this be closed?  I think bits and pieces of this PR have found their way into the dependency list already.  Was bundle update added to the tests?

scotje:
Yeah let's close this for now. We are tentatively planning a 2.0.0 release early next year that would drop Ruby 2.1 support anyway.

PR #762, opened 2019-10-01
Improve bundler speed

logicminds:
This is a WIP left here for comments.  This code negates the need to shell out and runs the same check with pure ruby instead.  This should prove to be a nice speed bump when checking if bundler has the gem or not.


logicminds:
I added #774  so we could see the difference in performance.  The performance difference is pretty drastic 1300 ms vs 10 ms.  While the numbers speak for themselves I still need to ensure installed? is working correctly.

spec/unit/pdk/util/bundler_spec.rb
```diff
@@ -347,52 +348,30 @@ def bundle_regex
...
-
-        instance.installed?
-      end
-
-      it 'returns true if `bundle check` exits zero' do
-        allow_command([bundle_regex, 'check', "--gemfile=#{gemfile}", '--dry-run'], exit_code: 0)
-
+      it 'returns true' do
         expect(instance.installed?).to be true
       end
```
scotje: Maybe wrap this expectation in a context that says "when all gems present" just so the full test description is more clear.

lib/pdk/util/bundler.rb
```diff
@@ -92,18 +93,32 @@ def locked?
...
+        # @param gem_overrides [Hash] one of three keys that contain puppet version
+        def gemfile_definition(gem_overrides = {})
+          gem_overrides ||= {}
+          @gemfile_definition ||= begin
+            env_vars = gemfile_env(gem_overrides)
+            env_vars.each { |name, value| ENV[name.to_s] = value }
+            ::Bundler.configure
+            definition = ::Bundler::Definition.build(gemfile, gemfile_lock, nil)
+            # reset env variable back to nil so we don't cause environment polution
+            env_vars.each { |name, _value| ENV[name.to_s] = nil }
```
scotje: Should this happen in an `ensure` block?

lib/pdk/util/bundler.rb
```diff
@@ -92,18 +93,32 @@ def locked?
...
+        # @example installed?({puppet: '6.10.0'})
+        # @note does not allow you to check arbitrary gems not already in the gemfile
+        def installed?(gem_overrides = {})
+          return false if !File.file?(gemfile) && ::Bundler.frozen_bundle?
 
-          result[:exit_code].zero?
+          definition = gemfile_definition(gem_overrides)
+          definition.validate_runtime!
+          # determines if something is not installed
+          !definition.missing_specs?
```
scotje: Perhaps log at `debug` level the gems that were missing? You can currently see that at `debug` in the raw `bundle` output.

lib/pdk/util/bundler.rb
```diff
@@ -92,18 +93,32 @@ def locked?
...
+        # @return [Boolean] - true if the gemfile with the given overrides is installed
+        # @param gem_overrides [Hash] one of three keys that contain puppet version
+        # @example installed?({puppet: '6.10.0'})
+        # @note does not allow you to check arbitrary gems not already in the gemfile
+        def installed?(gem_overrides = {})
+          return false if !File.file?(gemfile) && ::Bundler.frozen_bundle?
 
-          result[:exit_code].zero?
+          definition = gemfile_definition(gem_overrides)
+          definition.validate_runtime!
```
scotje: May need to rescue some potential bundler exceptions here.

logicminds: In a previous iteration I was rescuing but that is only needed for definition.missing_specs call which is different than definition.missing_specs?

I can revert to the previous method and capture the output for debugging purposes.

scotje: It looks like `validate_runtime!` can raise:

https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L495

https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L506

lib/pdk/util/bundler.rb
logicminds: I'll "ensure" I add that.

logicminds:
> Looking good @logicminds! I added a few comments and it looks like the AppVeyor failure is just that the new version runs in about 14ms on Windows. I think we could safely make the spec something like "under 50ms", what do you think?

I'll update this timing once #774 is merged.   Under 20ms seems unbelievable. I still have my doubts.

logicminds:
After some final testing I don't think this is going to work unless I spend more time than I want to.

The problem is two fold:

Given a pdkized gemfile where the ruby version determines the gem name we would need to pass in the desired ruby version.

```
# Gemfile
#  added below
desired_ruby_version = ENV["BUNDLER_SPEC_RUBY_VERSION"] || RUBY_VERSION
ruby_version_segments = Gem::Version.new(desired_ruby_version).segments
minor_version = ruby_version_segments[0..1].join(".")

group :development do
  gem "fast_gettext", "1.1.0", require: false if Gem::Version.new(desired_ruby_version.dup) < Gem::Version.new("2.1.0")
  gem "fast_gettext", require: false if Gem::Version.new(desired_ruby_version.dup) >= Gem::Version.new("2.1.0")
  gem "json_pure", "<= 2.0.1", require: false if Gem::Version.new(desired_ruby_version.dup) < Gem::Version.new("2.0.0")
  gem "json", "= 1.8.1", require: false if Gem::Version.new(desired_ruby_version.dup) == Gem::Version.new("2.1.9")
  gem "json", "= 2.0.4", require: false if Gem::Requirement.create("~> 2.4.2").satisfied_by?(Gem::Version.new(desired_ruby_version.dup))
  gem "json", "= 2.1.0", require: false if Gem::Requirement.create([">= 2.5.0", "< 2.7.0"]).satisfied_by?(Gem::Version.new(desired_ruby_version.dup))
  gem "rb-readline", "= 0.5.5", require: false, platforms: [:mswin, :mingw, :x64_mingw]
  gem "puppet-module-posix-default-r#{minor_version}", "~> 0.3", require: false, platforms: [:ruby]
  gem "puppet-module-posix-dev-r#{minor_version}", "~> 0.3", require: false, platforms: [:ruby]
  gem "puppet-module-win-default-r#{minor_version}", "~> 0.3", require: false, platforms: [:mswin, :mingw, :x64_mingw]
  gem "puppet-module-win-dev-r#{minor_version}", "~> 0.3", require: false, platforms: [:mswin, :mingw, :x64_mingw]
  gem "puppet-debugger", "~> 0.15", require: false
  gem "puppet-debugger-playbooks", require: false
  #  gem 'ruby-prof', '1.0.0'
end
```

So `desired_ruby_version = ENV["BUNDLER_SPEC_RUBY_VERSION"] || RUBY_VERSION` would need to be added to the pdk-templates.   

Not a big issue but without this addition, pdk users will be given false negatives for installed? every time and thus run bundle install way too often with pdk commands. 

This change can be seen here here: https://github.com/nwops/pdk/blob/bundler/lib/pdk/util/bundler.rb#L102

This appears to solve the [Gemfile eval issue](https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L94).

However, there is another issue that is related to where bundler looks for sources.  I suspect bundler is using RUBY_VERSION to find these.  I am not sure how to fix this, but am optimistic. 

Additionally there is another issue where the Gemfile can specify a ruby version to force.  However, the work around is to set env variable BUNDLER_SPEC_RUBY_VERSION to the desired ruby version.

https://github.com/bundler/bundler/blob/master/lib/bundler/ruby_version.rb#L106

So currently this is failing because we are installed? is returning false forcing bundle install due to a ruby runtime version gem source location.

`[✔] Installing missing Gemfile dependencies.`  

If we could modify RUBY_VERSION at runtime or control where bundler looks for gem sources I think this could be solved.

There is a bunch of 💎💎💎💎💎 waiting in this rabbit hole I dug.  130X speed improvement.

scotje:
@logicminds Do you think there are any incremental pieces of this that we could merge without the bundler changes?

logicminds:
I have been meaning to reach out to bundler support slack channel but have not had the time. 

If we can somehow tell bundler where to look for sources and direct it to the right ruby lib location than that would solve my issue and we could continue forward with fixes above.  

If we want to take a shot at this, we should update the pdk-templates with my changes above to solve the easiest part.   Unless of course we can mock `RUBY_VERSION` with any value we want.

scotje:
OK, that `pdk-templates` change seems simple enough and shouldn't break anyone.

daianamezdrea:
Hi @logicminds! Are you still working at this PR? Also, please, can you rebase it ?

PR #774, opened 2019-10-14
(performance) - Sets up basic performance testing

logicminds:
* Adds a basic performance test so we can start measuring.


sanfrancrisko:
Hi @logicminds, can I ask if there was a particular issue that was the driver behind adding performance tests?

DavidS:
Hi @logicminds ,

thanks for your work here, but most of the startup time we expect is coming from ruby insufficiencies on windows. @jpogran is currently experimenting with [bootsnap](https://github.com/Shopify/bootsnap) on bolt and we'll surely apply our learnings from there here when the time comes.

logicminds:
Curious why a performance test is not wanted here.  How will we know when the software degrades?  Who or what is tracking metrics?

PR #775, opened 2019-10-15
lib/pdk/cli.rb needs to require pdk/version 

Hexoplon:
lib/pdk/cli.rb needs to require pdk/version to fix pdk cli 'uninitialized constant PDK::VERSION' when executing pdk functions like `pdk validate` or `pdk update` from ruby gem install.

### DIFF
--- a/lib/pdk/cli.rb
+++ b/lib/pdk/cli.rb
@@ -115,7 +115,6 @@ module PDK::CLI
     default_subcommand 'help'
 
     flag nil, :version, _('Show version of pdk.') do |_, _|
-      require 'pdk/version'
       puts PDK::Util::Version.version_string
       exit 0
     end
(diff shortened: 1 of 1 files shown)


scotje:
Oops, good catch! I think we would want to add this require at line 118 so that it only gets required if you are asking for the version. (We are trying to reduce the amount of up-front filesystem access before every command.)

Hexoplon:
> Oops, good catch! I think we would want to add this require at line 118 so that it only gets required if you are asking for the version. (We are trying to reduce the amount of up-front filesystem access before every command.)

Tried that, but it did not fix the issue. Adding it at the bottom with the requires for the commands works however.

scotje:
Oh, hmm yes I guess because this is a class definition that doesn't work the same. @rodjek what do you think? Should we just require this at the top here?

glennsarti:
@rodjek Does https://github.com/puppetlabs/pdk/pull/778 supersede this? I suspect it does.

glennsarti:
@Hexoplon Can you try reproducing this problem from master (commit 3687f6c75dd)?

Hexoplon:
@glennsarti This seems to have fixed the issue

glennsarti:
Thanks @Hexoplon !

PR #777, opened 2019-10-15
Update json-schema dependency

smortex:
Version 2.8.0 of the json-schema gem assumed the WIP draft 6 URI would be "http://json-schema.org/draft-06/schema#".  It happens that it is in fact "http://json-schema.org/draft/schema#".

This was fixed in: ruby-json-schema/json-schema#376

json-schema PATCH vesion was bumped, so that version 2.8.1 include this fix.

Instead of depending on a buggy version of this gem and using the same wrong URI, switch to the fixed version of the gem and use the expected URI.


pdk.gemspec
```diff
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'ffi', '~> 1.9.0'
   spec.add_runtime_dependency 'gettext-setup', '~> 0.24'
   spec.add_runtime_dependency 'hitimes', '1.3.0'
-  spec.add_runtime_dependency 'json-schema', '2.8.0'
+  spec.add_runtime_dependency 'json-schema', '~> 2.8.1'
```
smortex: Technically speaking, `~> 2.8` might be more suitable (and improve my life as a packager)…  What do maintainers think about it?

glennsarti:
Thanks for the PR @smortex !!

I'm re-running the CI because those failures don't look related...

As for `~> 2.8.1` vs `~> 2.8`, I would prefer `2.8.1` due to how we package up pdk (We use fixed versions)

Thoughts @rodjek ?

Note - this would need a corresponding bump in pdk-vanagon
https://github.com/puppetlabs/pdk-vanagon/blob/master/configs/components/rubygem-json-schema.rb#L2-L3

glennsarti:
Hrmm...Will have to run this locally to see what's up.

glennsarti:
@smortex Still trying to repro this.  There's something weird going on.  In the meantime, did you want to quickly rebase onto master.

smortex:
I am not able to reproduce the failures on my FreeBSD box (lots of unrelated failures I had not the time to check yet)…  Hope to take some time for this in the next days.  In the meantime I rebased the commit on top of master.  Failures seems to be the very same.

glennsarti:
Oh intersting....

```
Failures:

  1) pdk validate tasks when run inside of a module with valid task Command "pdk validate tasks --format text:stdout --format junit:report.xml" exit_status is expected to eq 0
     Failure/Error: its(:exit_status) { is_expected.to eq(0) }

       expected: 0
            got: 256

       (compared using ==)
       pdk validate tasks --format text:stdout --format junit:report.xml
       pdk (INFO): Using Ruby 2.5.1
pdk (INFO): Using Puppet 6.10.1
[*] Checking task names (tasks/**/*).
[\] Checking task metadata style (tasks/*.json).pdk (FATAL): Unable to validate Task Metadata. Schema not found: http://json-schema.org/draft-06/schema#.

     # ./spec/acceptance/validate_tasks_spec.rb:30:in `block (5 levels) in <top (required)>'
     # ./spec/acceptance/support/with_a_fake_tty.rb:4:in `block (2 levels) in <top (required)>'
```
```

**Comment by glennsarti** [comment]
Ok.... so https://forgeapi.puppet.com/schemas/task.json holds the json schema for Tasks and lo-and-behold:

```
{
  "$schema": "http://json-schema.org/draft-06/schema#",
  "title": "Puppet Task Metadata",
  "description": "The metadata format for Puppet Tasks",
  "type": "object",
  "properties": {
    "description": {
      "type": "string",
...
```

glennsarti:
Looks like we're hitting - https://github.com/ruby-json-schema/json-schema/issues/424

The specification page still lists `http://json-schema.org/draft-06/schema` as the URI

ref - https://json-schema.org/specification-links.html#draft-6

glennsarti:
Bottom line json-schema doesn't _support_ draft 6 (draft 5 and below 😢) and json-schema 2.8.0 just _happens_ to work, whereas 2.8.1 will not.

And due to json-schema, in essence, being abandoned (https://github.com/ruby-json-schema/json-schema/issues/423) I doubt this will ever be addressed.

As much as I want to merge this, I don't think we can 😢

smortex:
Do you think it would be possible to switch to draft 5?  I would like to avoid to create a FreeBSD port for this release of json-schema to depend on and unbreak pdk…

glennsarti:
@scotje  Any chance we can get the JSON schemas "downgraded" to draft 5

I don't believe we're using anything draft 6 specific.

FYI - @rodjek

glennsarti:
Oh no.  There is no draft-05

> Implementors should not implement or advertise support for “draft-05”.

https://json-schema.org/draft-06/json-schema-release-notes.html

smortex:
Yeah, I just saw that…

https://github.com/uktrade/ci-pipeline/commit/abe40db3f3199bf1819fde51aadb838a1ae14f7f

I switched to 4 and the test suite pass on my machine. The acceptance test suite does not pass, but event before it did not pass :-/

glennsarti:
May have to seriously consider moving to https://github.com/davishmcclurg/json_schemer

Only problem is the ruby version restriction (2.4+)  PDK currently supports 2.1+

smortex:
Switching to a previous schema version does not help (but after a second though, the failure you see seems to be caused by a reference to the draft 6 schema from outside of this repository)…  No idea what to do next.

glennsarti:
@smortex , @rodjek and I had a chat about this today. The current plan is that we'll be releasing PDK v2.0.0 very early next year, where we'll be dropping EOL ruby.  That means we could use the (much better) json_schemer gem and avoid all of these draft issues in `json_schema` gem.

In the mean time, I'm going to raise a PR to turn off schema validation but leave the schema code in place and add some failing unit tests e.g. the draft urls should not be "schema.org/draft-06/schema#".

scotje:
Closing in lieu of #793

PR #778, opened 2019-10-16
(maint) PDK as a library

rodjek:
This is in response to a number of issues in the 1.14.0 release where there are some require statements missing as a result of the efforts to cut down on the amount of unnecessary files loaded by PDK.

There are a substantial number of lines changed here, so in summary:
 * `require 'pdk'` is now the main entry point for most files. In this base file, we set up the various accessors that are widely used in the code base (`PDK.config`, `PDK.logger`, `PDK.analytics`, and `PDK.answers`). It also sets up i18n and configures autoloads for all the backend classes and modules.
 * `pdk` is no longer globally required in the spec_helper, instead each spec includes only the file being tested.
 * A new task has been added to the CI configs. This task loads up each spec in its own ruby process to ensure that the classes and modules are tested in isolation. This ensures that users can do something like `require 'pdk/module/build'` and it will require all the necessary files automatically.

```
irb(main):001:0> require 'pdk/module/build'
=> true
irb(main):002:0> $LOADED_FEATURES.grep(/pdk\/lib/)
=> ["/home/tsharpe/code/puppetlabs/pdk/lib/pdk/version.rb", "/home/tsharpe/code/puppetlabs/pdk/lib/pdk/i18n.rb", "/home/tsharpe/code/puppetlabs/pdk/lib/pdk.rb", "/home/tsharpe/code/puppetlabs/pdk/lib/pdk/module.rb", "/home/tsharpe/code/puppetlabs/pdk/lib/pdk/module/build.rb"]
```

Excluding `lib/pdk/version.rb` (which is loaded by the gemspec which is evaluated by Bundler in the dev environment), the above example shows that even though `lib/pdk/module/build.rb` requires `pdk`, only the necessary files are loaded.

**Does this mean we should get rid of all the requires in the methods?**
No, it's still faster to explicitly require the files you need rather than rely on the autoloader. The autoloader provides a fallback method that allows ruby to look up and find any undefined constants that might have been missed.


logicminds:
I discovered a few of these issues myself testing the gem directly without running unit tests.  The unit tests do mask a number of loading issues so I am glad to see this.  I would like to see some documentation for how to best require a file in a unit test and when building commands.

logicminds:
Then there is also https://github.com/fxn/zeitwerk too.  Never used this.  But maybe it should be considered.

glennsarti:
Zeitwerk requires ruby 2.4.4+.  Until we drop 2.1.9 support, no zeitwerk love :-(

PR #779, opened 2019-10-17
(PDK-1525) Use less restrictive minitar version

glennsarti:
Previously the PDK used the minitar restriction similar to that of the Puppet
Windows build (~> 0.6.1).  However, from Puppet 6.10.1 this version pin
changed to `~> 0.9`. This means that using PDK as a gem on Windows, is causing
bundler to choose old versions of gems (e.g. doesn't include Litmus).

This commit changes the minitar pin to still be compatible with older Puppet
gems and still allow later Puppet too.

### DIFF
--- a/pdk.gemspec
+++ b/pdk.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'hitimes', '1.3.0'
   spec.add_runtime_dependency 'json-schema', '2.8.0'
   spec.add_runtime_dependency 'json_pure', '~> 2.1.0'
-  spec.add_runtime_dependency 'minitar', '~> 0.6.1'
+  spec.add_runtime_dependency 'minitar', '~> 0.6'
   spec.add_runtime_dependency 'pathspec', '~> 0.2.1'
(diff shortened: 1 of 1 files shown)


glennsarti:
This won't require pdk-vanagon changes.  This is only applicable when using PDK as a gem.

glennsarti:
Using puppetlabs-dsc as test (from the original ticket)

``` text
Resolving dependencies..........
Using rake 12.3.3
Using CFPropertyList 2.3.6
Using concurrent-ruby 1.1.5
...
Using puppet 6.10.1 (x64-mingw32)   <---- latest Puppet agent
...
Using multi_json 1.14.1
Using logging 2.2.2
Using minitar 0.9   <------ New minitar
Using net-ssh 5.2.0
...
Using pdk 1.15.0.pre from source at `C:/Source/pdk`  <--- using this PR code
Using powerpack 0.1.2
Using slop 3.6.0
...
Using puppet-module-win-dev-r2.5 0.4.2   <---  Using the correct version of the metagem
Bundle complete! 18 Gemfile dependencies, 143 gems now installed.
Gems in the groups system_tests, build and build,system_tests were not installed.
Bundled gems are installed into `./.bundle/windows`
```

PR #780, opened 2019-10-17
(MAINT) Fix new provider error message

michaeltlombardi:
Prior to this commit, if a user creates a module with
the PDK, then tries to create a provider, they get an
error - but the error leads to an anchor on the docs
for the resource_api readme which does not exist.

This commit corrects the URL of the error message,
pointing the user to the actual instructions they need
to follow to move forward.


DavidS:
This also needs fixing in https://github.com/puppetlabs/pdk/blob/9dd3176835e40feada2357145852e28c4d6873e3/lib/pdk/generate/transport.rb#L24

PR #782, opened 2019-10-18
(maint) Raise error when template content is empty

logicminds:
* Throws and error if the content is nil.  This fixes a problem
    when a template cannot be rendered correctly due to some filesytem
    issue.

    This was originally seen when config_defaults.yml had a
      delte: true set for any of the files.

    The error message previously thrown was confusing so this is meant
    to throw a better error message.

### DIFF
--- a/lib/pdk/util/filesystem.rb
+++ b/lib/pdk/util/filesystem.rb
@@ -2,6 +2,7 @@ module PDK
   module Util
     module Filesystem
       def write_file(path, content)
+        raise ArgumentError, "#{path} is empty" unless content
         raise ArgumentError unless path.is_a?(String) || path.respond_to?(:to_path)
 
         # Harmonize newlines across platforms.


lib/pdk/util/filesystem.rb
```diff
@@ -2,6 +2,7 @@ module PDK
   module Util
     module Filesystem
       def write_file(path, content)
+        raise ArgumentError, "#{path} is empty" unless content
```
glennsarti: I would probably go a step further here. `content` is expected to be a string due to the `...encode` command later.

An empty string would be a valid content, but a `nil` value would not.

glennsarti: Also this should be using the `gettext` format e.g.

```
raise PDK::CLI::ExitWithError, _(
          '%{message} Rename the file or exclude it from the package ' \
          'by adding it to the .pdkignore file in your module.',
        ) % { message: e.message }
```

rodjek:
See #794

PR #787, opened 2019-10-23
Loosening of minitar dependency

cardil:
Puppet has moved to using minitar `~> 0.9`, but pdk and bolt still have minitar as `~> 0.6`. That's a problem as users can't use them all in module testing.

Loosening constraint to accept any version before 1.0

```
Resolving dependencies...
Bundler could not find compatible versions for gem "minitar":
  In snapshot (Gemfile.lock):
    minitar (= 0.6.1)
  In Gemfile:
puppet-module-posix-dev-r2.5 (~> 0.3) was resolved to 0.4.2, which depends
on
      puppet_litmus (>= 0.4.0, < 1.0.0) was resolved to 0.12.0, which depends on
        pdk (>= 1.10.0, < 2.0.0) was resolved to 1.14.0, which depends on
          minitar (~> 0.6.1)
puppet-module-posix-dev-r2.5 (~> 0.3) was resolved to 0.4.2, which depends
on
      puppet_pot_generator (~> 1.0) was resolved to 1.0.1, which depends on
        puppet was resolved to 6.10.1, which depends on
          minitar (~> 0.9) x64-mingw32
Running `bundle update` will rebuild your snapshot from scratch, using only
```
Ref: https://ci.appveyor.com/project/cardil/puppet-passless/builds/28325323/job/p2dihyb39jwp9hil

### DIFF
--- a/pdk.gemspec
+++ b/pdk.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'hitimes', '1.3.0'
   spec.add_runtime_dependency 'json-schema', '2.8.0'
   spec.add_runtime_dependency 'json_pure', '~> 2.1.0'
-  spec.add_runtime_dependency 'minitar', '~> 0.6'
+  spec.add_runtime_dependency 'minitar', '>= 0.6', '< 1.0'
   spec.add_runtime_dependency 'pathspec', '~> 0.2.1'
   spec.add_runtime_dependency 'tty-prompt', '~> 0.13'
   spec.add_runtime_dependency 'tty-spinner', '~> 0.5'


joshcooper:
I think this issue was fixed in https://github.com/puppetlabs/pdk/commit/511d8156fb0cb7d81ac6b8356797a4497d43c456. The issue was pdk was overly restrictive in its `minitar` dependency, which it copied from puppet, and then was broken when puppet moved to `~> 0.9`

cardil:
> I think this issue was fixed in [511d815](https://github.com/puppetlabs/pdk/commit/511d8156fb0cb7d81ac6b8356797a4497d43c456). The issue was pdk was overly restrictive in its `minitar` dependency, which it copied from puppet, and then was broken when puppet moved to `~> 0.9`

Of course. I shouldn't be coding so late. I see that's isn't released yet. That's what cached me! Sorry!

PR #790, opened 2019-10-25
(PDK-1443) Windows safe Tempfiles & environment variable access

rodjek:


lib/pdk/cli/exec/command.rb
```diff
@@ -19,8 +21,8 @@ def initialize(*argv)
           @process = ChildProcess.build(*@argv)
           @process.leader = true
 
-          @stdout = Tempfile.new('stdout').tap { |io| io.sync = true }
-          @stderr = Tempfile.new('stderr').tap { |io| io.sync = true }
+          @stdout = Tempfile.new('stdout', mode: TEMPFILE_MODE).tap { |io| io.sync = true }
```
glennsarti: Confirmed this is the only place Tempfile is used in the PDK source.  There are many uses in spec tests but that's fine.

PR #791, opened 2019-10-28
(#773) Respect --verbose in interactive pdk test unit

### LINKED ISSUE #773: 'pdk test unit --verbose' does not change the rspec output format
**Describe the bug**
As per [the original feature request for making pdk test unit verbose by default](https://tickets.puppetlabs.com/browse/PDK-680?focusedCommentId=678885&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-678885), running `pdk test unit --verbose` should change rspec's output format from `progress` to `documentation`.

But in PDK 1.14.0, the `--verbose` flag doesn't do anything.

**To Reproduce**
1. Install PDK 1.14.0
1. Run `pdk test unit` and `pdk test unit --verbose`. Observe that the output doesn't change.

**Expected behavior**
1. `pdk test unit` uses rspec's `progress` format.
1. `pdk test unit --verbose` uses rspec's `documentation` format.

**Additional context**
- PDK 1.14.0 installed via puppet.com DMG

**Further Thoughts**

I'm realizing that this all will probably conflict with what's in the module's `./.rspec` file. The format can be defined there and I don't know if that should override or be overridden by the `pdk` command.

The standard PDK template lays down `--format documentation`:
https://github.com/puppetlabs/pdk-templates/blob/master/moduleroot/.rspec.erb#L2

rodjek:
`pdk test unit` will now use the `progress` formatter (dots) and `pdk
test unit --verbose` will now use the `documentation` formatter.

Fixes #773

### DIFF
--- a/spec/unit/pdk/test/unit_spec.rb
+++ b/spec/unit/pdk/test/unit_spec.rb
@@ -292,13 +292,16 @@ describe PDK::Test::Unit do
 
     context 'when running interactively' do
       subject(:exit_code) do
-        described_class.invoke(report, interactive: true)
+        described_class.invoke(report, options)
       end
 
       before(:each) do
         allow(PDK::CLI::Exec::InteractiveCommand).to receive(:new)
           .and_return(command)
+        allow(command).to receive(:execute!).and_return(exit_code: 0)
       end
+
+      let(:options) { { interactive: true } }
 
       let(:command) do
         instance_double(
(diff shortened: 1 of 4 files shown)


spec/unit/pdk/test/unit_spec.rb
```diff
@@ -309,9 +312,25 @@
       end
 
       it 'runs the rake task interactively' do
-        expect(command).to receive(:execute!).and_return(exit_code: 0)
```
glennsarti: So this is just removing a duplicate expectation (same as the line below?)

PR #793, opened 2019-10-28
(PDK-1530) Disable schema validation of config files

glennsarti:
Reference - #777 

Previously the PDK Analytics configuration file had a JSON schema enabled,
however the json-schema gem is causing issues.  This commit disables the schema
validation, while retaining the schema code.  Later commits will use the updated
json_schemer gem once EOL Rubies support is dropped.

### DIFF
--- a/lib/pdk/config.rb
+++ b/lib/pdk/config.rb
@@ -8,19 +8,26 @@ module PDK
     autoload :LoadError, 'pdk/config/errors'
     autoload :Namespace, 'pdk/config/namespace'
     autoload :Setting, 'pdk/config/setting'
+    autoload :Validator, 'pdk/config/validator'
     autoload :YAML, 'pdk/config/yaml'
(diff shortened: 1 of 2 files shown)


lib/pdk/config.rb
```diff
@@ -3,24 +3,29 @@
 module PDK
   class Config
     autoload :JSON, 'pdk/config/json'
-    autoload :JSONSchemaNamespace, 'pdk/config/json_schema_namespace'
-    autoload :JSONSchemaSetting, 'pdk/config/json_schema_setting'
```
rodjek: I would suggest keeping these autoloads in place rather than adding the top level requires as done below

glennsarti: +1. Will do

glennsarti: Done.

PR #799, opened 2019-11-07
(PDK-1523) Refactor filesystem operations to use PDK::Util::Filesystem

rodjek:
Refactor all filesystem operations to use the `PDK::Util::Filesystem` wrapper module. RuboCop cops with autocorrect have been added for all the methods to ensure that there is no future drift (`File.open` is the only cop without autocorrect as it requires a human to examine the intent of the block).


spec/unit/pdk/module/update_manager_spec.rb
```diff
@@ -98,14 +98,14 @@
           end
 
           it 'removes the file' do
-            expect(FileUtils).to receive(:rm).with(dummy_file)
```
glennsarti: Hrmm,  this commit is a bit odd.  Only tests were modified, nothing actually in `/lib`

lib/pdk/report.rb
```diff
@@ -94,22 +96,26 @@ def write_junit(target = self.class.default_target)
...
         tool_events.each do |event|
           if event.rspec_puppet_coverage?
             coverage_report = event.to_text
           else
-            target.puts(event.to_text) unless event.pass?
+            report << event.to_text unless event.pass?
           end
         end
       end
-    ensure
```
glennsarti: Shouldn't this ensure still be here?

rodjek: Nope, that ensure was only there to close the file, but that's handled by `PDK::Util::Filesystem.write_file` now.

spec/unit/pdk/module/update_manager_spec.rb
rodjek: Yeah, some of the methods had already been updated to use `PDK::Util::Filesystem` but the tests weren't updated at the time.

PR #802, opened 2019-11-14
(PDK-1169) Add VMWare fallback to PDK::Util::Filesystem.mv

rodjek:
`File.rename` (used inside of `FileUtils.mv`) raises `Errno::ENOENT`
when used inside of a VMWare shared folder on Windows. This adds
a fallback to `PDK::Util::Filesystem.mv` that first checks if the source
file exists (in case the error is legitimate) and then "moves" the file
using copy and remove operations.

As this is quite the corner case, there are no tests added but I've
tested it on a local VM and it resolves the issue.

### DIFF
--- a/lib/pdk/util/filesystem.rb
+++ b/lib/pdk/util/filesystem.rb
@@ -116,6 +116,20 @@ module PDK
 
       def mv(*args)
         FileUtils.mv(*args)
+      rescue Errno::ENOENT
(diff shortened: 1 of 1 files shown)


glennsarti:
@rodjek This looks like a manifestation of https://github.com/jruby/jruby/pull/5168

glennsarti:
i.e. can't rename across filesystems

PR #803, opened 2019-11-14
(PDK-1364) Allow non-git template directories to be used

glennsarti:
Previously the PDK could only use git based template directories which made
testing new templates difficult and non-intuitive.  This commit makes the
following changes which enables plain file system based directories to be used
as a PDK template during operations like PDK update and PDK convert:

* Changes TemplateURI so that is now a pure URI.  Previously many git type
  methods and functions where also mixed in, which muddied what the class
  actually did.
  - The Git functionality is moved to the PDK::Util::Git module
  - git_remote is now bare_uri (A URI without a fragment)
  - git_ref is now uri_fragment

* Moves the default template url methods in TemplateURI from class instance to
  module instance. This stops the expensive TemplateDir class being instantiated
  more than once

* Move the TemplateDir validate_module_template! from a class instance to a
  module instance method.  This means that we do not have to create the very
  expensive TemplateDir object, and can just quickly validate a local path as a
  template directory.

* Splits out the default default_template_uri method to default_template_uri and
  default_template_addressable_uri. This stops needless creations of TemplateURI
  objects.

* The default default_template_uri method no longer takes a ref parameter, as it
  should only return a URI, not a URI and ref.

* The TemplateURI now has no knowledge of _what_ it's pointing to. The
  TemplateDir class has been updated to detect what kind of Template the URI is
  pointing to.  This kind is then stored in the new template_dir_type variable
  which then modifies the behavior of the class. For example, the metadata now
  outputs the template-ref for Git based templates as it makes no sense for
  a non-git template directory.


lib/pdk/module/templatedir.rb
```diff
@@ -6,6 +6,11 @@ class TemplateDir
       attr_accessor :module_metadata
       attr_reader :uri
 
+      attr_reader :template_dir_type
+
+      TEMPLATE_DIR_FILESYSTEM_TYPE ||= :filesystem
```
glennsarti: Move these to the module, and then no need for `||=`

lib/pdk/module/templatedir.rb
```diff
@@ -91,17 +99,26 @@ def initialize(uri, module_metadata = {}, init = false)
...
+        result = { 'pdk-version' => PDK::Util::Version.version_string }
+        case @template_dir_type
+        when TEMPLATE_DIR_GIT_TYPE
+          result['template-url'] = uri.metadata_format
+          result['template-ref'] = cache_template_ref(@path)
+        when TEMPLATE_DIR_FILESYSTEM_TYPE
+          result['template-url'] = uri.bare_uri
+          result['template-ref'] = nil
+        else
+          raise PDK::CLI::FatalError, _('Unkown template directory type %{dir_type}') % { dir_type: @template_dir_type }
```
rodjek: s/Unkown/Unknown/

lib/pdk/module/templatedir.rb
rodjek: Shouldn't be needed in a class either as we're not (or shouldn't be) reloading the file and causing them to be redefined.

glennsarti: Done

lib/pdk/module/templatedir.rb
glennsarti: Done

lib/pdk/module/template_dir.rb
```diff
@@ -0,0 +1,115 @@
...
+      #   end
+      #
+      # @raise [ArgumentError] If no block is given to this method.
+      # @raise [PDK::CLI::FatalError] (see #clone_repo)
+      # @raise [ArgumentError] (see #validate_module_template!)
+      #
+      # @api public
+      def self.with(uri, module_metadata = {}, init = false)
+        unless block_given?
+          raise ArgumentError, _('%{class_name} must be initialized with a block.') % { class_name: self.class.name }
```
rodjek:
```suggestion
          raise ArgumentError, _('%{class_name}.with must be passed a block.') % { class_name: self.class.name }
```

lib/pdk/module/template_dir.rb
```diff
@@ -0,0 +1,115 @@
...
+        end
+        unless uri.is_a? PDK::Util::TemplateURI
+          raise ArgumentError, _('PDK::Module::TemplateDir.for must be initialized with a PDK::Util::TemplateURI, got a %{uri_type}') % { uri_type: uri.class }
```
rodjek:
```suggestion
          raise ArgumentError, _('%{class_name}.with must be passed a PDK::Util::TemplateURI, got a %{uri_type}') % { uri_type: uri.class, class_name: self.class.name }
```

lib/pdk/module/template_dir/base.rb
```diff
@@ -0,0 +1,268 @@
...
+        # @raise [PDK::CLI::FatalError] (see #clone_repo)
+        # @raise [ArgumentError] (see #validate_module_template!)
+        #
+        # @api public
+        def initialize(uri, module_metadata = {}, init = false)
+          unless block_given?
+            raise ArgumentError, _('%{class_name} must be initialized with a block.') % { class_name: self.class.name }
+          end
+          unless uri.is_a? PDK::Util::TemplateURI
+            raise ArgumentError, _('PDK::Module::TemplateDir::Base.new must be initialized with a PDK::Util::TemplateURI, got a %{uri_type}') % { uri_type: uri.class }
```
rodjek:
```suggestion
            raise ArgumentError, _('%{class_name} must be initialized with a PDK::Util::TemplateURI, got a %{uri_type}') % { uri_type: uri.class, class_name: self.class.name }
```

spec/unit/pdk/generate/module_spec.rb
```diff
@@ -29,11 +29,13 @@
 end
 
 shared_context 'mock template dir' do
-  let(:test_template_dir) { instance_double(PDK::Module::TemplateDir, metadata: {}) }
+  require 'pdk/module/template_dir/base'
```
rodjek: Rather than adding these requires into the test, it would be better to add `autoload`s for the subclasses at the top of `lib/pdk/module/template_dir.rb`

glennsarti: Maybe? Wasn't sure it was worth it as it's only in the spec.

lib/pdk/module/template_dir.rb
glennsarti: Done

lib/pdk/module/template_dir.rb
glennsarti: Done

lib/pdk/module/template_dir/base.rb
glennsarti: Done

lib/pdk/util/git.rb
```diff
@@ -81,10 +90,12 @@ def self.remote_repo?(maybe_repo)
 
       def self.work_tree?(path)
         return false unless PDK::Util::Filesystem.directory?(path)
+        result = cached_git_query(path, :work_tree?)
+        return result unless result.nil?
```
scotje: Is it too much optimization to check the cache before hitting the filesystem with `.directory?` here?

lib/pdk/util/git.rb
```diff
@@ -19,6 +19,8 @@ def initialze(args, result)
     end
 
     module Git
+      GIT_QUERY_CACHE_TTL ||= 10
```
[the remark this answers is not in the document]
glennsarti: The use case I was worried about was:

1. Check if a directory, which doesn't exist, is a git repo  (e.g. /foo)
2. Then git clone into that directory (git clone ...uri...  /foo)

Note that `.clone` isn't a method in the Git module.

foo is _NOW_ a valid git repo.

Yeah I'm not 100% happy with using an "arbitrary" time out either, but I couldn't figure out any other way to invalidate the cache.  There are _significant_ improvements with caching though.

10s was, in my mind, long enough to provide a performance benefit but slow enough so that it would be meaningful.

Open to other ideas on how to do this 👍

lib/pdk/util/git.rb
glennsarti: I felt it was. and as per the workflow above, if the directory is _later_ created then we'd get false cache hits.

lib/pdk/module/template_dir.rb
```diff
@@ -0,0 +1,115 @@
...
+      # @return [void]
+      #
+      # @api public
+      def self.validate_module_template!(template_root_dir)
+        # rubocop:disable Style/GuardClause
+        unless PDK::Util::Filesystem.directory?(template_root_dir)
+          require 'pdk/util'
+
+          if PDK::Util.package_install? && PDK::Util::Filesystem.fnmatch?(File.join(PDK::Util.package_cachedir, '*'), template_root_dir)
+            raise ArgumentError, _('The built-in template has substantially changed. Please run "pdk convert" on your module to continue.')
```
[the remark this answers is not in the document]
glennsarti: Unknown. 🤷‍♂️  This is a file rename from https://github.com/puppetlabs/pdk/blob/61bb2fc6b2/lib/pdk/module/templatedir.rb#L215-L219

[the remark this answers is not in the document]
scotje: Gotcha, well we won't worry about it right now then. :)

lib/pdk/util/git.rb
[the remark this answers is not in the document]
scotje: I see.

I would say that down the line it might be good to route all git operations through this module and then the methods that impact the validity of the cache can just handle the invalidation internally as part of like `PDK::Util::Git.clone` or whatever. Perhaps we should just add a maintenance ticket to centralize all Git operations like we did for filesystem stuff?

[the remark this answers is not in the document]
glennsarti: Raised https://tickets.puppetlabs.com/browse/PDK-1551 to track the work

PR #807, opened 2019-11-22
(#806) Use ASCII quotes instead of Unicode quotes

### LINKED ISSUE #806: (UI) Try not to use non-ASCII characters where possible
When running `pdk module build` I receive the following

```
C:\Source\tmp\pdk_release\puppetlabs-inifile [master ≡ +2 ~3 -0 !]> pdk module build
pdk (WARN): Modules are built using the â?~pdk buildâ?T command.                                  
```

Note the funny looking characters `â?~` and `â?T`

This is due to the "smart quotes" in https://github.com/puppetlabs/pdk/blob/c9a7ae5b775c1b29ba22c04145cb519598363044/lib/pdk/cli/module/build.rb#L8

We should probably do an audit and get rid of "smart quotes" as they don't render in some terminals

Windows 10 - 1803 - powershell.exe
PDK 1.14.1

rodjek:
Added a custom RuboCop cop to detect any smart quotes that sneak in in the future. It can automatically fix smart single quotes, fixing double quotes can be implemented if needed in the future.

Fixes #806

### DIFF
--- a/ext/rubocop.rb
+++ b/ext/rubocop.rb
@@ -777,6 +777,48 @@ module RuboCop
           end
         end
       end
+
+      class SmartQuotes < Cop
+        MSG = 'Use ASCII quotes instead of Unicode smart quotes'.freeze
+        SINGLE_QUOTES_PAT = %r{(?:\u2018|\u2019)}
+        DOUBLE_QUOTES_PAT = %r{(?:\u201C|\u201D)}
+
+        def on_str(node)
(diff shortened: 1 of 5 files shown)


glennsarti:
I'm kinda surprised there isn't a rubocop plugin for smart-quotes to be honest

rodjek:
Yeah, same. I had a search around on rubygems.org but couldn't find one :man_shrugging:

Issue #808, opened 2019-11-25
Implement `pdk release` command as per RFC (part 1)


glennsarti:
The issue will track the work to implement the `pdk release` subcommand as defined in [RFC #003](https://github.com/puppetlabs/pdk-planning/blob/master/RFCs/0003-add-pdk-release.md)

glennsarti:
Part 1 has been completed.  Closing the ticket.  #815 will track the remainder of the work.

PR #809, opened 2019-11-25
(GH-808) Implement pdk release subcommand

glennsarti:
RFC 003 [1] defines the experience and behavior of the new pdk release sub
command.  This commit:

* Adds the root subcommand `release` with all of the available skip options.
  This will allow the later `release build` and so on sub-sub-commands to be
  implemented much easier. Future commits will add these sub-sub commands.

* Adds a PDK::Module::Release class which does all of the release process

* Adds tests for Release process

[1] https://github.com/puppetlabs/pdk-planning/blob/master/RFCs/0003-add-pdk-release.md

### DIFF
--- a/lib/pdk/cli/release.rb
+++ b/lib/pdk/cli/release.rb
@@ -1,0 +1,179 @@
+require 'pdk/cli/util'
+require 'pdk/validate'
+require 'pdk/util/bundler'
+require 'pdk/cli/util/interview'
+require 'pdk/util/changelog_generator'
+require 'pdk/module/build'
+
+module PDK::CLI
+  @release_cmd = @base_cmd.define_command do
+    name 'release'
+    usage _('release [options]')
+    summary _('(Experimental) Release a module to the Puppet Forge.')
+
+    flag nil, :force,                _('Release the module automatically, with no prompts.')
+    flag nil, :'skip-validation',    _('Skips the module validation check.')
+    flag nil, :'skip-changelog',     _('Skips the automatic changelog generation.')
+    flag nil, :'skip-dependency',    _('Skips the module dependency check.')
+    flag nil, :'skip-documentation', _('Skips the documentation update.')
+    flag nil, :'skip-build',         _('Skips module build.')
+    flag nil, :'skip-publish',       _('Skips publishing the module to the forge.')
+
+    option nil, :'forge-upload-url', _('Set forge upload url path.'),
+           argument: :required, default: 'https://forgeapi.puppetlabs.com/v3/releases'
+
+    option nil, :'forge-token', _('Set Forge API token.'), argument: :required, default: nil
+
+    option nil, :version, _('Update the module to the specified version prior to release. When not specified, the new version will be computed from the Changelog where possible.'),
+           argument: :required
+
+    option nil, :file, _('Path to the built module to push to the Forge. This option can only be used when --skip-build is also used. Defaults to pkg/<module version>.tar.gz'),
+           argument: :required
+
+    run do |opts, _args, _cmd|
+      # Make sure build is being run in a valid module directory with a metadata.json
+      PDK::CLI::Util.ensure_in_module!(
+        message:   _('`pdk release` can only be run from inside a valid module with a metadata.json.'),
+        log_level: :info,
+      )
+
+      unless opts[:force]
+        Release.prepare_interview(opts)
+      end
+
+      # Don't pass tokens to analytics
+      PDK::CLI::Util.analytics_screen_view('release', opts.reject { |k, _| k == :'forge-token' })
+
+      release = PDK::Module::Release.new(nil, opts)
+
+      unless release.module_metadata.forge_ready?
+        if opts[:force]
+          PDK.logger.warn _(
+            'This module is missing the following fields in the metadata.json: %{fields}. ' \
+            'These missing fields may affect the visibility of the module on the Forge.',
+          ) % {
+            fields: release.module_metadata.missing_fields.join(', '),
+          }
+        else
+          release.module_metadata.interview_for_forge!
+          release.write_module_metadata!
+        end
+      end
+
+      unless release.pdk_compatible?
+        if opts[:force]
+          PDK.logger.warn _('This module is not compatible with PDK, so PDK can not validate or test this build.')
+        else
+          PDK.logger.info _('This module is not compatible with PDK, so PDK can not validate or test this build. ' \
+                            'Unvalidated modules may have errors when uploading to the Forge. ' \
+                            'To make this module PDK compatible and use validate features, cancel the build and run `pdk convert`.')
+          unless PDK::CLI::Util.prompt_for_yes(_('Continue build without converting?'))
+            PDK.logger.info _('Build cancelled; exiting.')
+            PDK::Util.exit_process(1)
+          end
+        end
+      end
+
+      release.run
+    end
+  end
+
+  module Release
+    def self.prepare_interview(opts)
+      questions = []
+
+      unless opts[:'skip-validation']
+        questions << {
+          name:     'validation',
+          question: _('Do you want to run the module validation ?'),
+          type:     :yes,
+        }
+      end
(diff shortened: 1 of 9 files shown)


glennsarti:
Note that this does not implement the sub-sub commands e.g. `pdk release build`.  That will come in later PRs.  This is quite a big PR already so no need to add more.

glennsarti:
Marked this as `(Experimental)` as well.

lib/pdk/module/release.rb
```diff
@@ -0,0 +1,260 @@
...
+        validate_publish_options!
+
+        run_validations(options) unless skip_validation?
+
+        PDK.logger.info _('Releasing %{module_name} - from version %{module_version}') % {
+          module_name:    module_metadata.data['name'],
+          module_version: module_metadata.data['version'],
+        }
+
+        PDK::Util::ChangelogGenerator.generate_changelog unless skip_changelog?
```
rodjek: If you move this below the conditional on lines 57-68, could you then remove the duplicate call at the end of that conditional (i.e. rather than generate the changelog and update the version number and then regenerate, just generate the changelog after possibly updating the version number)?

lib/pdk/util/changelog_generator.rb
```diff
@@ -0,0 +1,113 @@
...
+        elsif data =~ %r{^### Fixed}
+          # Patch Version bump
+          version[2] += 1
+        end
+
+        version.join('.')
+      end
+
+      def self.changelog_file
+        @changelog_file ||= PDK::Util::Filesystem.expand_path('CHANGELOG.MD')
```
rodjek: Probably need to look for `CHANGELOG.md` as well for case sensitive filesystems as it's pretty common to use the lowercase extension (or maybe use `fnmatch` with casefold)

glennsarti: Looks like the default is `CHANGELOG.md` so 👍 

https://github.com/github-changelog-generator/github-changelog-generator/blob/73f7a9a187c1f2d6a62b290cfcb984aacbbc45f8/lib/github_changelog_generator/parser.rb#L52

lib/pdk/module/release.rb
glennsarti: Unfortunately we can't (right now).  One workflow is:

* Generate the changelog with the existing version number
* Compute the next version by parsing the changelog looking for Major, Minor, Patch changes 
* Bump the metadata with new version
* Regenerate the Changelog with the now "correct" version number

Unfortunately, due to the way the task is being called (via a Rake task in the module's context) I can't just generate the changelog "on the fly" (e.g. output to a temp file, or capture on STDOUT).  Ideally it would be:

* Generate the current changelog and output to STDOUT (don't save)
* Parse the changelog ....
...
* Generate the changelog and save to CHANGELOG.md

glennsarti: If the changelog gem was in the PDK's context (not the modules), perhaps we'd have more control over it? and do smarter things.

lib/pdk/util/changelog_generator.rb
glennsarti: We're at the mercy of the changelog generator rake task here :-(

glennsarti: Fixed

lib/pdk/cli/release.rb
```diff
@@ -0,0 +1,179 @@
...
+require 'pdk/util/changelog_generator'
+require 'pdk/module/build'
+
+module PDK::CLI
+  @release_cmd = @base_cmd.define_command do
+    name 'release'
+    usage _('release [options]')
+    summary _('(Experimental) Release a module to the Puppet Forge.')
+
+    flag nil, 'force',               _('Release the module automatically, with no prompts.')
```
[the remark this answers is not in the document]
glennsarti: Fixed

lib/pdk/cli/release.rb
```diff
@@ -0,0 +1,179 @@
...
+
+    option nil, :'forge-upload-url', _('Set forge upload url path.'),
+           argument: :required, default: 'https://forgeapi.puppetlabs.com/v3/releases'
+
+    option nil, :'forge-token', _('Set Forge API token.'), argument: :required, default: nil
+
+    option nil, 'version', _('Update the module to the specified version prior to release. When not specified, the new version will be computed from the Changelog where possible.'),
+           argument: :required
+
+    option nil, 'file', _('Path to the built module to push to the Forge. This option can only be used when --skip-build is also used. Defaults to pkg/<module version>.tar.gz'),
```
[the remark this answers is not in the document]
glennsarti: Fixed

lib/pdk/cli/release.rb
```diff
@@ -0,0 +1,179 @@
...
+    flag nil, :'skip-documentation', _('Skips the documentation update.')
+    flag nil, :'skip-build',         _('Skips module build.')
+    flag nil, :'skip-publish',       _('Skips publishing the module to the forge.')
+
+    option nil, :'forge-upload-url', _('Set forge upload url path.'),
+           argument: :required, default: 'https://forgeapi.puppetlabs.com/v3/releases'
+
+    option nil, :'forge-token', _('Set Forge API token.'), argument: :required, default: nil
+
+    option nil, 'version', _('Update the module to the specified version prior to release. When not specified, the new version will be computed from the Changelog where possible.'),
```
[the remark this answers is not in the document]
glennsarti: Fixed

lib/pdk/module/release.rb
```diff
@@ -0,0 +1,260 @@
...
+      attr_reader :options
+
+      attr_reader :module_path
+
+      def initialize(module_path, options = {})
+        @options = options
+
+        # TODO: Currently the release process can ONLY be run if the working directory IS the module root. However, in the future
+        # this WILL change, so we have the API arguments for it, but only accept `nil` for the first parameter
+        raise _('Running the release process outside of the working directory is not supported') unless module_path.nil?
```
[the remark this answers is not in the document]
glennsarti: Fixed this and a bunch of other places I was using plain `raise`

lib/pdk/module/release.rb
```diff
@@ -0,0 +1,260 @@
...
+      end
+
+      def default_package_filename
+        return @default_tarball_filename unless @default_tarball_filename.nil?
+        builder = PDK::Module::Build.new(module_dir: module_path)
+        @default_tarball_filename = builder.package_file
+      end
+
+      def run_validations(opts)
+        # TODO: Surely I can use a pre-existing class for this?
```
rodjek: I assume the plan is to come back and dedupe this & the `pdk validate` CLI in a follow up PR?

glennsarti: Hasn't been ticketted yet. Need to do that first.  Seemed out of scope for this PR.

lib/pdk/module/release.rb
```diff
@@ -0,0 +1,260 @@
...
+        PDK::Util::ChangelogGenerator.generate_changelog unless skip_changelog?
+
+        # Calculate the new module version
+        new_version = specified_version
+        if new_version.nil? && !skip_changelog?
+          new_version = PDK::Util::ChangelogGenerator.compute_next_version(module_metadata.data['version'])
+        end
+        new_version = module_metadata.data['version'] if new_version.nil?
+
+        if new_version != module_metadata.data['version']
```
rnelson0: Could this be tied to a `--force` type flag? It seems more likely that someone typos the version (1.0.1 instead of 1.1.0) and automatically changing it without specific acknowledgement of the difference could lead to undesired results.

glennsarti: There is an implicit `force` in play here.  The user _has_ to specify `--version <...>` on the command line in order for that to happen.  Adding a requirement for `--force` as well seems unnecessary as this 2nd command line argument is not intuitive and gives no extra safety.

lib/pdk/module/release.rb
glennsarti: Raised https://tickets.puppetlabs.com/browse/PDK-1552

PR #813, opened 2019-12-05
(GH-808) Implement pdk release prep and publish subcommands

glennsarti:
Previously the pdk release command was added.  This commit adds the prep and
publish subcommands as per PDK RFC 003.

This commit also adds tests for new subcommands.

The `pdk release build` sub command needs more work, specifically around the tagging component.

### DIFF
--- a/spec/unit/pdk/cli/release/prep_spec.rb
+++ b/spec/unit/pdk/cli/release/prep_spec.rb
@@ -1,0 +1,71 @@
+require 'spec_helper'
+require 'pdk/cli'
+
+describe 'PDK::CLI release prep' do
+  let(:help_text) { a_string_matching(%r{^USAGE\s+pdk release prep}m) }
+  let(:cli_args) { %w[release prep] }
+
+  context 'when not run from inside a module' do
+    include_context 'run outside module'
+
+    it 'exits with an error' do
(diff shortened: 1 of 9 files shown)


spec/unit/pdk/cli/release/prep_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    it 'skips building and publishing' do
+      expect(PDK::Module::Release).to receive(:new).with(Object, hash_including(:'skip-build' => true, :'skip-publish' => true))
+
+      expect { PDK::CLI.run(cli_args.concat(%w[--force])) }.not_to raise_error
+    end
+
+    it 'does not start an interview when --force is used' do
+      expect(PDK::CLI::Util::Interview).to receive(:new).never
```
rodjek:
```suggestion
      expect(PDK::CLI::Util::Interview).not_to receive(:new)
```
nit: use a negative expectation rather than a negative matcher for consistency (here and in the other specs in this PR)

glennsarti: Whoops.  Good catch.

glennsarti: Added an extra commit as I found a bunch of other tests that should be changed as well.

PR #814, opened 2019-12-09
(PDK-1487) Add --default-template flag to pdk convert

rodjek:
Adds a simple flag for users to convert their module to the default
template.

```
pdk convert --default-template
```

is functionally equivalent to

```
pdk convert --template-url https://github.com/puppetlabs/pdk-templates
```

Additionally, it clears the saved `template-url` value so that future
operations will use the default template by default.

### DIFF
--- a/lib/pdk/cli/convert.rb
+++ b/lib/pdk/cli/convert.rb
@@ -11,6 +11,7 @@ module PDK::CLI
     flag nil, :noop, _('Do not convert the module, just output what would be done.')
     flag nil, :force, _('Convert the module automatically, with no prompts.')
     flag nil, :'add-tests', _('Add any missing tests while converting the module.')
+    flag nil, :'default-template', _('Convert the module to use the default PDK template.')
 
     run do |opts, _args, _cmd|
       require 'pdk/module/convert'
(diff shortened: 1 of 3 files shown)


lib/pdk/cli/convert.rb
```diff
@@ -22,12 +23,19 @@ module PDK::CLI
...
-      PDK::CLI::Util.validate_template_opts(opts)
-
       if opts[:noop] && opts[:force]
         raise PDK::CLI::ExitWithError, _('You can not specify --noop and --force when converting a module')
       end
 
+      if opts[:'default-template']
+        raise PDK::CLI::ExitWithError, _('You can not specify --template-url and --default-template.') if opts[:'template-url']
+
+        opts[:'template-url'] = PDK::Util::TemplateURI.default_template_addressable_uri.to_s
```
glennsarti: Should this be resetting template-ref as well?

rodjek: I thought that being able to do something like `pdk convert --default-template --template-ref 1.14.0` might be useful, so I left --template-ref alone. In terms of the saved ref in the answer file, it's stored as the fragment of the template-url uri, so it gets reset automatically for free.

glennsarti: Ahhh ok 👍

PR #816, opened 2019-12-11
(PDK-1488) Inform user if updating a pinned module

rodjek:


lib/pdk/cli/update.rb
```diff
@@ -48,6 +48,18 @@ module PDK::CLI
 
       updater = PDK::Module::Update.new(opts)
 
+      if updater.pinned_to_puppetlabs_template_tag?
+        PDK.logger.info _(
```
glennsarti: Potentially (not right now) the default template could be pointed to a different location via the PDK config.  Should this instead be about the `default` template URI instead of the `puppetlabs` template URI?

lib/pdk/util/template_uri.rb
```diff
@@ -7,6 +7,7 @@ class TemplateURI
 
       PACKAGED_TEMPLATE_KEYWORD = 'pdk-default'.freeze
       DEPRECATED_TEMPLATE_URL = 'https://github.com/puppetlabs/pdk-module-template'.freeze
+      PDK_TEMPLATE_URL = 'https://github.com/puppetlabs/pdk-templates'.freeze
```
glennsarti: Is this a semi-duplicate of   https://github.com/puppetlabs/pdk/blob/bd5e6cea2d528a758a6d93cf756accc100abd16a/lib/pdk/util/template_uri.rb#L133 ?

glennsarti: if it is can we just `Addressable::URI.new(PDK_TEMPLATE_URL) ` instead?

PR #822, opened 2019-12-18
(#821) Allow for unbalanced JSON fragments in RSpec output

### LINKED ISSUE #821: Generated descriptions for certain test comparisons break JUnit output
Generated descriptions can cause the JSON parsing for JUnit output to fail.  When this occurs the created XML document will be empty.

### How to reproduce
Generate a new module with PDK.
```
$ pdk new module test-broken_junit
```

Create a unit test with the following content.
```
# ./spec/unit/broken_junit_spec.rb
# This test will break JUNIT output
RSpec.describe "broken-junit" do
  let(:large_nested_hash) do
    {
      single_one: 'single_one',
      single_two: 'single_two',
      single_three: 'single_three',
      single_four: 'single_four',
      single_five: 'single_five',
      nested: {
        nested_one: 'nested_one',
        nested_two: 'nested_two',
        nested_three: 'nested_three',
        nested_four: 'nested_four',
        nested_five: 'nested_five',
      },
    }
  end

  subject do
    large_nested_hash
  end

  it { is_expected.to be_a Hash }
  it { is_expected.to eq large_nested_hash }
end
```

Running the test should result in a success.
```
$ pdk test unit -v
pdk (INFO): Using Ruby 2.5.7
pdk (INFO): Using Puppet 6.11.0
[✔] Preparing to run the unit tests.
/opt/puppetlabs/pdk/private/ruby/2.5.7/bin/ruby -I/opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/rspec-core-3.9.0/lib:/opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/rspec-support-3.9.0/lib /opt/puppetlabs/pdk/share/cache/ruby/2.5.0/gems/rspec-core-3.9.0/exe/rspec --pattern spec/\{aliases,classes,defines,functions,hosts,integration,plans,tasks,type_aliases,types,unit\}/\*\*/\*_spec.rb --format documentation

broken-junit
  is expected to be a kind of Hash
  is expected to eq {:nested=>{:nested_five=>"nested_five", :nested_four=>"nested_four", :nested_one=>"nested_one", :nest...>"single_four", :single_one=>"single_one", :single_three=>"single_three", :single_two=>"single_two"}

Finished in 0.00334 seconds (files took 0.17747 seconds to load)
2 examples, 0 failures
```

It is important to notice that the hash comparison test description output (`is expected to eq {:nested...`) has two opening curly braces (`{`) and only one closing curly brace (`}`).  The is an artifact of how the test descriptions are condensed for output.  This causes the processing for JUnit output to fail.

Running the test with the JUnit format specified should result in an empty document.
```
pdk (INFO): Using Ruby 2.5.7
pdk (INFO): Using Puppet 6.11.0
[✔] Preparing to run the unit tests.
[✔] Running unit tests.
<?xml version='1.0'?>
<testsuites/>
```

If you comment out enough of the `single_` lines to ensure that the output produces two closing curly braces (`}`) it will generate the XML as expected.

### Extra info
I dug through the `lib/pdk/tests/unit.rb` file and found that the output is processed by the `PDK::Util.find_first_json_in` method ([line 112](https://github.com/puppetlabs/pdk/blob/master/lib/pdk/tests/unit.rb#L112)).  I was able to confirm that the RSPEC command output JSON is valid.  So, I went ahead and chased that method into the `lib/pdk/util.rb` file.  [Line 219](https://github.com/puppetlabs/pdk/blob/master/lib/pdk/util.rb#L219) appears to be the culprit.  I was able to confirm that when that condition exists it does not properly capture the JSON output from the command.  Unfortunately, I don't have a fix to propose as I don't know the conditions that need to be satisfied with that REGEX.

scotje:
This is an initial attempt at making the PDK::Util.find_valid_json_in
method tolerate unbalanced JSON fragments nested as values inside of
a larger, valid JSON document.

The simple change of making the closing brace optional allows the regex to more greedily match the largest valid document, but I would like to see this tested against more complex samples before declaring this a sufficient fix.

Resolves #821

### DIFF
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -14,6 +14,7 @@ module PDK
     autoload :Env, 'pdk/util/env'
     autoload :Filesystem, 'pdk/util/filesystem'
     autoload :Git, 'pdk/util/git'
+    autoload :JSONFinder, 'pdk/util/json_finder'
     autoload :PuppetStrings, 'pdk/util/puppet_strings'
     autoload :PuppetVersion, 'pdk/util/puppet_version'
     autoload :RubyVersion, 'pdk/util/ruby_version'
@@ -196,7 +197,7 @@ module PDK
     # @return [Hash, nil] subset of text as Hash of first valid JSON found, or nil if no valid
     #   JSON found in the text
     def find_first_json_in(text)
-      find_valid_json_in(text)
+      find_all_json_in(text).first
     end
     module_function :find_first_json_in
 
@@ -206,41 +207,9 @@ module PDK
     # @return [Array<Hash>] subset of text as Array of all JSON object found, empty Array if none are found
     #   JSON found in the text
     def find_all_json_in(text)
-      find_valid_json_in(text, break_on_first: false)
+      PDK::Util::JSONFinder.new(text).objects
     end
     module_function :find_all_json_in
-
-    # Iterate through possible JSON documents until we find one that is valid.
-    #
-    # @param [String] text the text in which to find a JSON document
-    # @param [Hash] opts options
-    # @option opts [Boolean] :break_on_first Whether or not to break after valid JSON is found, defaults to true
-    #
-    # @return [Hash, Array<Hash>, nil] subset of text as Hash of first valid JSON found, array of all valid JSON found, or nil if no valid
-    #   JSON found in the text
-    #
-    # @private
-    def find_valid_json_in(text, opts = {})
-      break_on_first = opts.key?(:break_on_first) ? opts[:break_on_first] : true
-
-      json_result = break_on_first ? nil : []
-
-      text.split("\n").select { |r| r.start_with?('{') && r.end_with?('}') }.each do |str|
-        begin
-          if break_on_first
-            json_result = JSON.parse(str)
-            break
-          else
-            json_result.push(JSON.parse(str))
-          end
-        rescue JSON::ParserError
-          next
-        end
-      end
-
-      json_result
-    end
-    module_function :find_valid_json_in
 
     # Returns the targets' paths relative to the working directory
     #
--- a/lib/pdk/util.rb
+++ b/lib/pdk/util.rb
@@ -225,7 +225,7 @@ module PDK
 
       json_result = break_on_first ? nil : []
 
-      text.scan(%r{\{(?:[^{}]|(?:\g<0>))*\}?}x) do |str|
+      text.split("\n").select { |r| r.start_with?('{') && r.end_with?('}') }.each do |str|
         begin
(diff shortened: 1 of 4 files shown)


scotje:
This isn't working for `--parallel` mode, going to iterate a bit more.

lib/pdk/util.rb
```diff
@@ -196,7 +197,7 @@ def in_module_root?(path = Dir.pwd)
     # @return [Hash, nil] subset of text as Hash of first valid JSON found, or nil if no valid
     #   JSON found in the text
     def find_first_json_in(text)
-      find_valid_json_in(text)
+      PDK::Util::JSONFinder.new(text).objects.first
```
glennsarti: nit: Should this just be `find_all_json_in(text).first` ?

spec/acceptance/test_unit_spec.rb
```diff
@@ -211,6 +211,77 @@
...
+
+              it { is_expected.to be_a Hash }
+              it { is_expected.to eq large_nested_hash }
+            end
+          EOF
+        end
+      end
+
+      after(:all) do
+        FileUtils.rm_rf(File.join('spec', 'unit'))
```
glennsarti: This seems dangerous. What's cwd at this point?

lib/pdk/util.rb
rodjek: :+1:

spec/acceptance/test_unit_spec.rb
rodjek: cwd is the test module directory, but i can replace this with a couple of safer methods

glennsarti: 👍 I imagine my face if this ever went hay-wire would be 😕💥😢

Absolute path would be nice but that could be hard to get 🤷‍♂️

spec/acceptance/test_unit_spec.rb
```diff
@@ -4,6 +4,16 @@
 describe 'pdk test unit', module_command: true do
   include_context 'with a fake TTY'
 
+  shared_context 'with spec file' do |filename, content|
+    around(:all) do |example|
+      path = File.join('spec', 'unit', filename)
+      FileUtils.mkdir_p(File.dirname(path))
+      File.open(path, 'w') { |f| f.puts content }
```
glennsarti: Do we care about line endings here?  (not `wb` so on Windows it'll use CRLF)

rodjek: Nope, for the purposes of these short lived files, the line endings don't matter

PR #824, opened 2020-01-08
(PDK-1556) Use the module root when generating objects

glennsarti:
Previously the `pdk new` command used the current directory as the module root
which caused errors when users executed `pdk new` when deep inside a module.
This commit changes the CLI to use the actual module root when creating the
generators, and updates the tests for this scenario.


glennsarti:
This PR does change the output of `pdk convert` and `pdk update` to now display the full path to the user, instead of just the relative path to current directory.

PR #825, opened 2020-01-10
(maint) Use case insensitive environment variable names on Windows

glennsarti:
Previously the PDK Env class had case sensitive keys which meant 'PATH' could
be nil but 'Path' would have the correct value.  This commit updates the
PDK::Util::Env class to be case insensitive on Windows, by having it's own
Hash implementation. And delegates to ENV on non-windows platforms.  This
commit also adds tests for these scenarios.

### DIFF
--- a/lib/pdk/util/env.rb
+++ b/lib/pdk/util/env.rb
@@ -4,25 +4,42 @@ require 'forwardable'
 module PDK
   module Util
     class Env
+      class WindowsEnv
+        extend Forwardable
+
+        # Note, these delegators may not have case insensitive keys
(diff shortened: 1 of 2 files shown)


glennsarti:
Adds tests to test suite.  Coverage drops....  Wha?!?!?!

rodjek:
Yeah, coverage % is really the wrong metric to report on a PR as increasing the lines of code (even when they have 100% coverage) can reduce the overall %. Ideally it should be notifying on the number of lines changed (e.g. 3 new uncovered lines introduced in this PR, etc)

PR #826, opened 2020-01-14
(PDK-1557) Detect Control Repositories

glennsarti:
Previously the PDK treated everything as a module.  This commit begins the work
to add Control Repo support by detecting Control Repositories and Bolt Project
directories.


spec/unit/pdk/control_repo_spec.rb
```diff
@@ -0,0 +1,112 @@
+require 'spec_helper'
+require 'pdk/util'
+
+describe PDK::ControlRepo do
+  describe '.control_repo_root' do
+    [
+      { name: 'with strict_check set to false', strict_check: false },
+      { name: 'with strict_check set to true but is not a Bolt project dir', strict_check: true },
+    ].each do |testcase|
```
scotje: Is there a reason to not just do this with shared examples and two explicit `context`s? Dynamically generated specs can be a pain to debug.

lib/pdk/control_repo.rb
```diff
@@ -0,0 +1,45 @@
...
+    #   not also a Puppet Bolt project directory  Default is false.
+    #
+    # @return [String, nil] Fully qualified base path to Control Repo, or nil if
+    #   the current working dir does not appear to be within a Control Repo.
+    def control_repo_root(strict_check = false)
+      environment_conf_path = PDK::Util.find_upwards('environment.conf')
+      path = if environment_conf_path
+               File.dirname(environment_conf_path)
+             elsif in_control_repo_root?(Dir.pwd)
+               Dir.pwd
```
scotje: If we are requiring an `environment.conf` to be present to consider something a control-repo, what additional benefit does this branch get us versus the `find_upwards` search? (`find_upwards` checks in `pwd` first before moving up the tree)

lib/pdk/control_repo.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+    module_function :control_repo_root
+
+    # Returns true or false depending on if any of the common files in a Control Repo
+    # are found in the specified directory. If a directory is not specified, the current
+    # working directory is used.
+    #
+    # @return [boolean] True if any folders from MODULE_FOLDERS are found in the current dir,
+    #   false otherwise.
+    def in_control_repo_root?(path = Dir.pwd)
```
scotje:
```suggestion
    def is_control_repo_root?(path = Dir.pwd)
```

Same as in the Bolt code, I think "is" makes slightly more sense given you can pass any arbitrary path.

lib/pdk/control_repo.rb
```diff
@@ -0,0 +1,45 @@
...
+      PDK::Bolt.in_bolt_project_root?(path) ? nil : path
+    end
+    module_function :control_repo_root
+
+    # Returns true or false depending on if any of the common files in a Control Repo
+    # are found in the specified directory. If a directory is not specified, the current
+    # working directory is used.
+    #
+    # @return [boolean] True if any folders from MODULE_FOLDERS are found in the current dir,
+    #   false otherwise.
```
scotje: Comment needs to be updated

spec/unit/pdk/control_repo_spec.rb
glennsarti: Fixed

lib/pdk/bolt.rb
```diff
@@ -0,0 +1,19 @@
...
+module PDK
+  module Bolt
+    # Returns true or false depending on if any of the common files and directories in
+    # a Bolt Project are found in the specified directory. If a directory is not specified,
+    # the current working directory is used.
+    #
+    # @see https://puppet.com/docs/bolt/latest/bolt_project_directories.html
+    #
+    # @return [boolean] True if any folders from MODULE_FOLDERS are found in the current dir,
+    #   false otherwise.
```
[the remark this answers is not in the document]
glennsarti: Fixed

lib/pdk/bolt.rb
```diff
@@ -0,0 +1,19 @@
...
+    def in_bolt_project_root?(path = Dir.pwd)
```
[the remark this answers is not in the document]
glennsarti: Fixed.  Changed to `bolt_project_root?` because rubocop doesn't like "is_*?"

lib/pdk/control_repo.rb
glennsarti: `in_control_repo_root?` looks for things that like _like_ a control repo (e.g. Puppetfile) whereas this method looks for something explicit.  TBH this is mostly a copy-paste from the module detection. 

I'm still in two minds about the whole explicit vs fuzzy search.  The docs clearly state that "environment.conf" == Control Repo and Puppet itself will error if it's not present.  Whereas in module land, Puppet will use a directory without a metadata.json happily, but we prefer a metadata.json.

lib/pdk/control_repo.rb
glennsarti: Fixed

lib/pdk/control_repo.rb
glennsarti: Changed to `control_repo_root?`

lib/pdk/control_repo.rb
glennsarti: @rodjek Thoughts?

glennsarti: Had a chat with @rodjek.  And while non-directory based environment control repos do exist, they are not the preferred way.  Also detecting them is problematic as they don't have an easily identifiable thing to search for.

So for this PR  `find_control_repo_root` will look for environment.conf whereas `control_repo_root?` will look for something that _could_ be a control repo. Use cases:

* `find_control_repo_root` would be used in a `pdk validate` operation
* `control_repo_root?` would be used in a `pdk convert` operation

glennsarti:
@scotje Ready for review again.

PR #827, opened 2020-01-16
(PDK-1588) Increase granularity of `pdk bundle` analytics

rodjek:
Up until now, when a user ran a `pdk bundle` command we would only
report a `bundle` screen view to our analytics. This resulted in us
finding out that a lot of users use `pdk bundle` but provided no insight
into why.

When a user runs `pdk bundle`, we will now capture the bundler
subcommand (but not any of the other arguments). The only exception to
this is if the user runs `pdk bundle exec` in which case we will also
capture the name of the command that is being executed (but again, none
of the arguments).

e.g.

 * `pdk bundle show some_gem` will now be reported as a `bundle_show`
    screen view.
 * `pdk bundle exec rspec path/to/some/ruby_spec.rb` will now be
    reported as a `bundle_exec_rspec` screen view.

### DIFF
--- a/spec/unit/pdk/cli/bundle_spec.rb
+++ b/spec/unit/pdk/cli/bundle_spec.rb
@@ -2,7 +2,68 @@ require 'spec_helper'
 require 'pdk/cli'
 
 describe 'Running `pdk bundle`' do
-  subject(:test_cmd) { PDK::CLI.instance_variable_get(:@bundle_cmd) }
+  let(:command_args) { ['bundle'] }
+  let(:command_result) { { exit_code: 0 } }
 
-  it { is_expected.not_to be_nil }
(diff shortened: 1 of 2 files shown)


spec/unit/pdk/cli/bundle_spec.rb
```diff
@@ -2,7 +2,68 @@
...
-  it { is_expected.not_to be_nil }
+  context 'when it calls bundler successfully' do
+    after(:each) do
+      expect {
+        PDK::CLI.run(command_args)
+      }.to exit_zero
+    end
+
+    before(:each) do
+      mock_command = instance_double(
```
glennsarti: All this mocking is a bit of a smell, but that's not for this PR.

PR #832, opened 2020-01-19
(PDK-1587) Reject paths with non-ASCII characters when building

rodjek:


lib/pdk/module/build.rb
```diff
@@ -212,6 +214,27 @@ def validate_ustar_path!(path)
...
+      # characters that are not supported by the hosts locale. In order to
+      # maximise compatibility we limit the paths to contain only ASCII
+      # characters, which should be part of any locale character set.
+      #
+      # @param path [String] the relative path to be added to the tar file.
+      #
+      # @raise [ArgumentError] if the path contains non-ASCII characters.
+      #
+      # @return [nil]
+      def validate_path_encoding!(path)
```
glennsarti: Encoding may or may not be an issue in this regex. A quick look on my Windows laptop showed the string encoding is coming in as the system locale (In my case EN-AU)

```
[2] pry(#<PDK::Module::Build>)> path.encoding
=> #<Encoding:Windows-1252>
[3] pry(#<PDK::Module::Build>)> relative_path
=> #<Pathname:.github>
[4] pry(#<PDK::Module::Build>)> relative_path.to_s.encoding
=> #<Encoding:Windows-1252>
[5] pry(#<PDK::Module::Build>)> whereami

From: C:/Source/tmp/pdk-pr832/lib/pdk/module/build.rb @ line 119 PDK::Module::Build#stage_path:

    113:       def stage_path(path)
    114:         require 'pathname'
    115:
    116:         relative_path = Pathname.new(path).relative_path_from(Pathname.new(module_dir))
    117:         dest_path = File.join(build_dir, relative_path)
    118: require 'pry'; binding.pry
 => 119:         validate_path_encoding!(relative_path.to_path)
    120:
    121:         if PDK::Util::Filesystem.directory?(path)
    122:           PDK::Util::Filesystem.mkdir_p(dest_path, mode: PDK::Util::Filesystem.stat(path).mode)
    123:         elsif PDK::Util::Filesystem.symlink?(path)
    124:           warn_symlink(path)
    125:         else
    126:           validate_ustar_path!(relative_path.to_path)
    127:           PDK::Util::Filesystem.cp(path, dest_path, preserve: true)
    128:         end
    129:       rescue ArgumentError => e
    130:         raise PDK::CLI::ExitWithError, _(
    131:           '%{message} Rename the file or exclude it from the package ' \
    132:           'by adding it to the .pdkignore file in your module.',
    133:         ) % { message: e.message }
    134:       end
```

glennsarti: So the question is, is there a case where the regex will fail to match a non-ASCII character for a given encoding and vice versa, is there a case of false positives

spec/unit/pdk/module/build_spec.rb
```diff
@@ -181,6 +181,14 @@
       allow(instance).to receive(:release_name).and_return(release_name)
     end
 
+    context 'when the path contains non-ASCII characters' do
```
glennsarti: Perhaps

```
    context 'when the path contains non-ASCII characters' do
      RSpec.shared_examples "a failing path" do |relative_path|
        let (:path) do
          # We enforce encoding here as the filesystem encoding can be different from the default
          # String encoding. And encoding can make the resulting string behave differently
          "#{module_dir}\\#{relative_path}".force_encoding(Encoding.find("filesystem"))
        end

        before(:each) do
          allow(PDK::Util::Filesystem).to receive(:directory?).with(path).and_return(false)
          allow(PDK::Util::Filesystem).to receive(:symlink?).with(path).and_return(false)
          allow(PDK::Util::Filesystem).to receive(:cp).with(path, Object, Object).and_return(true)
        end

        it "exits with an error for #{relative_path}" do
          expect {
            instance.stage_path(path)
          }.to raise_error(PDK::CLI::ExitWithError, %r{can not include non-ASCII characters})
          #puts instance.stage_path(path)
        end
      end

      include_examples "a failing path", "strange_unicode_\u{000100}"
      include_examples "a failing path", "\330\271to"
    end
```

lib/pdk/module/build.rb
```diff
@@ -212,6 +214,27 @@ def validate_ustar_path!(path)
...
+        return unless path =~ %r{[^\x00-\x7F]}
+
+        raise ArgumentError, _(
+          "'%{path}' can not include non-ASCII characters in its path or " \
```
glennsarti: nit: Not a huge fan of double negatives here.  Perhaps `can only include ASCII characters in its path ...`

lib/pdk/module/build.rb
glennsarti: I tried a bunch of things but couldn't make it break.  So this seems ok!

glennsarti:
Rekicking the build to get the license_finder fix

glennsarti:
The license_finder failures are not a result of this PR.

glennsarti:
If you rebase onto master it should go green.

PR #835, opened 2020-01-23
(PDK-1590) Use short path names when calling Bundler ruby bin

glennsarti:
Previously the full path was used to call the Bundler bin. Unforunately as in
commit 10736a4 Bundler is sensitive to spaces in path names.  This commit
modifies the bundler bin path generation to use short path name.

Note that this does not have unit tests as it can't really be tested. But it is
tested as part of the packaging tests.

Note that an additional Windows detection method (PDK::Util.on_windows?) is
introduced.  This method is designed specifically to _NOT_ be mocked and any
code using this method will probably raise errors on non-Windows platforms e.g.
calling Windows API via FFI.  Gem.win_platform? is commonly used in the PDK
codebase and is also commonly mocked.

### DIFF
--- a/lib/pdk/cli/exec.rb
+++ b/lib/pdk/cli/exec.rb
@@ -52,7 +52,14 @@ module PDK
         bundle_bin = Gem.win_platform? ? 'bundle.bat' : 'bundle'
         vendored_bin_path = File.join('private', 'ruby', PDK::Util::RubyVersion.active_ruby_version, 'bin', bundle_bin)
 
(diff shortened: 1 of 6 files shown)


scotje:
Has this passed through acceptance testing now?

glennsarti:
@scotje Not yet.  @rodjek is doing some of this but VMPooler is have "some technical issues"

rodjek:
Going to close this for now and we can reopen when we revisit the rubygems 3.1.x update in vanagon

PR #841, opened 2020-02-03
(PDK-1607)(PDK-1608) Implement system-level settings for PDK configuration

glennsarti:
~~Blocked by #849~~

---

**PDK-1607**

Previously in commit 5cb9b92 the `pdk config get` command was added however
this was in contrary to the Verb-Noun UX for the PDK CLI.  This commit:

* Copies the `pdk config` and `pdk config get` code (and tests) to be
  `pdk get` and `pdk get config` respectively
* Adds deprecation notices to the `pdk config *` commands
* Updates tests for the deprecation notices

**PDK-1608**

Previously only per-user settings existed for the PDK::Config classes.  This
commit:
* Adds the concept and file locations for system level settings
* Adds helper methods to PDK::Config to make it easier to query multilevel
  settings, and to query using precedence
* Makes it easier to eventually add per-project settings
* Adds tests for the new methods
* Deprecates the use of ".user". Once a ".set(..)" method is completed in the
  future, this method will be removed entirely

Note that the various PDK Modules and Classes still won't consume system-level
settings until they use the `PDK.config.pdk_setting(...` method.  Currently they
still use `PDK.config.user[...`.  This will changed in later commits.

### DIFF
--- a/lib/pdk/cli/config.rb
+++ b/lib/pdk/cli/config.rb
@@ -2,10 +2,12 @@ module PDK::CLI
   @config_cmd = @base_cmd.define_command do
     name 'config'
     usage _('config [subcommand] [options]')
-    summary _('Configure the Puppet Development Kit.')
+    summary _('(Deprecated) Configure the Puppet Development Kit.')
     default_subcommand 'help'
 
     run do |_opts, args, _cmd|
+      PDK.logger.warn _('The \'pdk config\' command is deprecated, please use \'pdk get config\' and \'pdk set config\' instead.')
+
       if args == ['help']
         PDK::CLI.run(%w[config --help])
         exit 0
--- a/lib/pdk/config.rb
+++ b/lib/pdk/config.rb
@@ -11,7 +11,27 @@ module PDK
     autoload :Validator, 'pdk/config/validator'
     autoload :YAML, 'pdk/config/yaml'
 
+    # The user configuration settings.
+    # @deprecated This method is only provided as a courtesy until the `pdk set config` CLI and associated changes in this class, are completed.
+    #             Any read-only operations should be using `.get` or `.get_within_scopes`
+    # @return [PDK::Config::Namespace]
     def user
+      user_config
+    end
+
+    # The system level configuration settings.
+    # @return [PDK::Config::Namespace]
+    # @api private
+    def system_config
+      @system ||= PDK::Config::JSON.new('system', file: PDK::Config.system_config_path) do
+        mount :module_defaults, PDK::Config::JSON.new(file: PDK::Config.system_answers_path)
+      end
+    end
+
+    # The user level configuration settings.
+    # @return [PDK::Config::Namespace]
+    # @api private
+    def user_config
       @user ||= PDK::Config::JSON.new('user', file: PDK::Config.user_config_path) do
         mount :module_defaults, PDK::Config::JSON.new(file: PDK.answers.answer_file_path)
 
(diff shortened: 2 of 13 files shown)


lib/pdk/config.rb
```diff
@@ -43,7 +49,26 @@ def user
     # @param filter [String] Only resolve setting names which match the filter. See PDK::Config::Namespace.be_resolved? for matching rules
     # @return [Hash{String => Object}] All resolved settings for example {'user.module_defaults.author' => 'johndoe'}
     def resolve(filter = nil)
-      user.resolve(filter)
+      self.system.resolve(filter).merge(user.resolve(filter)) # rubocop:disable Style/RedundantSelf Use an explicit call so as to not confuse with Kernel.system
```
scotje: Hmmm, yeah how to we feel about shadowing `Kernel.system` in this class? Is the brevity of `.user` and `.system` worth the potential collision? Should we just rename the methods to `.user_config` and `.system_config` even though that is slightly redundant?

lib/pdk/config.rb
```diff
@@ -43,7 +49,26 @@ def user
...
+      else
+        nil
+      end
+    end
+
+    # Gets a named setting using precedence from the user and system levels
+    # Note that name must NOT include user or system prefix
+    def pdk_setting(*name)
+      value = get(*['user'].concat(name))
+      value.nil? ? get(*['system'].concat(name)) : value
```
scotje: I feel like we are mostly going to want the "lookup using precedence" behavior that is currently implemented in `pdk_setting()`, so maybe what is currently `get()` should be like `get_within_scope(scope, *name)` and then the "lookup using precedence" can just become `get()`?

lib/pdk/config.rb
glennsarti: Yeah I thought long and hard about this too.  I eventually conceded that the brevity is useful because, that method will _normally_ be called via 'PDK,config.system`.  The shadowing is only really a problem as an implementation detail within `Config` class.

In fact, ideally `PDK,config.system` should never actually be called.  I should really mark that as `@api private` because once `Config.get` and `Config.set` is complete ALL access to `.system` will be internal to the object and won't need to be on the public API at all.

glennsarti: So in a round_about way I'm saying....

I should change it to `user_config` and `system_config` with the intention to mark those methods as private ASAP.

lib/pdk/config.rb
glennsarti: I disagree here, but I believe I can the point your trying to make here. And I guess it depends on the consumers point of view.

In the future, we'll have at least one addition setting level, `project.`, or even multiple different ones (`control_repo.` or `module.`).

So I see

`PDK::Config.get(...)` to be the most specific form. "Get me this setting name"

`PDK::Config.pdk_setting` to be a helper, to stop copying the precedence logic.  I chose `pdk_setting` so it would (hopefully) be obvious that this is only a PDK setting, not a Project setting.

I do concede that `PDK::Config.pdk_setting` seems like a redundant name.

As for your suggestion:

`get_within_scope(scope, *name)`  makes perfect sense

but `get()` only returning `user` and `system` level settings may be a confusing method name.

glennsarti: That said, I do like that we both agree on the layout of the methods, just the method names need some discussion 👍

lib/pdk/cli/config.rb
```diff
@@ -2,10 +2,12 @@ module PDK::CLI
   @config_cmd = @base_cmd.define_command do
     name 'config'
     usage _('config [subcommand] [options]')
-    summary _('Configure the Puppet Development Kit.')
+    summary _('(Deprecated) Configure the Puppet Development Kit.')
     default_subcommand 'help'
 
     run do |_opts, args, _cmd|
+      PDK.logger.warn _('The \'pdk config\' command is deprecated and will be removed.')
```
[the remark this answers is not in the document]
glennsarti: Fixed

lib/pdk/cli/config/get.rb
```diff
@@ -2,9 +2,11 @@ module PDK::CLI
   @config_get_cmd = @config_cmd.define_command do
     name 'get'
     usage _('config get [name]')
-    summary _('Retrieve the configuration for <name>. If not specified, retrieve all configuration settings')
+    summary _('(Deprecated) Retrieve the configuration for <name>. If not specified, retrieve all configuration settings')
 
     run do |_opts, args, _cmd|
+      PDK.logger.warn _('The \'pdk config get\' command is deprecated and will be removed.')
```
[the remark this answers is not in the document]
glennsarti: Fixed

lib/pdk/config.rb
glennsarti: Done.

glennsarti: Not marking as resolved for the moment, but has been implemented

lib/pdk/config.rb
glennsarti: @scotje You happy for merge are would prefer a name change?

glennsarti:
The Travis failure is incidental to this PR.

lib/pdk/config.rb
scotje: Doesn't it seem possible people will want to override a "PDK" setting at the `project` scope at some point as well? Or is it clear that the system/user settings will never overlap with project settings?

It seems like what we need for a general interface is:

- get a setting based on default precedence rules (I think that would be `project > user > system`?)
- get a setting from a specific precedence level
- (maybe?) get a setting from a specific subset of precedence levels (for when the caller has a special reason for bypassing normal precedence)

So maybe it's just that `get_within_scope` should also accept a list for the `scope` parameter? Or we could have a `get_within_scopes`?

glennsarti: Okay, I've basically implemented what you stated. I've gone with:

* `.get`  This gets a fully named setting e.g. `user.foo.bar`.  No scope precedence or anything

* `.get_within_scopes` : This gets a setting name in zero or more specified scopes (in order).  If no scopes are given (nil) then it searches in the default scopes in the default order.

Which ends up with:
* For the bulk of calls within /lib/cli/get/config `.get` is probably the best
* For the bulk of calls within /lib/* `.get_within_scopes` is probably the best
* For critical calls like analytics information, it can be restricted to `.get_within_scopes('....', ['user', 'system'])`.  Later on, I may add a shortcut method (`def critical_scopes; ['user', 'system`].freeze; end;`) if I find that we use it too often.

lib/pdk/config/namespace.rb
```diff
@@ -98,6 +98,8 @@ def [](key)
         return nil if settings[key.to_s].nil?
         return settings[key.to_s].value unless settings[key.to_s].value.nil?
         default_value = settings[key.to_s].default
+        # Duplicate arrays and hashes so that they are isolated from changes being made
+        default_value = default_value.dup if default_value.is_a?(Hash) || default_value.is_a?(Array)
```
scotje: Are these still shallow copies? Is there a risk if there are nested hashes or arrays?

lib/pdk/config/setting.rb
```diff
@@ -46,8 +46,9 @@ def qualified_name
         [namespace.name, @name].join('.')
       end
 
-      def value # rubocop:disable Style/TrivialAccessors
-        @value
+      def value
+        # Duplicate arrays and hashes so that they are isolated from changes being made
+        (@value.is_a?(Hash) || @value.is_a?(Array)) ? @value.dup : @value
```
scotje: Same question as above, is there a risk if these are only shallow copies?

lib/pdk/config/namespace.rb
glennsarti: Whoops. forgot to comment.  Created a deep_dup method to do the copying.

PR #842, opened 2020-02-04
(PDK-1113) Use PDK configuration instead of AnswerFile class

glennsarti:
~~Builds on  #841~~ Merged

---

Previously the PDK had two ways to retrieve the same configuration information.
One via the PDK::Config classes and the other the AnswerFile class.  This should
only be one.  This commit removes the AnswerClass (and associated features) and
instead uses PDK::Config for setting management

* Removes the PDK::AnswerFile class, and instead has a single helper method for
  the default file location.  This may eventually be moved into the PDK::Config
  namespace as well

* Removes the hidden `--answer-file` command line option and instead uses an
  environment variable PDK_ANSWER_FILE.  This feature is not supported and is
  only used to acceptance testing.  Note that we moved to an environment
  variable as there was a subtle race condition where the CLI reads from
  PDK::Config, but the CLI also moves the user answer file location. By using
  an environment variable this race condition can no longer occur.

* Move setting reads from PDK.answers to PDK.config.pdk_setting(....)

* Move setting writes from PDK.answers.update! to
  PDK.config.user['module_defaults']['....'] =

* Modify tests for the setting read/write changes

* Added a Mock configuration RSpec Context to stop any test information
  affecting actual real setting files.  Previously module answer files were
  being modified


scotje:
+10 to folding answer file functionality into pdk config.

This does not actually remove the file yet though correct? Just a unified frontend to the existing file alongside other config?

glennsarti:
> This does not actually remove the file yet though correct? Just a unified frontend to the existing file alongside other config?

Correct, and I don't think we'll _actually_ need to move the file either.  It's doing a perfectly fine job right now, and for the most part hidden to users day-to-day activities.

I'm actually thinking of adding a `pdk get config-files` command line option to list all of the configuration file locations, for each of the PDK setting roots e.g.

```
> pdk get config-files
{
   "user": "something/somewhere.json",
   "user.module_defaults": "something/other_dir/somewhere.json",
... and so on
}
```

scotje:
To me, it makes sense to leave the config files referenced by other tools in place and just provide a frontend to them as you say.

For files only referenced by PDK however, I'm in favor of eventually converging those into a single file for each config layer for simplification.

glennsarti:
The rebase has broken lots of stuff.  Back to WIP.

glennsarti:
Need to change this PR for the changes of PR 841

lib/pdk/generate/module.rb
```diff
@@ -135,13 +135,13 @@ def self.prepare_metadata(opts = {})
...
 
-        opts[:username] = (opts[:username] || PDK.answers['forge_username'] || username_from_login).downcase
+        opts[:username] = (opts[:username] || PDK.config.get_within_scopes('module_defaults.forge_username') || username_from_login).downcase
 
         defaults = PDK::Module::Metadata::DEFAULTS.dup
 
         defaults['name'] = "#{opts[:username]}-#{opts[:module_name]}" unless opts[:module_name].nil?
-        defaults['author'] = PDK.answers['author'] unless PDK.answers['author'].nil?
-        defaults['license'] = PDK.answers['license'] unless PDK.answers['license'].nil?
+        defaults['author'] = PDK.config.get_within_scopes('module_defaults.author') unless PDK.config.get_within_scopes('module_defaults.author').nil?
```
rodjek: We should come up with a helper around this get-and-set-unless-nil pattern that isn't so verbose. Maybe a method that yields the value to a block if not nil?

Something like
```
PDK.config.with_scoped_value('module_defaults.author') { |v| defaults['author'] = v }
```

spec/acceptance/convert_spec.rb
```diff
@@ -156,17 +156,28 @@
...
 
-    answer_file = File.join('..', 'non_default_template_answers.json')
+    before(:each) do
+      @old_answer_file = ENV['PDK_ANSWER_FILE']
+      ENV['PDK_ANSWER_FILE'] = File.expand_path(File.join('..', 'non_default_template_answers.json'))
+    end
+
+    after(:each) do
+      ENV['PDK_ANSWER_FILE'] = @old_answer_file # rubocop:disable RSpec/InstanceVariable This is fine.
+    end
```
rodjek: We can get rid of the instance variable and make rubocop happy using an `around` hook here instead.

```ruby
around(:each) do |example|
  old_answer_file = ENV['PDK_ANSWER_FILE']
  ENV['PDK_ANSWER_FILE'] = File.expand_path(File.join('..', 'non_default_template_answers.json'))
  example.run
  ENV['PDK_ANSWER_FILE'] = old_answer_file
end
```

spec/acceptance/convert_spec.rb
```diff
@@ -156,17 +156,28 @@
...
+      # Note - The following file(...) tests can not be run in isolation and require the above
+      # `its(:exit_status)` to have run first.
       describe file('metadata.json') do
         its(:content_as_json) { is_expected.to include('template-url' => "#{template_repo}#master") }
       end
 
-      describe file(answer_file) do
-        its(:content_as_json) { is_expected.to include('template-url' => nil) }
+      describe file(File.join('..', 'non_default_template_answers.json')) do
+        it 'has a nil template-url' do
```
rodjek: We should be able to simplify this it block as `its(:content_as_json) { is_expected.to include('template-url' => nil) }`

spec/acceptance/template_ref_spec.rb
```diff
@@ -12,10 +14,18 @@
       'pdk', 'new', 'module', 'foo',
       '--skip-interview',
       '--template-url', 'https://github.com/puppetlabs/pdk-templates',
-      '--template-ref', '1.7.0',
-      '--answer-file', 'foo_answers.json'
+      '--template-ref', '1.7.0'
     ]
 
+    before(:each) do
```
rodjek: `around(:each)` hook - see convert_spec.rb

glennsarti: Will fix

spec/acceptance/convert_spec.rb
glennsarti: Updated the test with better subject name and comments as to why I had to do this.

spec/acceptance/convert_spec.rb
glennsarti: Fixed

lib/pdk/generate/module.rb
glennsarti: I like that!  Will implement.

glennsarti: Added a commit which does this.

glennsarti:
CI is green!

PR #849, opened 2020-02-12
(maint) Use Ruby 2.5 on Travis for PDK as a library tests

glennsarti:
Concurrent-ruby 1.1.6 has a strange thread timing issue under Ruby 2.4 and the
spec/unit/pdk/analytics/util_spec.rb file. Causing facter to error with
"cannot alloc thread" which then halts the test. It doesn't occur on Appveyor
(Windows) or the regular rspec under Ruby 2.4 testing. So just use Ruby 2.5
instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
-sudo: false
+---
+os:
+  - linux
+dist: xenial
 language: ruby
 bundler_args: "--without development"
 script:
@@ -15,7 +18,7 @@ before_install:
       gem install bundler --no-document;
     fi
 cache: bundler
-matrix:
+jobs:
   include:
     - rvm: 2.4.3
       env: CHECK=rubocop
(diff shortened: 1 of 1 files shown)


rodjek:
@DavidS noticed an issue with concurrent-ruby 1.1.6 on ruby 2.5 https://github.com/ruby-concurrency/concurrent-ruby/issues/849, perhaps for now we should just pin to the previous version of concurrent-ruby instead?

glennsarti:
I thought about that too and initially tested it (works fine).  I decided to not go down that path:

1. The regular ruby 2.4 specs work fine, so it's definitely _something_ to do with the Travis config.
2. Ruby 2.5.5 is working (so far, which is what RUBY_VER=2.5 is picking up).  David said it broke on Ruby 2.5.1.
3. The failing test cell is only testing that we can run the _spec.rb files individually, not so much that the tests _actually_ work.  So I went the route of smallest change (one travis cell as opposed to pinning the Gem for everyone)
4. Also means we'll pick up the new updated gem when (/if) it gets released

PR #853, opened 2020-02-18
(PDK-1612) Add PDK::Context and context detection

glennsarti:
Previously the PDK assumed it was always within a Puppet Module, but the PDK can
be used in far more different contexts.  This commit implements the PDK Context
RFC [1], in preparation for other parts of the PDK to consume it, for example
PDK validators:

* Adds the base Context and three concrete contexts: None, Module and
  ControlRepo
* Adds a memoized context object to PDK (PDK.context), for consumption by other
  parts of the PDK
* Updates the module pdk compatibility methods to accept a module path instead
  of assume the PDK is always in a module. This is needed for the Module context
  to check if an arbitrary module is PDK compatible
* Adds two test fixtures (a Puppet module and a control repo) for use by rspec
  for contex detection
* Adds tests for context detection via PDK::Context.create

[1] https://github.com/puppetlabs/pdk-planning/blob/master/RFCs/0007-add-pdk-context.md

### DIFF
--- a/lib/pdk/context.rb
+++ b/lib/pdk/context.rb
@@ -1,0 +1,96 @@
+require 'pdk'
+
+module PDK
+  module Context
+    autoload :None, 'pdk/context/none'
+    autoload :Module, 'pdk/context/module'
+    autoload :ControlRepo, 'pdk/context/control_repo'
+
+    # Automatically determines the PDK Context given a path. Create will continue up the directory tree until it
+    # finds a valid context
+    # @return [PDK::Context::AbstractContext] Returns a PDK::Context::None if the context could not be determined
+    def self.create(context_path)
+      return PDK::Context::None.new(context_path) unless PDK::Util::Filesystem.directory?(context_path)
+
+      previous = nil
+      current = PDK::Util::Filesystem.expand_path(context_path)
+      until !PDK::Util::Filesystem.directory?(current) || current == previous
+        # Control Repo detection
+        return PDK::Context::ControlRepo.new(current, context_path) if PDK::ControlRepo.control_repo_root?(current)
+
+        # Puppet Module detection
+        metadata_file = File.join(current, 'metadata.json')
+        if PDK::Util::Filesystem.file?(metadata_file) || PDK::Util.in_module_root?(context_path)
+          return PDK::Context::Module.new(current, context_path)
+        end
+
+        previous = current
+        current = PDK::Util::Filesystem.expand_path('..', current)
+      end
+      PDK::Context::None.new(context_path)
+    end
+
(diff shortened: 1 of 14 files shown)


lib/pdk/context.rb
```diff
@@ -0,0 +1,97 @@
...
+      return PDK::Context::None.new(context_path) unless PDK::Util::Filesystem.directory?(context_path)
+
+      previous = nil
+      current = PDK::Util::Filesystem.expand_path(context_path)
+      until !PDK::Util::Filesystem.directory?(current) || current == previous
+        # Control Repo detection
+        return PDK::Context::ControlRepo.new(current, context_path) if PDK::ControlRepo.control_repo_root?(current)
+
+        # Puppet Module detection
+        # Note - The metadata.json file check should really be in PDK::Util.in_module_root?
```
rodjek: This was done intentionally so that modules without a `metadata.json` could be converted.

lib/pdk/util.rb
```diff
@@ -248,16 +248,18 @@ def targets_relative_to_pwd(targets)
     module_function :targets_relative_to_pwd
 
     # TO-DO: Refactor replacement of lib/pdk/module/build.rb:metadata to use this function instead
-    def module_metadata
+    # @param module_path [String] The path to the root of the module. Default is determine the module root automatically
+    def module_metadata(module_path = nil)
       require 'pdk/module/metadata'
-
-      PDK::Module::Metadata.from_file(File.join(module_root, 'metadata.json')).data
+      module_path = module_root if module_path.nil?
```
rodjek:
```suggestion
      module_path ||= module_root
```

Or just use `module_path || module_root` in the `File.join` below

lib/pdk/context.rb
glennsarti: Ok.  Will remove the comment

lib/pdk/util.rb
glennsarti: Will change.

lib/pdk/context.rb
glennsarti: Hrmm... so `.in_module_root?` only checks for common directories. So adding a check for metadata.json wouldn't affect that scenario.

In fact in a empty directory with only `metadata.json`:
`PDK::Util.module_root` will return the module path but
`PDK::Util.in_module_root?(..)` for that same path will return false which is "odd"

glennsarti: Removed the comment.  Can debate this elsewhere.

lib/pdk/util.rb
glennsarti: Fixed

PR #856, opened 2020-02-19
(PDK-1615) Add Ini File configuration support

glennsarti:
This commit adds an Ini File configuration namespace which can read and write to
Ini files. This is needed for reading environment.conf in control repositories.

### DIFF
--- a/lib/pdk/config.rb
+++ b/lib/pdk/config.rb
@@ -2,6 +2,8 @@ require 'pdk'
 
 module PDK
   class Config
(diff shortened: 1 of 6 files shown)


glennsarti:
@rodjek Yeah, understand. I kept the original code formatting to make it easier in the future to copy-pasta any updates from the puppet code-base.  But it's basically pick your poison.

glennsarti:
>  this is just based on the internal puppetlabs/puppet INI file implementation?

Yes.

PR #857, opened 2020-02-21
(PDK-1614) Add project.environment settings

glennsarti:
Now that the PDK can read the environment.conf file, it can be updated to
display that the user, and be consumed by the Control Repo context. This commit:

* Adds a helper method to PDK::ControlRepo to create an an object to read/edit
  an arbitrary environment.conf file.  The helper method also adds the default
  settings as per Puppet documentation
* Updates the PDK::Config class to take in a PDK Context. The configuration
  settings then are mutated depending on the Context, for example, the
  'puppet.environment.*' settings now exist if the user is in a Control Repo,
  and won't if they're in a Module.

Note that because PDK::Config now depends on a PDK::Context object being
passed in (or default via PDK.context), any method calls to configuration
information (ala PDK.config) are not allowed and can cause circular
references and stack overflows. Thus there is a helper class to create the
environment.conf IniFile object.

Note that right now there is no project-level PDK configuration file therefore
the mock configuration RSpec context does not mock this and that the
PDK::Config.project_config method has no filename associated with it.  Future
commits may add this feature.

### DIFF
--- a/lib/pdk/context/control_repo.rb
+++ b/lib/pdk/context/control_repo.rb
@@ -10,12 +10,35 @@ module PDK
       def initialize(repo_root, context_path)
         super(context_path)
         @root_path = repo_root
+        @environment_conf = nil
       end
 
+      # @see PDK::Context::AbstractContext.pdk_compatible?
       def pdk_compatible?
         # Currently there is nothing to determine compatibility with the PDK for a
         # Control Repo. For now assume everything is compatible
         true
+      end
+
+      # The modulepath setting for this control repository as an array of strings. These paths are relative
+      # and may contain interpolation strings (e.g. $basemodulepath)
+      # @see https://puppet.com/docs/puppet/latest/config_file_environment.html#allowed-settings
+      # @return [Array[String]] The modulepath setting for this control repository
+      def module_paths
+        return @module_paths unless @module_paths.nil?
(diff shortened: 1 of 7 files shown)


lib/pdk/context/control_repo.rb
```diff
@@ -10,20 +10,51 @@ class ControlRepo < PDK::Context::AbstractContext
...
+        value = environment_conf['modulepath'] || ''
+        # We have to use a hardcoded value here because File::PATH_SEPARATOR is ';' on Windows.
+        # As the environment.conf is only used on Puppet Server, it's always ':'
+        # Based on - https://github.com/puppetlabs/puppet/blob/f3e6d7e6d87f46408943a8e2176afb82ff6ea096/lib/puppet/settings/environment_conf.rb#L98-L106
+        @module_paths = value.split(':')
+      end
+
+      # The relative module_paths that exist on disk.
+      # @see https://puppet.com/docs/puppet/latest/config_file_environment.html#allowed-settings
+      # @return [Array[String]] The absolute module paths
```
rodjek: This returns the array of relative paths, not absolute paths.

glennsarti: Whoops.  Fixed.

glennsarti:
Typo fixed.  Merging!

PR #858, opened 2020-02-22
(PDK-1618)(PDK-1613)(PDK-1616) Add Control Repo support to Validators

glennsarti:
~~Builds on #857~~ Merged

---

Now that Validators have a context to know whether they are operating in a
Module or Control Repo, the valdiator concrete classes need to be updated to
change their behaviour depending on the context.

* When in a Control Repo, previously module level patterns now need to be
  changed to each modulepath in the Control Repo.  The helper method
  `contextual_pattern` on the InvokableValidator class provides an easy way to
  do this
* Update the concrete validators to handle patterns which are Arrays. Previously
  many just assumed a String
* Update bundler helper to check for a Gemfile before processing binstubs.
  Previously this threw an obtuse error.  Now that Control Repos are supported
  it is common that they do not have a Gemfile.
* Updated `pdk validate` to support the controlrepo feature flag
* Updated `pdk validate` to support NOT having a Gemfile, which is common in
  control repos.
* Updated the ExternalCommandValidator to support being used in a non-Gemfile
  environment.  This meant searching for the external command not only as a
  binstub, but also in packaged or default bin directories

(PDK-1618) Add Feature Flags

(PDK-1618) Add controlrepo feature flag

(PDK-1613) Pass context into the Validators

(PDK-1616) Extend validators for a control-repo

(maint) Validators should use unique target list

---

TODO

- [x] Working on a real control-repo, with no Gemfile
- [x] Behind a featureflag
- [x] Acceptance tests green?
- [ ] Adhoc package testing
- [x] Fix duplicate target issue.

### DIFF
--- a/lib/pdk.rb
+++ b/lib/pdk.rb
@@ -47,7 +47,9 @@ module PDK
   end
 
   def self.available_feature_flags
-    @available_feature_flags ||= %w[].freeze
+    @available_feature_flags ||= %w[
+      controlrepo
+    ].freeze
   end
 
   def self.requested_feature_flags
--- a/lib/pdk.rb
+++ b/lib/pdk.rb
@@ -46,6 +46,19 @@ module PDK
     @context ||= PDK::Context.create(Dir.pwd)
   end
 
+  def self.available_feature_flags
+    @available_feature_flags ||= %w[].freeze
+  end
+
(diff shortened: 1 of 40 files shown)


glennsarti:
IT'S HAPPENING!!

PR #859, opened 2020-02-24
(PDK-1108) Add pdk set config command

glennsarti:
Previously the PDK::Config class could only read (get) values, it could not set
them easily, and lacked logic to deal with different value types, for example
if you set a deep hash value, then the save data method on the namespace was
not triggered.

This commit:

* Adds a PDK::Config.set method which takes into account nested namespaces,
  setting deep hashes, creates missing hash structures, and can add to, or
  explicitly set array, values.  All of this is necessary for the CLI component

* Adds tests for the different scenarios

---

Now that the PDK::Config class now has a set method, we can add a `set config`
CLI command so users can set configuration values.  Later commits will add the
`remove config` CLI command.

This commit:
* Adds the basic `pdk set config` command
* Adds type conversion using the `--type` parameter
* Adds tests for the CLI command

---

Previously many calls within the PDK used `.user` directly.  However now that
there is a formal get and set method, accessing these directly is no longer
required and is a private API.  This commit updates any references to these
private methods and changes them to use .get and .set appropriately.

---

TODO
- [x] Add PDK::Config.set
- [x] Add `pdk set config` command
- [x] Add tests (unit/acceptance)
- [x] Replace any internal PDK calls that set config to use the new set method e.g. `PDK.config.user['foo']['bar'] = 'baz'` to `PDK.config.set('foo.bar', 'baz')`

### DIFF
--- a/lib/pdk/cli/set/config.rb
+++ b/lib/pdk/cli/set/config.rb
@@ -1,0 +1,119 @@
+module PDK::CLI
+  module Set
+    module Config
+      ALLOWED_TYPE_NAMES = %w[array boolean number string].freeze
+
+      # :nocov:
+      def self.pretty_allowed_names
+        ALLOWED_TYPE_NAMES.map { |name| "'#{name}'" }.join(', ')
+      end
+      # :nocov:
+
+      def self.transform_value(type_name, value)
+        normalized_name = type_name.downcase.strip
+        unless ALLOWED_TYPE_NAMES.include?(normalized_name)
+          raise PDK::CLI::ExitWithError, _('Unknown type %{type_name}. Expected one of %{allowed}') % { type_name: type_name, allowed: pretty_allowed_names }
+        end
+
+        # Short circuit string conversions as it's trivial
+        if normalized_name == 'string'
+          raise PDK::CLI::ExitWithError, _('An error occured converting \'%{value}\' into a %{type_name}') % { value: value.nil? ? 'nil' : value, type_name: type_name } unless value.is_a?(String)
+          return value
+        end
+
+        begin
+          case normalized_name
+          when 'array'
+            convert_to_array(value)
+          when 'boolean'
+            convert_to_boolean(value)
+          when 'number'
+            convert_to_number(value)
+          else
+            value
+          end
+        rescue ArgumentError, TypeError
+          raise PDK::CLI::ExitWithError, _('An error occured converting \'%{value}\' into a %{type_name}') % { value: value.nil? ? 'nil' : value, type_name: type_name }
+        end
+      end
+
+      def self.convert_to_array(value)
+        return [] if value.nil?
+        value.is_a?(Array) ? value : [value]
+      end
+      private_class_method :convert_to_array
+
+      def self.convert_to_boolean(value)
+        string_val = value.to_s.strip.downcase
+
+        return true  if %w[yes true -1 1].include?(string_val)
+        return false if %w[no false 0].include?(string_val)
+
+        raise ArgumentError
+      end
+      private_class_method :convert_to_boolean
+
+      def self.convert_to_number(value)
+        float_val = Float(value)
+        # Return an Integer if this is actually and Integer, otherwise return the float
+        (float_val.truncate == float_val) ? float_val.truncate : float_val
+      end
+      private_class_method :convert_to_number
+
+      def self.run(opts, args)
+        item_name = (args.count > 0) ? args[0] : nil
+        item_value = (args.count > 1) ? args[1] : nil
+
+        opts[:type] = opts[:as] if opts[:type].nil? && !opts[:as].nil?
+        force = opts[:force] || false
+
+        # Transform the value if we need to
+        item_value = PDK::CLI::Set::Config.transform_value(opts[:type], item_value) unless opts[:type].nil?
+
+        raise PDK::CLI::ExitWithError, _('Configuration name is required') if item_name.nil?
+        raise PDK::CLI::ExitWithError, _('Configuration value is required. If you wish to remove a value use \'pdk remove config\'') if item_value.nil?
+
+        current_value = PDK.config.get(item_name)
+        raise PDK::CLI::ExitWithError, _("The configuration item '%{name}' can not have a value set.") % { name: item_name } if current_value.is_a?(PDK::Config::Namespace)
+
+        # If we're forcing the value, don't do any munging
+        unless force
+          # Check if the setting already exists
+          if current_value.is_a?(Array) && current_value.include?(item_value)
+            PDK.logger.info(_("No changes made to '%{name}' as it already contains value '%{to}'") % { name: item_name, to: item_value })
+            return 0
+          end
+        end
+
+        new_value = PDK.config.set(item_name, item_value, force: opts[:force])
+        if current_value.nil? || force
+          PDK.logger.info(_("Set initial value of '%{name}' to '%{to}'") % { name: item_name, to: new_value })
+        elsif current_value.is_a?(Array)
+          # Arrays have a special output format
+          PDK.logger.info(_("Added new value '%{to}' to '%{name}'") % { name: item_name, to: item_value })
+        else
+          PDK.logger.info(_("Changed existing value of '%{name}' from '%{from}' to '%{to}'") % { name: item_name, from: current_value, to: new_value })
+        end
+
+        # Same output as `get config`
+        $stdout.puts _('%{name}=%{value}') % { name: item_name, value: PDK.config.get(item_name) }
+        0
+      end
+    end
+  end
+
+  @set_config_cmd = @set_cmd.define_command do
+    name 'config'
+    usage _('config [name] [value]')
+    summary _('Set or update the configuration for <name>')
+
+    option :f, :force, _('Force the configuration setting to be overwitten.'), argument: :forbidden
+
+    option :t, :type, _('The type of value to set. Acceptable values: %{values}') % { values: PDK::CLI::Set::Config.pretty_allowed_names }, argument: :required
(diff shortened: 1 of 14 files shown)


spec/acceptance/set_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe 'pdk set' do
+  include_context 'with a fake TTY'
+
+  context 'when run outside of a module' do
+    describe command('pdk set') do
+      its(:exit_status) { is_expected.to eq 0 }
+      # Should show the command help
+      its(:stdout) { is_expected.to match(%r{pdk set \[subcommand\] \[options\]}) }
+      its(:stderr) { is_expected.to have_no_output }
+    end
```
rodjek: It'd be good to get some functional tests here as well

glennsarti: The unit tests cover A LOT of those use cases.  So it felt strange to duplicate them

lib/pdk/cli/set/config.rb
```diff
@@ -0,0 +1,120 @@
...
+      def self.convert_to_array(value)
+        return [] if value.nil?
+        value.is_a?(Array) ? value : [value]
+      end
+      private_class_method :convert_to_array
+
+      def self.convert_to_boolean(value)
+        string_val = value.to_s.strip.downcase
+
+        return true  if %w[yes true -1].include?(string_val)
```
scotje: Maybe include positive 1 here as well? I think that's more common than -1 in Unix-like environments.

lib/pdk/cli/set/config.rb
```diff
@@ -0,0 +1,120 @@
...
+        # Transform the value if we need to
+        item_value = PDK::CLI::Set::Config.transform_value(opts[:type], item_value) unless opts[:type].nil?
+
+        raise PDK::CLI::ExitWithError, _('Configuration name is required') if item_name.nil?
+        raise PDK::CLI::ExitWithError, _('Configuration value is required. If you wish to remove a value use \'pdk remove config\'') if item_value.nil?
+
+        current_value = PDK.config.get(item_name)
+        raise PDK::CLI::ExitWithError, _("The configuration item '%{name}' can not have a value set.") % { name: item_name } if current_value.is_a?(PDK::Config::Namespace)
+
+        item_value = item_value
```
scotje: Does this do something I'm not realizing? If so it should have an explanatory comment.

lib/pdk/cli/set/config.rb
```diff
@@ -0,0 +1,120 @@
...
+        end
+
+        new_value = PDK.config.set(item_name, item_value, force: opts[:force])
+        if current_value.nil? || force
+          PDK.logger.info(_("Set initial value of '%{name}' to '%{to}'") % { name: item_name, to: new_value })
+        elsif current_value.is_a?(Array)
+          # Arrays have a special output format
+          PDK.logger.info(_("Added new value '%{to}' to '%{name}'") % { name: item_name, to: item_value })
+        else
+          PDK.logger.info(_("Changed existing valuie of '%{name}' from '%{from}' to '%{to}'") % { name: item_name, from: current_value, to: new_value })
```
scotje:
```suggestion
          PDK.logger.info(_("Changed existing value of '%{name}' from '%{from}' to '%{to}'") % { name: item_name, from: current_value, to: new_value })
```

spec/acceptance/set_spec.rb
rodjek: Just a few cases to ensure the expected changes on disk which unit tests mock out. Set at system level, user level, project level, ensure the file gets created on disk at the appropriate path. Seems redundant but it covers cases like https://github.com/puppetlabs/pdk/pull/862

glennsarti: Will do

glennsarti:
Putting back to WIP until the acceptance tests are updated.

spec/acceptance/set_config_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    context 'with a setting that does not exist' do
+      describe command('pdk set config user.module_defaults.mock value') do
+        include_context 'with a fake answer file'
+
+        its(:exit_status) { is_expected.to eq 0 }
+        its(:stdout) { is_expected.to match(%r{user.module_defaults.mock=value}) }
+        its(:stderr) { is_expected.to match(%r{Set initial value of 'user.module_defaults.mock' to 'value'}) }
+
+        it_behaves_like 'a saved configuration file', "{\n  \"mock\": \"value\"\n}\n"
```
scotje: Is it worth having a JSON-specific shared behavior that parses the file contents so we don't have to serialize the whitespace and everything?

E.g.

```ruby
it_behaves_like 'a saved JSON configuration file', { "mock" => "value" }
```

spec/acceptance/set_config_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      describe command('pdk set config user.module_defaults.mock value') do
+        include_context 'with a fake answer file', 'mock' => []
+
+        its(:exit_status) { is_expected.to eq 0 }
+        its(:stdout) { is_expected.to match(%r{user.module_defaults.mock=\["value"\]}) }
+        its(:stderr) { is_expected.to match(%r{Added new value 'value' to 'user.module_defaults.mock'}) }
+
+        it_behaves_like 'a saved configuration file', "{\n  \"mock\": [\n    \"value\"\n  ]\n}\n"
+      end
+    end
```
scotje:
```suggestion
    context 'with an existing array setting, not forced' do
      describe command('pdk set config user.module_defaults.mock value') do
        include_context 'with a fake answer file', 'mock' => []

        its(:exit_status) { is_expected.to eq 0 }
        its(:stdout) { is_expected.to match(%r{user.module_defaults.mock=\["value"\]}) }
        its(:stderr) { is_expected.to match(%r{Added new value 'value' to 'user.module_defaults.mock'}) }

        it_behaves_like 'a saved configuration file', "{\n  \"mock\": [\n    \"value\"\n  ]\n}\n"
      end
    end
```

scotje: Is there a test that covers replacing an existing scalar setting?

spec/unit/pdk/cli/set/config_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+      def user_config
+        @user_config ||= PDK::Config::Namespace.new('user') { ; }
+      end
+
+      def system_config
+        @system_config ||= PDK::Config::Namespace.new('system') { ; }
+      end
+
+      def project_config
+        @system_config ||= PDK::Config::Namespace.new('system') { ; }
```
scotje: Should this be?

```suggestion
        @project_config ||= PDK::Config::Namespace.new('project') { ; }
```

glennsarti: Yes.  Yes it should!

lib/pdk/cli/set/config.rb
glennsarti: Nope...that's going!

lib/pdk/cli/set/config.rb
glennsarti: Fixed.

lib/pdk/cli/set/config.rb
glennsarti: Fixed

spec/acceptance/set_config_spec.rb
glennsarti: Changed the bulk of the tests to use JSON (ala Hash) matching and just one for file content.  I need that one to prove it's outputting "pretty" JSON instead of mini-fied JSON which is hard to read.

spec/acceptance/set_config_spec.rb
glennsarti: Added

spec/unit/pdk/cli/set/config_spec.rb
glennsarti: Fixed

spec/acceptance/set_spec.rb
glennsarti: Actually, this will be VERY tricky to do in acceptance.  I'd either:

1. Have to mock File.write calls to the expected places on disk

OR

2. Make the config classes write to a temp file location

Point 1 is already done in `spec/unit/pdk/util_spec.rb` once #862 is merged

Point 2 is what's being used in this test suite already except only for user files.  If I was to use temporary `system` and `project` locations then it still wouldn't have picked up the issue #862

If I don't mock, then I'd be overwriting _actual_ file content which is not expected when running acceptance. If I want to write to _actual_ disk locations, this would be appropriate in the package-testing which is using VMs/Containters, not the localhost.

---

So yeah, that kind of testing needs to go in package-testing.

glennsarti: @rodjek  I've added a new commit which adds some package-testing tests.

spec/unit/pdk/cli/set/config_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+
+    RSpec.shared_examples 'a un-settable setting error' do
+      it 'raises with failed conversion' do
+        expect { run }.to raise_error(PDK::CLI::ExitWithError, %r{can not have a value set})
+      end
+    end
+
+    RSpec.shared_examples 'a saved setting' do |setting_name, expected_value, force = nil|
+      it 'saves the setting' do
+        expect(pdk_config).to receive(:set).with(setting_name, Object, force: force).and_call_original
```
rodjek: nit: The idiomatic way to express a wildcard parameter in rspec-mocks is `anything` rather than `Object`

glennsarti: Fixed.

PR #862, opened 2020-02-27
(FIXUP) Make system config path absolute instead of relative on POSIX hosts

scotje:
Pretty much what it says on the tin.

### DIFF
--- a/spec/unit/pdk/util_spec.rb
+++ b/spec/unit/pdk/util_spec.rb
@@ -349,7 +349,7 @@ describe PDK::Util do
       end
 
       it 'returns a path inside the system opt directory' do
-        is_expected.to eq(File.join('opt', 'puppetlabs', 'pdk', 'config'))
+        is_expected.to match(%r{\A/opt/})
       end
     end
   end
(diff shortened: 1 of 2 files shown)


spec/unit/pdk/util_spec.rb
```diff
@@ -349,7 +349,7 @@
       end
 
       it 'returns a path inside the system opt directory' do
-        is_expected.to eq(File.join('opt', 'puppetlabs', 'pdk', 'config'))
+        is_expected.to match(%r{\A/opt/})
```
scotje: We can make this more specific if we want, but this now test now better matches it's description.

rodjek:
We should look at getting some acceptance tests going for this as well

glennsarti:
@rodjek Did you want the acceptance tests in this PR?

glennsarti:
I've added some tests in #859 9https://github.com/puppetlabs/pdk/pull/859/commits/a3b68dbcb8928d0bef3972f30bff4a30cc6222dc) Which should cover this.

PR #864, opened 2020-02-28
(MAINT) Bump package-testing nokogiri to ~> 1.10.8

scotje:
Does anyone remember if we were still pinning this for a reason? The last commit to touch that line was removing pinning inside the `update_module_spec` test.


scotje:
I'll do an adhoc package testing build and see what happens.

rodjek:
There wasn't a particular reason IIRC, those deps were all pinned just to avoid beaker fragility.

scotje:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/PDK/view/adhoc/job/platform_pdk_adhoc_pdk-int-sys-testing_adhoc/109/ was successful

PR #865, opened 2020-03-02
(MAINT) Bump nokogiri and rake for security advisories

scotje:


scotje:
Passed ad-hoc testing here: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/PDK/view/adhoc/job/platform_pdk_adhoc_pdk-int-sys-testing_adhoc/110/

PR #866, opened 2020-03-03
(PDK-1615) Add validator for environment.conf

glennsarti:
This commit adds a validator for the environment.conf file when being used
inside a control repository.  The validator complies with the rules set out in
https://puppet.com/docs/puppet/latest/config_file_environment.html.

This commit also adds tests for the validator setup, and various example file
content and expected validation results.

---

TODO

- [x] Should this be in the `puppet` validator group OR should it be in a separate validator group e.g. `controlrepo` ?

### DIFF
--- a/lib/pdk/validate.rb
+++ b/lib/pdk/validate.rb
@@ -9,8 +9,8 @@ module PDK
     autoload :ValidatorGroup, 'pdk/validate/validator_group'
 
     module ControlRepo
-      autoload :EnvironmentConfValidator, 'pdk/validate/control_repo/environment_conf_validator'
       autoload :ControlRepoValidatorGroup, 'pdk/validate/control_repo/control_repo_validator_group'
+      autoload :EnvironmentConfValidator, 'pdk/validate/control_repo/environment_conf_validator'
     end
 
     module Metadata
@@ -19,9 +19,11 @@ module PDK
       autoload :MetadataValidatorGroup, 'pdk/validate/metadata/metadata_validator_group'
     end
 
-    module YAML
-      autoload :YAMLSyntaxValidator, 'pdk/validate/yaml/yaml_syntax_validator'
-      autoload :YAMLValidatorGroup, 'pdk/validate/yaml/yaml_validator_group'
+    module Puppet
+      autoload :PuppetEPPValidator, 'pdk/validate/puppet/puppet_epp_validator'
+      autoload :PuppetLintValidator, 'pdk/validate/puppet/puppet_lint_validator'
+      autoload :PuppetSyntaxValidator, 'pdk/validate/puppet/puppet_syntax_validator'
+      autoload :PuppetValidatorGroup, 'pdk/validate/puppet/puppet_validator_group'
     end
 
     module Ruby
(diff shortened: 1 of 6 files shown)


lib/pdk/validate/puppet/puppet_environment_conf_validator.rb
```diff
@@ -0,0 +1,92 @@
...
+            )
+            return 1
+          end
+
+          is_valid = true
+          begin
+            env_conf = PDK::ControlRepo.environment_conf_as_config(target)
+
+            env_conf.resolve.each do |setting_name, setting_value|
+              setting_name = setting_name.slice(12..-1)
```
scotje: I'm guessing this is slicing off an `environment.` prefix? Adding a comment about that or even using `gsub` (performance implications should be pretty minor) would help clarify this for future us.

lib/pdk/validate/puppet/puppet_environment_conf_validator.rb
```diff
@@ -0,0 +1,92 @@
...
+              next if ALLOWED_SETTINGS.include?(setting_name)
+              # A hash indicates that the ini file has a section in it.
+              message = setting_value.is_a?(Hash) ? _("Invalid section '%{name}'") % { name: setting_name } : _("Invalid setting '%{name}'") % { name: setting_name }
```
scotje: Nit: I would prefer multiline if/else instead of a line this long.

```suggestion
              message = if setting_value.is_a?(Hash)
                _("Invalid section '%{name}'") % { name: setting_name }
              else
                _("Invalid setting '%{name}'") % { name: setting_name }
              end
```

spec/unit/pdk/validate/puppet/puppet_environment_conf_validator_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+          source:   validator.name,
+          state:    :passed,
+          severity: 'ok',
+        )
+        expect(return_value).to eq(0)
+      end
+    end
+
+    context 'when a target is provided that has all allowed settings and is valid' do
+      let(:target) { { name: 'environment.conf', content: "modulepath=foo\nmanifest=foo\nconfig_version=foo\nenvironment_timeout=0" } }
```
scotje: Maybe move `content` to it's own var/`let` block and use a heredoc to make it easier to maintain?

spec/unit/pdk/validate/puppet/puppet_environment_conf_validator_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+          source:   validator.name,
+          state:    :passed,
+          severity: 'ok',
+        )
+        expect(return_value).to eq(0)
+      end
+    end
+
+    context 'when a target is provided that has invalid settings and sections' do
+      let(:target) { { name: 'environment.conf', content: "modulepath=foo\nmanifest=foo\nconfig_version=foo\nenvironment_timeout=foo\ninvalid=true\n[invalid_section]\ninvalid=true" } }
```
scotje: Ditto: "Maybe move `content` to it's own var/`let` block and use a heredoc to make it easier to maintain?"

rodjek:
> Should this be in the puppet validator group OR should it be in a separate validator group e.g. controlrepo ?

I think this should be in a control repo validator group

lib/pdk/validate/puppet/puppet_environment_conf_validator.rb
glennsarti: Added a comment.

spec/unit/pdk/validate/puppet/puppet_environment_conf_validator_spec.rb
glennsarti: I managed to use a heredoc inside the hash which looks ok e.g.

``` ruby
      let(:target) do
        {
          name: 'environment.conf',
          content: <<-EOT
                   modulepath=foo
                   manifest=foo
                   config_version=foo
                   environment_timeout=0
                   EOT
        }
      end
```

spec/unit/pdk/validate/puppet/puppet_environment_conf_validator_spec.rb
glennsarti: See above.

lib/pdk/validate/puppet/puppet_environment_conf_validator.rb
glennsarti: Fixed as per your suggestion

glennsarti:
Moved it to it's own validator group `control-repo`

PR #867, opened 2020-03-03
(PDK-1632) Only show validators that are available in the given PDK Context

glennsarti:
Previously when a validator was created which was not valid in the PDK Context,
for example environment.conf validation in a Puppet Module, the validator would
not be invoked but a message would be logged saying that the validator had no
targets.  This is slightly confusing for users.

This commit changes the behaviour so that validators which are not valid in the
current PDK Context (Validator.is_valid_context?) are never run or have any
targets parsed.

This commit also updates the tests for these scenarios.


glennsarti:
Updated comment in validate_spec.rb as per @rodjek comments in slack.  Will merge once CI is green.

PR #883, opened 2020-05-19
(PDK-1653) Ensure template have access to metadata during update/convert

scotje:


scotje:
cc @DavidS @sanfrancrisko

DavidS:
I've verified that the issue reported is reproducible on current PDK master and that this PR resolves the problem using this test:

```
 
david@zion:~/git/pdk (1653_module_metadata)$ git diff
diff --git a/spec/acceptance/convert_spec.rb b/spec/acceptance/convert_spec.rb
index 803bcfb..27776ff 100644
--- a/spec/acceptance/convert_spec.rb
+++ b/spec/acceptance/convert_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper_acceptance'
 # Force the modules generated in these specs to use the public template repo to
 # avoid unnecessary changes in the metadata.json during convert (template-url
 # key).
-template_repo = 'https://github.com/puppetlabs/pdk-templates'
+template_repo = 'https://github.com/treydock/pdk-templates'
 pdk_convert_base = "pdk convert --template-url=#{template_repo}"
 
 describe 'pdk convert', module_command: true do
@@ -19,6 +19,11 @@ describe 'pdk convert', module_command: true do
     describe file('convert_report.txt') do
       it { is_expected.not_to be_file }
     end
+
+    describe file('spec/acceptance/nodesets/centos-6.yml') do
+      it { is_expected.to be_file }
+      its(:content) { is_expected.to include("docker_container_name: 'clean_module-el6'") }
+    end
   end
 
   context 'when run with --noop' do
```

We do not have any use of `module_metadata` in the `moduleroot` of the regular templates, so it's not easy to test for this.

Issue #886, opened 2020-06-03
Unit tests fail with can't modify frozen String (RuntimeError)


edergillian:
**Describe the bug**
When trying to run a unit test using `pdk test unit --tests <path>` on Windows, I get this error:
```
C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/pdk-1.18.0/lib/pdk/tests/unit.rb:96:in `tr!': can't modify frozen String (RuntimeError)
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/pdk-1.18.0/lib/pdk/tests/unit.rb:96:in `invoke'
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/pdk-1.18.0/lib/pdk/cli/test/unit.rb:80:in `block (2 levels) in <module:CLI>'
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/cri-2.10.1/lib/cri/command.rb:329:in `run_this'
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/cri-2.10.1/lib/cri/command.rb:269:in `run'
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/cri-2.10.1/lib/cri/command.rb:287:in `run'
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/cri-2.10.1/lib/cri/command.rb:287:in `run'
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/pdk-1.18.0/lib/pdk/cli.rb:68:in `run'
        from C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/pdk-1.18.0/exe/pdk:6:in `<top (required)>'
        from C:/PROGRA~1/PUPPET~1/DEVELO~1/private/ruby/2.4.10/bin/pdk:23:in `load'
[...]

PR #891, opened 2020-07-14
Don't attempt to modify a frozen string when parsing '--tests' paths

natemccurdy:
Fixes issue #886

Prior to this, when specifying the path(s) to a spec file to test with the `--tests` option, PDK was failing with:
> /pdk-1.18.0/lib/pdk/tests/unit.rb:96:in `tr!': can't modify frozen String (RuntimeError)

After this, `pdk test unit --path some/spec/file.rb` works and does not throw an error.


natemccurdy:
This could use some tests.

joshcooper:
Tests are failing due to https://github.com/rodjek/rspec-puppet/pull/797. You'll need to rebase your PR after the other PR is merged

PR #900, opened 2020-07-21
(MAINT) Bump version back to 1.19.0.pre

scotje:


scotje:
Failure was transient, I rebuilt that cell and it passed.

PR #903, opened 2020-08-03
update to Debian 10 and remove Debian 8

kenyon:
Update various usages of the Debian version to the current stable version. Also remove Debian 8, which is EOL.


daianamezdrea:
Hi @kenyon, this looks good! Thank you for your contribution!

PR #912, opened 2020-08-26
Add ability to set a forge token for release

logicminds:
* Previously when using the release command the user
    had to enter the forge token every time they released.
    This is annoying and is not compatible with CI systems.

    This commit allows the user or CI to set PUPPET_FORGE_TOKEN
    environment variable so they are not burdened with prompts.

    This also makes the default cli option the value of the environment variable.

    Note: when help is displayed the token is also displayed.


scotje:
Superseded by #913

PR #913, opened 2020-08-27
Add forge token env

scotje:
@logicminds it wouldn't let me push to your fork, but I took your commit and changed a couple things:

- renamed the env var to be more consistent with existing PDK env vars
- removed the default value from the CLI help and instead just fall-back to env if CLI value is empty
- use the PDK::Util env helpers
- added unit tests

Supersedes #912

### DIFF
--- a/lib/pdk/cli/release/publish.rb
+++ b/lib/pdk/cli/release/publish.rb
@@ -11,7 +11,7 @@ module PDK::CLI
     option nil, :'forge-upload-url', _('Set forge upload url path.'),
            argument: :required, default: 'https://forgeapi.puppetlabs.com/v3/releases'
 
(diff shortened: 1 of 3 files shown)


logicminds:
The cli argument is required and I suspect the cli will fail even if the env variable is set because no default is supplied.  Was this covered in your test?

scotje:
@logicminds `argument: required` on a CLI option just means that, if you specify that option, you must include a value. So, you can't just say `pdk release publish --forge-token`, you would have to say `pdk release publish --forge-token=foo`. But it does not mean that the option itself is required.

```bash
$ pdk release publish
pdk (ERROR): You must supply a Forge API token either via `--forge-token` option or PDK_FORGE_TOKEN environment variable.
```
```bash
$ pdk release publish --forge-token
publish: option requires an argument -- forge-token
```
```bash
$ PDK_FORGE_TOKEN=abc123 pdk release publish
pdk (INFO): Releasing jesse-testmod - from version 0.1.0
pdk (INFO): Uploading tarball to puppet forge...
```

PR #916, opened 2020-10-12
(MAINT) Switch default pdk-templates branch to `main`

scotje:


david22swan:
@scotje Thanks for the work 👍

PR #927, opened 2020-11-26
Adds AIX support when creating a new module

logicminds:
* Previously when creating a new module AIX was not
    an option when choosing supported operating systems.
    This resulted in many modules written with PDK to not support
    AIX.  This fixes that issue by adding AIX.

### DIFF
--- a/lib/pdk/module/metadata.rb
+++ b/lib/pdk/module/metadata.rb
@@ -54,6 +54,10 @@ module PDK
           'operatingsystem'        => 'windows',
           'operatingsystemrelease' => %w[2019 10],
         },
+        'AIX' => {
+          'operatingsystem'        => 'AIX',
+          'operatingsystemrelease' => %w[6.1 7.1 7.2],
+        },
       }.freeze
 
       DEFAULT_OPERATING_SYSTEMS = [


lib/pdk/module/metadata.rb
```diff
@@ -54,6 +54,10 @@ class Metadata
           'operatingsystem'        => 'windows',
           'operatingsystemrelease' => %w[2019 10],
         },
+        "AIX" => {
+          'operatingsystem'        => 'AIX',
+          'operatingsystemrelease' => %w[6.1 7.1 7.2]
+        }
```
michaeltlombardi, line 60: @logicminds it looks like you've got some rubocop errors, this should fix it:

```suggestion
        'AIX' => {
          'operatingsystem'        => 'AIX',
          'operatingsystemrelease' => %w[6.1 7.1 7.2],
        },
```

If you can amend your commit with these changes it looks good to me - you can do that by editing the file and then running the following commands:

```bash
git add lib/pdk/module/metadata.rb
git commit --amend --no-edit
git push --force
```

I would've just patched it and updated myself but I don't have write permissions to you fork :(

michaeltlombardi (changes requested):

lib/pdk/module/metadata.rb
logicminds: How should I fix these permission issues.  Just add all the people that want to fix for me?

scotje: There should be a checkbox on the right-hand side of the PR page that says "Allow edits by maintainers", I thought it was on by default though? (It's a per-PR setting as far as I can tell.)

logicminds: Fix was just pushed and correct style. There seems to be an unrelated error though with rubocop in local acceptance tests.

logicminds:
All requested changes have been addressed.

DavidS (approved):

PR #932, opened 2020-12-23
Add ability to generate functions

logicminds:
Before this can be merged the the following PR will need to be accepted.

https://github.com/puppetlabs/pdk-templates/pull/379

I added tests but they don't seem to actually get executed which is puzzling.  Guessing there is some undocumented tribal knowledge.

Some screen shots of the new function ality.  

Screen Shot 2020-12-23 at 12 41 39 PM


logicminds:
Looks like the tests are failing with unrelated rubocop errors on Rakefile and config.rb

lib/pdk/version.rb
```diff
@@ -1,4 +1,4 @@
 module PDK
-  VERSION = '1.19.0.pre'.freeze
+  VERSION = '1.19.1.pre'.freeze
```
DavidS, line 2:
```suggestion
  VERSION = '1.19.0.pre'.freeze
```

DavidS (approved):
Code looks good to me.

I would only suggest reverting the version number change as 1) the actual version will be determined during the release process and 2) this new feature would require a minor, not a patch-level bump.

I'm going to merge the associated pdk-templates changes so we can run this through our internal pipelines.

lib/pdk/version.rb
logicminds: I think the rake build task updates that automatically.  Anyways I set it back to 0.

PR #933, opened 2021-01-08
(MAINT) Fix up acceptance tests and configuration for recent releases of dependencies

scotje:


logicminds:
LGTM.  Is this the last commit to go into master?  When do we start merging into main?

scotje:
Yeah we can probably do the `main` switchover after this.

PR #942, opened 2021-02-16
Update lower puppet requirements bound when creating new modules

DavidS:
Since puppet5 is EOL, new modules should be created with up-to-date version
requirements.


DavidS:
See https://github.com/puppetlabs/pdk-templates/pull/404 for the second part of this work.

PR #954, opened 2021-03-03
Docs: Install on macOS with brew

rwaffen:
- cask command option is removed in 2021
- it is now a parameter of install or upgrade
- `brew cask install` is now `brew install --cask`


scotje (approved):
Thanks for the contribution!

Issue #958, opened 2021-03-10
Create Github Workflow for Static & Spec Tests


sanfrancrisko:
Create a Github Workflow for the PR Tests that will cover:

**On a hosted runner:**
- For each Ruby version (`2.5`, `2.7`):
  - `rubocop`
  - `spec:coverage`
  - `license_finder`
  - `test_pdk_as_library`
  - `spec`

jpogran:
We should target `windows-latest` and `ubuntu-latest` in hosted as a early warning system.

For acceptance testing, there is little difference (for our purposes) between 2016 and 2019, so we can limit to just that version and simplify things. For bonus points we could add a recent Windows 10 to test a desktop scenario, but I'm not sure what GCP images we have available.

I'm less certain about linux, I will check analytics but I think our most popular there is RHEL. I think ubuntu 20.04 should be good for the moment, and we should add as we get more info.

sanfrancrisko:
Sounds good. We'll go for parity with what we have at the moment, by the sounds of it.

I'll also update the description to plug in to Honeycomb with metrics.

Regarding Windows 10 (or any desktop Windows flavour), it seems to be missing from GCP's list of available images. The provision_service will be gaining support for AWS and Azure at some stage in the future, so I suspect the path of least resistance is to wait until that implementation and then add it in for ourselves.

jpogran:
Updated ticket description with target implementation

sanfrancrisko:
An additional piece of work spurred from this was to refactor the [validator](https://github.com/puppetlabs/provision_service/blob/11f5b7691b03fa9c4a131641e68cead9a5296d1b/backend/lib/utilities/validator.rb) in the [`provision_service`](https://github.com/puppetlabs/provision_service) to allow the `pdk` repo to provision resources in GCP.

See: https://github.com/puppetlabs/provision_service/pull/280

sanfrancrisko:
Given the complexities discovered during implementation for the Acceptance Test workflows, I'll raise separate tickets to cover:

- Acceptance tests on Linux (Hosted Runner on Github)
- Acceptance tests on Windows (GCP provisioned image via [provision_service](https://github.com/puppetlabs/provision_service) - see https://github.com/puppetlabs/provision_service/pull/283)

PR #1030, opened 2021-03-24
(GH-958) Add Spec tests for Ruby 2.5, 2.7 on Github Actions

sanfrancrisko:
This PR adds the following test / coverage suites to a
Github Actions workflow on both Ruby 2.5 and 2.7:

- `rubocop`
- `spec:coverage`
- `license_finder`
- `test_pdk_as_library`
- `spec`

Closes: #958

### DIFF
--- a/.github/workflows/spec.yml
+++ b/.github/workflows/spec.yml
@@ -1,0 +1,63 @@
+name: "Static, Spec & License Tests"
+
+on: [pull_request]
+
+jobs:
+  StaticAndSpecTests:
+    name: "Static & Spec Tests (Ruby ${{ matrix.ruby_version }})"
+    env:
+      BUNDLE_JOBS: 4
+      BUNDLE_WITHOUT: "development"
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby_version: ['2.5', '2.7']
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout Source
+      uses: actions/checkout@v2
+
+    - name: "Activate Ruby ${{ matrix.ruby_version }}"
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby_version }}
+        bundler-cache: true
+
+    - name: Rubocop
+      run: bundle exec rake rubocop
+
+    - name: 'Spec:Coverage'
+      run: bundle exec rake spec:coverage
+
+    - name: 'Test PDK as Library'
+      run: bundle exec rake test_pdk_as_library
+
+    - name: 'Upload CodeCov Report'
+      uses: codecov/codecov-action@v1
+      with:
+        directory: coverage
+        fail_ci_if_error: true
+
+  LicenseFinder:
+    name: "License Finder (Ruby ${{ matrix.ruby_version }})"
+    env:
+      BUNDLE_JOBS: 4
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby_version: ['2.5', '2.7']
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout Source
+      uses: actions/checkout@v2
+
+    - name: "Activate Ruby ${{ matrix.ruby_version }}"
+      uses: ruby/setup-ruby@v1
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/spec.yml
```diff
@@ -0,0 +1,38 @@
...
+      run: bundle exec rake spec:coverage
+
+    - name: 'License Finder'
+      run: bundle exec rake license_finder
+
+    - name: 'Test PDK as Library'
+      run: bundle exec rake test_pdk_as_library
+
+    - name: 'Spec'
+      run: bundle exec rake spec
```
da-ar, line 38: I believe this step can be removed as is should be the same as `bundle exec rake spec:coverage` just without the coverage information.

.github/workflows/spec.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        ruby-version: ${{ matrix.ruby_version }}
+        bundler-cache: true
+        without: "development"
+
+    - name: Rubocop
+      run: bundle exec rake rubocop
+
+    - name: 'Spec:Coverage'
+      run: bundle exec rake spec:coverage
```
da-ar, line 29: There's a silent failure in this step for connecting to coveralls. With us still being connected with travis, I'm not sure if this is something that can be changed?

.github/workflows/spec.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+    - name: Checkout Source
+      uses: actions/checkout@v2
+
+    - name: "Activate Ruby ${{ matrix.ruby_version }}"
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby_version }}
+        bundler-cache: true
+        without: "development"
```
da-ar, line 23: there's a warning regarding `without` in the output https://github.com/puppetlabs/pdk/pull/1030/checks#step:3:1 so this may not be doing what is expected.

jpogran: You'll want to add this env declaration:

```
    env:
      BUNDLE_JOBS: 4
      BUNDLE_WITHOUT: "development"
```

.github/workflows/spec.yml
sanfrancrisko: Good point, I'll maybe look in to the Coveralls Github Action: https://github.com/marketplace/actions/coveralls-github-action

.github/workflows/spec.yml
```diff
@@ -0,0 +1,38 @@
+name: "Static & Spec Tests"
+
+on: [pull_request, workflow_dispatch]
```
sanfrancrisko, line 3: @jpogran , @da-ar Thoughts on this being executable manually? Do we want to restrict to PRs only? I included `workflow_dispatch` as I thought it might be handy for us to trigger the job manually if we ever wanted to.

.github/workflows/spec.yml
sanfrancrisko: ...but yeah, it may not like this being reported in from two sources simultaneously

.github/workflows/spec.yml
jpogran: my two cents is no, no manual execution. i can't imagine the scenario at the moment

.github/workflows/spec.yml
sanfrancrisko: Given we decided to go for the [codecov Github Action](https://github.com/puppetlabs/pdk/pull/1030#issuecomment-807061849), we can safely have coveralls deprecated the same time we bin `.travis.yml`

.github/workflows/spec.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    - name: "Activate Ruby ${{ matrix.ruby_version }}"
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby_version }}
+        bundler-cache: true
+
+    - name: 'License Finder'
+      run: bundle exec rake license_finder
+
```
jpogran, line 64: nit: no end newline

jpogran (approved):

da-ar (approved):
LGTM

PR #1032, opened 2021-03-24
(MAINT) Remove 'Feature Request' issue template

sanfrancrisko:
We are moving feature requests to Discussions and therefore should remove the ability to raise features directly against the repo, bypassing the Discussions route.


jpogran:
Maybe we should keep the template and point to the discussions feature category like we are doing in the contributing.md just in case the contributor doesn't read the doc before opening an issue. Worst case, they open an issue and we convert it

sanfrancrisko:
Closing and will reopen another PR with the suggestion you made @jpogran

PR #1081, opened 2021-06-03
(SPIKE) - Pin `childprocess` to 4.0.0

david22swan:
Prior to this commit when attempting to spin up a child process from inside the pdk in a github actions runner a launch error is raised  due to a bug in child process. This commit raises the pin to v4.0.0 which includes the patch for this bug.

### DIFF
--- a/lib/pdk/cli/exec/command.rb
+++ b/lib/pdk/cli/exec/command.rb
@@ -27,7 +27,7 @@ module PDK
           @argv = argv
 
           @process = ChildProcess.build(*@argv)
-          @process.leader = true
+          # @process.leader = true
 
           @stdout = Tempfile.new('stdout', mode: TEMPFILE_MODE).tap { |io| io.sync = true }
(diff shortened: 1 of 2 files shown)


michaeltlombardi:
@jpogran do we have a standard for commit message prepending? Been using `(maint)` to indicate "no ticket" not "maintenance commit" - I don't see a way to indicate a breaking change in the labels, is there one?

jpogran:
This should be a ticketed item. There is a bug that needs to be fixed and we're upping a dependency 3 major versions. This will also need to run through acceptance testing in jenkins.

michaeltlombardi:
We can write up a decent ticket after we confirm that this fixes the GHA issue - for now, we just wanted to get a way to update the gemspec in the temporarily vendored puppetized DSC module in puppetlabs/Puppet.dsc#159 so we could validate.

@david22swan can you move this into draft mode?

sanfrancrisko:
Will be fixed as part of #1084

Issue #1083, opened 2021-06-03
Error when spinning up a child process from inside of ruby


david22swan:
When running the Puppet.Dsc acceptance tests on GHA we hit an error during the setup for the first test.
This error occurs whenever the pdk is called upon in order to create a module and states that there has been a `ChildProcess::LaunchError` due to our access having been denied. 

The exact error that we have been getting can be seen [here](https://github.com/puppetlabs/Puppet.Dsc/runs/2730261890?check_suite_focus=true#step:6:44).
And the PR on which we were working when we discovered this bug can be shown [here](https://github.com/puppetlabs/Puppet.Dsc/pull/159).

A conversation between @michaeltlombardi  the inimitable @glennsarti, who went out of his way to try a minimal local repro using the GHA docker image, discovered the problem seems to be this line in the PDK, which sets `process.leader`:

https://github.com/puppetlabs/pdk/blob/2b03b72ed71be20bf77b415b5709265164ce88dc/lib/pdk/cli/exec/command.rb#L30

In his manual testing, attempting to start a process in GHA with this flag breaks

```ruby
p = ChildProcess.build('ls')
p.leader = true
p.start
p.wait
puts p.exit_code
# Access is denied
```

But with it removed, no error is returned:

```ruby
p = ChildProcess.build('ls')
p.start
p.wait
puts p.exit_code
```

We had opened a PR on the pdk in order to try and implement two possible solutions, shown [here](https://github.com/puppetlabs/pdk/pull/1081).

The first of these solutions was to move the pin of the `childprocess` gem from `v0.7.1` to `v4.0.0` in order to bring in the patch shown [here](https://github.com/enkessler/childprocess/pull/167/commits), however this did not resolve the issue as the error still occured.

The second of our attempted solutions was to comment out the above linked line:
https://github.com/puppetlabs/pdk/blob/2b03b72ed71be20bf77b415b5709265164ce88dc/lib/pdk/cli/exec/command.rb#L30

But this did not work either as the change somehow caused a dependency error in the bundle install as shown [here](https://github.com/puppetlabs/Puppet.Dsc/runs/2737918820?check_suite_focus=true#step:6:399)

A draft for the experimental branch that we used is currently open can can be found here: #1081 

We are hoping for some help in finding a fix for this as it is a major blocker in moving our CI off of appveyor.

jpogran:
Attempted to run locally, works fine.
Attempted to run in a windows container, works fine.
[...]

PR #1084, opened 2021-06-09
(GH-1083) Bump childprocess to '~> 4.0.0'; Disable @process.leader

sanfrancrisko:
Prior to this commit, it was observed on Windows Server 2016, 2019
Github Actions instances that invoking the pdk would result in
ACCESS_DENIED errors.

When the ChildProcess.leader parameter is set to true, it will set
the `CREATE_BREAKAWAY_FROM_JOB` and `JOB_OBJECT_LIMIT_BREAKAWAY_OK`
flags in the Win32. This is not required in Windows > Server 2008 / 7
and can potentially cause issues when Task Scheduler attempts to
schedule a job that invokes a program with insufficient privileges.

This commit also bumps `childprocess` to `~> 4.0.0` and `hitimes`
to `2.0.0` to resolve dependency issues after bumping `childprocess`

Closes: #1083

### DIFF
--- a/lib/pdk/cli/exec/command.rb
+++ b/lib/pdk/cli/exec/command.rb
@@ -27,7 +27,17 @@ module PDK
           @argv = argv
 
           @process = ChildProcess.build(*@argv)
-          @process.leader = true
+          # https://github.com/puppetlabs/pdk/issues/1083:
+          # When @process.leader is set, childprocess will set the CREATE_BREAKAWAY_FROM_JOB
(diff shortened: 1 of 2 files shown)


jpogran (approved):

sanfrancrisko:
Appears dependency resolution is timing out on Windows systems w/ Ruby 2.5. Cannot reproduce on macOS - will try to do so on an ABS provisioned Windows host

jpogran (approved):
Approving as it was both manually and automatically tested as detailed #1083

PR #1089, opened 2021-06-10
(MAINT) Update bundler on all AppVeyor CI jobs

sanfrancrisko:


jpogran:
We've investigated this as far as we can in appveyor, the errors are down to some bundler permissions error or some complexity with the partially installing msys2 suite. The appveyor suite has never completely worked since we took over the pdk. In comparison, we have working examples in Github Actions and have working manual runs so we accept the risk in merging this as is.

PR #1092, opened 2021-06-15
(PDK-1085) Fail gracefully when no unit tests available

da-ar:
Avoids throwing when `--parallel` and `--format` are used together and there are no unit tests available


jpogran (approved):
codecov check has nothing to do with the change here

Issue #1113, opened 2021-06-28
pdk release: `--forge-token` is `true` instead of string


michaeltlombardi:
**Describe the bug**
While trying to release updated versions of the `dsc-*` modules, discovered a bug where the forge token is somehow turning into `true` instead of the passed string.

```
pdk (INFO): Releasing dsc-octopusdsc - from version 4.0.934-0-2
pdk (INFO): Uploading tarball to puppet forge...
C:/Program Files/Puppet Labs/DevelopmentKit/private/ruby/2.4.10/lib/ruby/gems/2.4.0/gems/pdk-2.1.1/lib/pdk/module/release.rb:160:in `+': no implicit conversion of true into String (TypeError)
```

https://github.com/puppetlabs/pdk/blob/dca0eb783de78d11007b07d7734ebe95f4597447/lib/pdk/module/release.rb#L160

**To Reproduce**
On Windows 10 with PDK 2.1.1
```powershell
pdk release --forge-token=my_token_here --skip-changelog --skip-validation --skip-documentation --skip-dependency --force
```

**Expected behavior**
I expect the forge token passed at the commandline (in the example repro code, `my_token_here` to be used for bearer auth in a Forge API call.

**Additional context**
- PDK Installation Method: Chocolatey
- PDK Version: 2.1.1
- Windows 10

sanfrancrisko:
Can reproduce this locally. I believe the `--forge-token` arg value is being mangled somewhere in the [`Cri::Command.run`](https://www.rubydoc.info/gems/cri/Cri/Command#run-instance_method) method (or further down). 

At this point it's the string value of what is passed to `--forge-token`:
https://github.com/puppetlabs/pdk/blob/dca0eb783de78d11007b07d7734ebe95f4597447/lib/pdk/cli.rb#L69

But it then appears to be mangled in to a boolean somewhere:

```ruby
   #<Cri::OptionDefinition:0x00007f84ac0f9fc8
     @argument=:forbidden,
     @block=nil,
     @default=false,
     @desc="Set Forge API token.",
     @hidden=false,
     @long="forge-token",
     @multiple=false,
     @short=nil,
     @transform=nil>,
```

Wasn't able to quickly determine where this is. Also tried a few things to see if we were setting this argument incorrectly, but it didn't yield any results.

Will need to dig in further.

Issue #917, opened 2020-10-21
pdk release fails to build without token


logicminds:
**Describe the bug**
The pdk release command does not perform a release because the token for the forge was not entered.  There was not a chance for the token to be entered because I answered 'N' to publishing to the forge.  So even If I did have the token it would never be used.

**To Reproduce**
```
[Q 1/6] Do you want to run the module validation ?
--> Yes

[Q 2/6] Do you want to run the automatic changelog generation ?
--> Yes

[Q 3/6] Do you want to set the module version ?
--> no

[Q 4/6] Do you want to run the dependency-checker on this module?
--> no

[Q 5/6] Do you want to update the documentation for this module?
--> no

[Q 6/6] Do you want to publish the module on the Puppet Forge?
--> no

pdk (ERROR): Missing forge-token option
```

**Expected behavior**
That it would finish the command without error or at least a different error.

**Additional context**
PDK 1.18.1

sanfrancrisko:
@logicminds After some further testing, we realised that this issue wasn't resolved in `2.1.1`. There's a fix going in to the next release (`2.2.0`) - apologies for the incomplete fix

PR #1121, opened 2021-07-06
(GH-1113) (GH-917) Fix forge-token handling

sanfrancrisko:
This PR contains two fixes:

## (GH-1113) Fix incorrect opt type for `forge-token`

Prior to this fix, the `forge-token` option defined in the PDK::CLI's
`release` subcommand did not have the `argument:` attribute set.
This meant that the underlying Cri lib treated it as a boolean option
and the forge token value was set to `true` and passed to the
`PDK::Module::Release` object.

This fix sets the `argument:` attribute with the value `:required`,
which emulates the way it is set in the `publish` subcommand.

Closes: #1113

------------------

## (GH-917) Fix incorrect 'Missing forge-token' error throwing

Prior to this fix, the `forge-token` option defined in the PDK::CLI's
`release` subcommand did not have the `argument:` attribute set.
This meant that the underlying Cri lib treated it as a boolean option
and the forge token value was set to `true` and passed to the
`PDK::Module::Release` object.

This fix sets the `argument:` attribute with the value `:optional`,
which allows it to be optional but treated as an arg that needs the
value extracted after `=`

Resolves: #917

### DIFF
--- a/lib/pdk/cli/release.rb
+++ b/lib/pdk/cli/release.rb
@@ -146,6 +146,7 @@ module PDK::CLI
         opts[:'skip-changelog'] = !answers['changelog']
         opts[:'skip-dependency'] = !answers['dependency']
         opts[:'skip-documentation'] = !answers['documentation']
+        opts[:'skip-publish'] = !answers['publish']
 
         prepare_version_interview(prompt, opts) if answers['setversion']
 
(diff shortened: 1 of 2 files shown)


sanfrancrisko:
- [ ] Ensure the forge token is [still optional for the `new module` interview process](https://github.com/puppetlabs/pdk/commit/2e70d47c085af01339ef584f5cebeb634cdb01a1)

jpogran (approved):

da-ar (approved):
veriried locally, appears to work as expected

Issue #1115, opened 2021-06-30
Update json_pure gem to latest


jpogran:
Latest version of json_pure resolves argument warnings in ruby 2.7.

We can't update this yet inside the packaging but can update this in the gem if we introduce a conditional.

PR #1124, opened 2021-07-13
(GH-1115) Bump `json_pure` to `~> 2.5.1` on Ruby `>= 2.7`

sanfrancrisko:
See also https://github.com/puppetlabs/pdk-vanagon/pull/292

Closes: #1115


sanfrancrisko:
## Tests:
- [ ] `bundle install` using Ruby `< 2.7` installs `json_pure 2.1.0`
- [ ] `bundle install` using Ruby `>= 2.7` installs `json_pure 2.5.1`
- [ ] Removes warnings when running with `json_pure 2.5.1` on Ruby `2.7`
- [ ] `pdk validate metadata` (using `json_pure 2.5.1`) does not flag issues for valid `metadata.json`
- [ ] `pdk validate metadata` (using `json_pure 2.5.1`) flags invalid `metadata.json`
------
### `json_pure` `2.5.1` on Ruby `2.7`:
```sh
~/bin/pdk validate metadata
pdk (WARN): This module is compatible with an older version of PDK. Run `pdk update` to update it to your version of PDK.
pdk (INFO): Using Ruby 2.7.3
pdk (INFO): Using Puppet 7.8.0
┌ [✔] Running metadata validators ...
├── [✔] Checking metadata syntax (metadata.json tasks/*.json).
└── [✔] Checking module metadata style (metadata.json).
```

jpogran (approved):

RandomNoun7:
Hi Content Team. I don't think this PR actually worked for at least one use case. When specifying PDK as a dependency in a module's Gemfile, the version of json_pure added to the Bundle is still 2.1.0. I think the reason is that Rubygems doesn't know how to add conditional runtime dependencies. You can see in the [Rubygems page for pdk](https://rubygems.org/gems/pdk/versions/2.2.0) that the json_pure dependency is still listed as 2.1.0

The Puppet team ran into this issue and had to use a workaround as well because of the Rubygems not doing conditionals problem https://tickets.puppetlabs.com/browse/PUP-10797

The only thing I know worked in my testing was to add a specific entry for json_pure to the Gemfile status specifically that it must be >= 2.3.0.

I also found that just removing PDK from my Gemfile worked, and I don't know why it was there in the first place, but the fact remains that I don't think Rubygems has picked up the fix in this PR as intended.

PR #1177, opened 2022-06-24
Fixing deprecated message on api_types.rb

davidsandilands:
When installing PDK as a rubygem on windows the following error occurs "warning: Using the last argument as keyword parameters is deprecated"

Updating with ** as per 
https://docs.knapsackpro.com/2021/fix-warning-using-the-last-argument-as-keyword-parameters-is-deprecated resolves this and PDK seems to work successfully


chelnak (approved):
Looks fine! Thanks 🙏

PR #1189, opened 2022-11-04
Adding more to description of puppet-syntax

davidsandilands:
As per https://github.com/voxpupuli/puppet-syntax description, it does templates and hiera yaml too


chelnak (approved):
Thank you @davidsandilands

PR #1203, opened 2023-01-26
(CONT-479) Bump nokogiri

chelnak:
Nokogiri < v1.13.6 does not type-check all inputs into the XML and HTML4 SAX parsers. For CRuby users, this may allow specially crafted untrusted inputs to cause illegal memory access errors (segfault) or reads from unrelated memory.

This commit bumps the nokogiri gem used in package testing to 1.13.6.


jordanbreen28 (approved):
LGTM

PR #1209, opened 2023-02-14
(maint) Include known issue for top_scope_facts

pmcmaw:


docs/pdk_known_issues.md
```diff
@@ -2,6 +2,20 @@
 
 ## PDK 2.6.1
 
+### Running autocorrect on puppet-lint top_scope_facts
+
+Currently there is an issue with a third party app that the pdk currently pulls in. When using the autocorrect functionality, it introduces issues in the codebase.
```
GSPatton, line 7: should we specify what the third party app with the issue is?

pmcmaw: suppose it would do no harm, I have also noticed I should probably refer to it as a plugin rather than an app 👍

chelnak (approved):

PR #1211, opened 2023-02-15
(CONT-370) Remove i18n support

chelnak:
This change removes i18n support from PDK as it is no longer required.

Additionally we have also seen a small performance improvement as a result of this work.


lib/pdk/cli.rb
```diff
@@ -49,22 +49,21 @@ def self.anonymised_args
     end
   end
 
-  def self.deprecated_runtime?
-    Gem::Version.new(RbConfig::CONFIG['ruby_version']) < Gem::Version.new('2.4.0')
-  end
+  # def self.deprecated_runtime?
```
pmcmaw, line 52: This is commented out, should it be removed?

jordanbreen28 (approved):

GSPatton (approved):
this looks good to me

PR #1215, opened 2023-02-24
(CONT-669) Use bundle info command

chelnak:
Prior to this change `github_changelog_generator_available!` would call `bundle show github_changelog_generator` to check for the existence of the gem.

This command has been deprecated deprecated which causes the following warning to be displayed when a user runs a `release` command

`[DEPRECATED] use bundle info github_changelog_generator instead of `bundle show github_changelog_generator

This change replaces the `show` command with `info` as suggested.


GSPatton (approved):
looks good

Issue #1210, opened 2023-02-14
Error in pdk release publish


coreone:
**Describe the bug**
When running `pdk release publish`, a Ruby error occurs when trying to call the `URI` method:

```sh
pdk (INFO): Releasing module-name - from version 4.0.0
pdk (INFO): Uploading tarball to puppet forge...
Traceback (most recent call last):
	10: from /usr/local/bin/pdk:23:in `<main>'
	 9: from /usr/local/bin/pdk:23:in `load'
	 8: from /var/lib/gems/2.7.0/gems/pdk-2.6.1/exe/pdk:6:in `<top (required)>'
	 7: from /var/lib/gems/2.7.0/gems/pdk-2.6.1/lib/pdk/cli.rb:71:in `run'
	 6: from /var/lib/gems/2.7.0/gems/cri-2.15.11/lib/cri/command.rb:316:in `run'
	 5: from /var/lib/gems/2.7.0/gems/cri-2.15.11/lib/cri/command.rb:316:in `run'
	 4: from /var/lib/gems/2.7.0/gems/cri-2.15.11/lib/cri/command.rb:298:in `run'
	 3: from /var/lib/gems/2.7.0/gems/cri-2.15.11/lib/cri/command.rb:362:in `run_this'
	 2: from /var/lib/gems/2.7.0/gems/pdk-2.6.1/lib/pdk/cli/release/publish.rb:47:in `block (2 levels) in <module:CLI>'
	 1: from /var/lib/gems/2.7.0/gems/pdk-2.6.1/lib/pdk/module/release.rb:91:in `run'
/var/lib/gems/2.7.0/gems/pdk-2.6.1/lib/pdk/module/release.rb:158:in `run_publish': undefined method `URI' for #<PDK::Module::Release:0x000055c4eac01340> (NoMethodError)
```

**To Reproduce**

* Download version 2.6.1 of `pdk`
* cd to a directory containing a pdk-configured repository
* Run `pdk release publish --skip-changelog --skip-dependency --skip-documentation --force --forge-token='abc123'`

**Expected behavior**
This particular example should return an access denied error due to the use of a fake forge token. Using a real token should publish the module to the forge.

**Additional context**
- Your PDK installation method: Gemfile (so, this is run with `bundle exec`)
  - (If via Rubygems, please include your Ruby version `ruby -v`): `ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux-gnu]`
[...]

PR #1220, opened 2023-03-10
(GH-1210) Require uri

chelnak:
As reported in #1210, in some situations pdk will fail to publish a module due to a missing require for URI.

For this specific occurance we will eventually move to the forge SDK which will elimiate the issue. However, due to some dependency constraints we cannot to that yet.

This change simply adds a new require for uri, which will solve the issue described above until we are able to migrate to the Forge SDK.


LukasAud (approved):
LGTM

PR #1230, opened 2023-03-15
(CONT-755) Ruby 3 / Puppet 8 support

chelnak:
This PR contains a number of changes that ensure PDK is compatible with Ruby 3 / Puppet 8.


Rakefile
```diff
@@ -1,6 +1,6 @@
 require 'bundler/gem_tasks'
 require 'puppet_litmus/rake_tasks' if Bundler.rubygems.find_name('puppet_litmus').any?
-require 'puppetlabs_spec_helper/tasks/fixtures' if Bundler.rubygems.find_name('puppetlabs_spec_helper').any?
+# require 'puppetlabs_spec_helper/tasks/fixtures' if Bundler.rubygems.find_name('puppetlabs_spec_helper').any?
```
david22swan, line 3: should this e removed instead of commented?

david22swan (approved):
One small comment but otherwise LGTM

lib/pdk/config/analytics_schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "definitions": {},
-  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$schema": "http://json-schema.org/draft/schema#",
```
jordanbreen28, line 3: This URL returns 404, where as the original `http://json-schema.org/draft-06/schema#` downloads the schema.
Just checking this references the new file that you created `lib/pdk/config/task_schema.json` instead?

chelnak: This is a really good catch @jordanbreen28.

It looks to me that there was some confusion with what the schema url would actually be. A change was made to the json-schema gem that made the validator check for the url in this change... however it actually turned out that it was released under the original url.

[This pr](https://github.com/voxpupuli/json-schema/pull/388) is the fix. I'm going to help get that merged now and will be able to correct the url in this PR.

.gitignore
```diff
@@ -84,3 +84,7 @@ Gemfile.local
 # generated artifacts from rake tasks
 /output
 modules
+
+# TEMP
+/.rubocop.yml
+/.rubocop_todo.yml
```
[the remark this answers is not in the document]
chelnak: nope!

PR #1245, opened 2023-05-04
pdk.md: update links

kenyon:


chelnak (approved):
Thanks for this @kenyon

PR #1247, opened 2023-05-09
(MAINT) Fix package tests

chelnak:
This changes includes fixes for the post packaging smoke tests.

### DIFF
--- a/package-testing/spec/package/update_module_spec.rb
+++ b/package-testing/spec/package/update_module_spec.rb
@@ -3,15 +3,13 @@ require 'open-uri'
 require 'json'
 
 modules = [
-  'puppetlabs/puppetlabs-motd',
-  'puppetlabs/puppetlabs-concat',
-  'puppetlabs/puppetlabs-inifile'
+  'puppetlabs/puppetlabs-motd'
 ]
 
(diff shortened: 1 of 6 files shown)


package-testing/spec/package/update_module_spec.rb
```diff
@@ -3,15 +3,13 @@
 require 'json'
 
 modules = [
-  'puppetlabs/puppetlabs-motd',
-  'puppetlabs/puppetlabs-concat',
-  'puppetlabs/puppetlabs-inifile'
+  'puppetlabs/puppetlabs-motd'
```
GSPatton, line 6: what's the rationale behind only testing against motd?

package-testing/spec/test_hosts.yml
GSPatton, line 1: is this needed?

chelnak: Good spot

package-testing/spec/package/update_module_spec.rb
chelnak: It reduces the surface of issues. MOTD changes are less frequent than the others.

GSPatton (approved):

PR #1283, opened 2023-10-25
(CAT-270) Adjusting expected Puppet/Ruby versions

LukasAud:
During a PDK release, it was noticed that there were several failures related to outdated Puppet/Ruby version expectations.

This commit aims to update those expectations so that the integration tests look for the correct values.


david22swan (approved):
LGTM

Issue #1224, opened 2023-03-13
Executable files in a PDK template do not end up as executable in the module


chelnak:
Issue moved from [PDK-1803](https://tickets.puppetlabs.com/browse/PDK-1803)

## The Problem

I have a PDK template where the moduleroot/ has a folder with scripts that are marked executable.

```
~/src/pdk-template$  ls -l moduleroot/ci_scripts
total 8
-rwxr-xr-x@ 1 nate  nate  287 Mar 10 15:34 validate_things.sh
```

When I build a module based on this PDK template, the validate_things.sh script does not get created as executable. Also when updating a module based on this template, the script remains not executable.

## What is expected

A file that's executable in a PDK template (either in moduleroot or moduleroot_init) should be created/updated by the PDK as executable in the module.

## What actually happens

Executable files are created without the execute bit set, making them non-executable.
This also happens when pdk update'ing a module.

natemccurdy:
# Reproduction

Here’s a reproduction case. It’s a new PDK module based on a fork of PDK templates that includes an executable file:
* https://github.com/natemccurdy/pdk-templates/blob/PDK-1803/reproduction/moduleroot/validate_things.sh

 
```
~ $ pdk new module --template-url=https://github.com/natemccurdy/pdk-templates --template-ref=PDK-1803/reproduction

...(snip interview dialog)...

~ $ cd test                                     
~/test $ ls -l validate_things.sh               
.rw-r--r-- 218 nate 15 Mar 13:50 validate_things.sh

~/src/test $ bash validate_things.sh
validate_things.sh is NOT executable
```

ghoneycutt:
+1 on fixing this as it makes development harder since templated scripts cannot just be executed locally and work arounds for CI/CD have to be put in place.

chelnak:
Hey! With tasks in the pipeline for puppet 8 readiness I don't think I'll personally have time to work on this.

However! This is open source! So if anyone wants to contribute towards this feature I'm happy to assist.

My only concern here is shipping around executable files could be seen as a security risk by some.. so we would need to cover that.

jordanbreen28:
see 
https://github.com/puppetlabs/pdk/pull/1289
 Merged
https://github.com/puppetlabs/pdk-templates/pull/539

PR #1284, opened 2023-11-02
Support executable templates

nabertrand:
## Summary

This adds support for having executable templates as described in #1224.

## Additional Context

This PR makes use of a new `manage_execute_permissions` boolean config setting, which can be set on an individual template or enabled globally in the `common` config key. The existing code relied on a `:global` config setting:
https://github.com/puppetlabs/pdk/blob/4e46ce2a74acb29b1be4165e1281a7e797100dc9/lib/pdk/template/renderer/v1/legacy_template_dir.rb#L59
But as far as I can tell that key is no longer valid as neither the `Symbol` class nor the `:global` symbol are allowed during YAML parsing: https://github.com/puppetlabs/pdk/blob/4e46ce2a74acb29b1be4165e1281a7e797100dc9/lib/pdk/template/renderer/v1/legacy_template_dir.rb#L94
I noticed the current config_defaults file contains a `common` key, so I thought we could reuse that here.

No spec or acceptance tests have been added and no documentation has been updated because I wanted to wait for consensus on this approach before spending time on testing or documentation.

## Related Issues (if any)
#1224

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/lib/pdk/module/convert.rb
+++ b/lib/pdk/module/convert.rb
@@ -148,7 +148,7 @@ module PDK
           module_name = new_metadata.nil? ? 'new-module' : new_metadata.data['name']
           metadata_for_render = new_metadata.nil? ? {} : new_metadata.data
 
-          template_dir.render_new_module(module_name, metadata_for_render) do |relative_file_path, file_content, file_status|
+          template_dir.render_new_module(module_name, metadata_for_render) do |relative_file_path, file_content, file_status, file_executable|
             absolute_file_path = File.join(module_dir, relative_file_path)
             case file_status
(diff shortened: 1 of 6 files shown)


jordanbreen28 (approved):

nabertrand:
@jordanbreen28 as discussed in Slack, I've added spec tests in #1289. There's also a related documentation update in https://github.com/puppetlabs/pdk-templates/pull/539

nabertrand:
Closing in favor of #1289

PR #1289, opened 2023-11-16
Support executable templates

nabertrand:
## Summary

This adds support for having executable templates as described in #1224. This is a duplicate of #1284 with spec tests added.

## Additional Context

This PR makes use of a new `manage_execute_permissions` boolean config setting, which can be set on an individual template or enabled globally in the `common` config key. The existing code relied on a `:global` config setting:
https://github.com/puppetlabs/pdk/blob/4e46ce2a74acb29b1be4165e1281a7e797100dc9/lib/pdk/template/renderer/v1/legacy_template_dir.rb#L59
But as far as I can tell that key is no longer valid as neither the `Symbol` class nor the `:global` symbol are allowed during YAML parsing: https://github.com/puppetlabs/pdk/blob/4e46ce2a74acb29b1be4165e1281a7e797100dc9/lib/pdk/template/renderer/v1/legacy_template_dir.rb#L94
I noticed the current config_defaults file contains a `common` key, so I thought we could reuse that here.

## Related Issues (if any)
#1224

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/lib/pdk/module/convert.rb
+++ b/lib/pdk/module/convert.rb
@@ -148,7 +148,7 @@ module PDK
           module_name = new_metadata.nil? ? 'new-module' : new_metadata.data['name']
           metadata_for_render = new_metadata.nil? ? {} : new_metadata.data
 
-          template_dir.render_new_module(module_name, metadata_for_render) do |relative_file_path, file_content, file_status|
+          template_dir.render_new_module(module_name, metadata_for_render) do |relative_file_path, file_content, file_status, file_executable|
             absolute_file_path = File.join(module_dir, relative_file_path)
             case file_status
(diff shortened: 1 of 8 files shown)


nabertrand:
Documentation update in https://github.com/puppetlabs/pdk-templates/pull/539

jordanbreen28 (approved):
Spot on @nabertrand!

nabertrand:
@jordanbreen28 hoping this and https://github.com/puppetlabs/pdk-templates/pull/539 can still get merged before the next PDK release. Let me know if there's anything I can do to help get them merged.

PR #1313, opened 2024-02-01
(CAT-1713) Add retry to flaky tests

david22swan:
## Summary
Provide a detailed description of all the changes present in this pull request.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -26,6 +26,30 @@ def default_installed_bin_dir
   end
 end
 
+# This method allows a block to be passed in and if an exception is raised
+# that matches the 'error_matcher' matcher, the block will wait a set number
+# of seconds before retrying.
+# Params:
+# - max_retry_count - Max number of retries
+# - retry_wait_interval_secs - Number of seconds to wait before retry
+# - error_matcher - Matcher which the exception raised must match to allow retry
+# Example Usage:
+# retry_on_error_matching(3, 5, /OpenGPG Error/) do
+#   apply_manifest(pp, :catch_failures => true)
+# end
+def retry_on_error_matching(max_retry_count = 3, retry_wait_interval_secs = 5, error_matcher = nil)
+  try = 0
+  begin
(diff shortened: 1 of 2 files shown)


LukasAud (approved):
LGTM

spec/spec_helper_acceptance.rb
```diff
@@ -26,6 +26,30 @@ def default_installed_bin_dir
...
+# retry_on_error_matching(3, 5, /OpenGPG Error/) do
+#   apply_manifest(pp, :catch_failures => true)
+# end
+def retry_on_error_matching(max_retry_count = 3, retry_wait_interval_secs = 5, error_matcher = nil)
+  try = 0
+  begin
+    try += 1
+    yield
+  rescue StandardError => e
+    raise unless try < max_retry_count && (error_matcher.nil? || e.message =~ error_matcher)
```
jordanbreen28, line 46: we should probably add an error matcher so we don't get unexpected behaviour with `error_matcher.nil?` or mask errors we aren't expecting

david22swan: Tried that and the `expected File to exist`

david22swan: Also, is that really an error?? At a glance it doesn't look like it has anything to do with the test and is just the last thing in the output before it tried to find the file and couldn't??

david22swan: To me, reading through the tests, that bit of output just reads as a part of the standard info dump that the PDK makes.

jordanbreen28: yeah my thoughts too.. Alrighty if theres no real error matcher you can supply its all good

PR #1335, opened 2024-04-03
(CAT-1628) Generalize version selection tests

david22swan:
Update tests to check against installing the Major versions of the PDK. This is due to the test environment only having certain versions available to install and thus necessitating that the version be updated more often than is preferred.
With this change that is no longer necessary and the test should not need to be updated again in this manner until the release of Puppet 9.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.


jordanbreen28 (approved):
LGTM!

PR #1337, opened 2024-04-10
(CAT-1796) Fix missing locale gem

LukasAud:
Following some investigation, it was noted that
the pdk gem could not work correctly when analytics were enabled. This was due to the fact that the
locale gem was missing for its correct operation.

This commit adds the locale gem to the pdk.gemspec to solve the current issue.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.


david22swan (approved):

Magisus:
This version of the locale gem (currently 2.1.4) triggers a license violation in CD4PE's Mend scans. There isn't a "fixed" version to go to, so we may need to get an exception or something. Just wanted to note that this might cause wider issues.
image

PR #1342, opened 2024-04-19
Support package testing with Docker

h0tw1r3:


h0tw1r3:
Confused why there's one failure. Unable to reproduce locally.

david22swan:
@h0tw1r3 Look's like it was just a transient failure.
Re-kick got it green.

PR #1356, opened 2024-05-15
(CAT-1643) Update childprocess and remove patch

david22swan:
Previously a specific patch was required in order to  have the code work as intended, however a patch has since been moved into the gem and so we can remove our patch and update the dependency.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.


david22swan:
Just waiting on pipeline fix to merge

jordanbreen28 (approved):

PR #1374, opened 2024-07-08
(CAT-1896) - Use puppet-modulebuilder for `PDK build`

jordanbreen28:
## Summary

This PR updates the pdk to use the module builder gem for its build functionality. The PDK build code had largely duplicated this, therefore we were maintaining two codebases with very similiar purpose.

The use of the modulebuilder gem means we now only have to maintain the build code in one place.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [x] Manually verified.


lib/pdk/module/release.rb
```diff
@@ -95,7 +95,7 @@ def write_module_metadata!
       def default_package_filename
         return @default_tarball_filename unless @default_tarball_filename.nil?
 
-        builder = PDK::Module::Build.new(module_dir: module_path)
+        builder = Puppet::Modulebuilder::Builder.new(module_path, nil, PDK.logger)
```
danadoherty639, line 98: we dont care about the target dir for releases?

jordanbreen28: Was trying to keep it as consistent with the previous code as before, no target-dir was passed to `PDK::Module::Build.new(module_dir: module_path)`. But, looks like a default was set from within the initalize method of that class, so nice catch! I'll update

jordanbreen28: And I should have said, we aren't actually building the module here, rather retrieving the package_file name. So target-dir isn't required!

gavindidrichsen (approved):
Looks fantastic!!

PR #1381, opened 2024-07-29
(CAT-1928) Update Supported OS on new module to match what is expected

david22swan:
Update the OS listed as supported in a default module to more accurately match what we would expect support for each OS to be.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/lib/pdk/module/metadata.rb
+++ b/lib/pdk/module/metadata.rb
@@ -9,7 +9,7 @@ module PDK
         'RedHat based Linux' => [
           {
             'operatingsystem' => 'CentOS',
-            'operatingsystemrelease' => ['7']
+            'operatingsystemrelease' => ['7', '8', '9']
           },
           {
             'operatingsystem' => 'OracleLinux',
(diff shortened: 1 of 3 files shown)


danadoherty639 (approved):

lib/pdk/module/metadata.rb
```diff
@@ -52,11 +60,11 @@ class Metadata
         },
         'Windows' => {
           'operatingsystem' => 'windows',
-          'operatingsystemrelease' => ['2019', '10']
+          'operatingsystemrelease' => ['2019', '10', '11']
```
jordanbreen28, line 63:
```suggestion
          'operatingsystemrelease' => ['2019', '2022', '10', '11']
```
should we add 2022?

david22swan: .... updated to include

jordanbreen28 (approved):

PR #1404, opened 2024-10-09
(CAT-2093) Enable shellcheck in CI

LukasAud:
Following the implementation of shellcheck in the CI, we are enabling it in the repositories we consider necessary.


gavindidrichsen (approved):
LGTM!

danadoherty639 (approved):

PR #1407, opened 2024-10-23
Disable the windows 2019 ruby 3.2 acceptance test

gavindidrichsen:
## Summary

This PR disables the windows 2019, ruby 3.2 acceptance test in the nightly.yml.  The main reason for this is that it can potentially hide actual failures as the nightly is always red.  Therefore, I've created an internal ticket [CAT-2122](https://perforce.atlassian.net/browse/CAT-2122) to fix the issue and then rollback this PR.

-----

For example, [3 days ago our nightly.yml](https://github.com/puppetlabs/pdk/actions/runs/11430936125) looked like this, which was what we'd come to expect over the last lot of weeks:

image

However, [2 days ago nightly](https://github.com/puppetlabs/pdk/actions/runs/11470686046) revealed further errors (and hidden)

image

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -38,6 +38,10 @@ jobs:
         ruby_version:
           - "2.7"
           - "3.2"
+        # TEMPORARILY exclude the following until CAT-2122 is resolved
+        exclude:
+          - os: "windows-2019"
+            ruby_version: "3.2"
         include:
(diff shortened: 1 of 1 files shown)


gavindidrichsen:
Testing the nightly produces output without the Windows 2019, ruby 3.2 (as below).  Note that all acceptance tests are failing as expected with the new error that needs to be fixed.

image

david22swan (approved):

PR #1410, opened 2024-10-24
(CAT-2130) Pin json to prevent issues

david22swan:
```
Failure/Error: require 'json'
ArgumentError:
  wrong number of arguments (given 1, expected 0)
```
As of the release of json v2.7.3 the above error has been occurring across the entirety of the PDK test environment.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.


david22swan:
https://github.com/ruby/json/issues/646#issuecomment-2435982078
Seems that if both are present they need to be set to the same

david22swan:
Associated issue on json: https://github.com/ruby/json/issues/646

david22swan:
Failures are expected, they are present on main and a followup PR will handle them

pmcmaw (approved):

PR #1411, opened 2024-10-28
(CAT-2128) Remove `json_pure` dependency

david22swan:
This was originally added to ensure JSON would be available in all environments but should no longer necessary as JSON has been added as a default ruby gem.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.


danadoherty639 (approved):

david22swan:
Screenshot 2024-11-04 at 4 08 08 PM

PR #1417, opened 2024-11-11
(CAT-2122) Re-initialize windows-2019/ruby-3.2 nightly test run

david22swan:
With the recent removal of `JSON::Pure` from the code, the dependency error that had been causing failures on this specific box, and not on local vagrant/docker/vmpooler runs with the same  windows-2019/ruby-3.2 setup has been resolved.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -38,10 +38,6 @@ jobs:
         ruby_version:
           - "2.7"
           - "3.2"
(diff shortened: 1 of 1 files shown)


LukasAud (approved):
LGTM

PR #1433, opened 2025-01-31
(DOC) Update pdk-template known issues

david22swan:
## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/docs/pdk_known_issues.md
+++ b/docs/pdk_known_issues.md
@@ -20,7 +20,7 @@ sudo chown -R example.user /opt/puppetlabs/pdk/share/cache/pdk-templates.git
 git config --global --add safe.directory /opt/puppetlabs/pdk/share/cache/pdk-templates.git
 ```
 
-The other solution is to target a seperate template location a shown below:
+The other solution is to target a seperate template location as shown below:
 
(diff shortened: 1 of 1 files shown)


docs/pdk_known_issues.md
```diff
@@ -10,11 +10,13 @@ pdk (FATAL): Unable to find a valid module template to use.
...
 
-This error is most commonly found when creating a new module or attempting to update a module that was previously created from the default templates, within an airgapped environment. There are two possible solutions that we have found for this, with the first one being to grant the `pdk-templates.git` directory packaged with the pdk the permissions that it requires in order for it to be used, either by confirming the local user as the owner of the files through the `chown` command or by setting it as a safe.directory going forward.
+This error is most commonly found when creating a new module or attempting to update a module that was previously created from the default templates, within an airgapped environment. There are two possible solutions that we have found for this, with the first one being to grant the `pdk-templates.git` directory packaged with the pdk the permissions that it requires in order for it to be used, either by confirming the local user as the owner of the files through the `chown` command or by setting it as a safe.directory going forward through git.
```
pmcmaw, line 13:
```suggestion
This error is most commonly found when creating a new module or attempting to update a module that was previously created from the default templates, within an airgapped environment. There are two possible solutions that we have found for this, with the first one being to grant the `pdk-templates.git` directory packaged with the pdk the permissions that it requires in order for it to be used, either by confirming the local user as the owner of the files through the `chown` command or by setting it as a safe directory via git.
```

pmcmaw (approved):

jordanbreen28 (approved):

PR #1438, opened 2025-02-19
(CAT-2238) Update latest_template? check to compare version numbers

david22swan:
Previously the code would check only if the given tag matched the shipped pdk-templates version number. With this change it will now also first check if the given tag is a version number and then if it is check if it is greater than the shipped number.
As a note, this change is related to an info block that is given when running pdk update and has no effect on the actual function of the PDK.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [x] Manually verified.

### DIFF
--- a/lib/pdk/module/update.rb
+++ b/lib/pdk/module/update.rb
@@ -81,6 +81,8 @@ module PDK
       private
 
       def latest_template?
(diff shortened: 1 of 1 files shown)


danadoherty639 (approved):
LGTM

